PR #98, opened 2022-11-04
Types of curried functions in OCaml

GuillaumeGen:
This creates commands `append_spaced_soft_ancestor_line` and analogous, which behaves as if several nodes of the same type nested one under the other were only one node.

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -317,9 +317,17 @@
 )
 
 (
-  "->" @append_spaced_soft_ancestor_line
+  "->" @append_spaced_softline
   .
   [ (comment) ]* @do_nothing
+)
+
+(function_type
+  "->"
+  .
+  (function_type
+    "->" @append_spaced_soft_ancestor_line
+  )
 )
 
 ; Always put softlines before these:
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -306,7 +306,6 @@
     "of"
     "struct"
     "then"
-    "->"
     "{"
     ":"
     ";"
@@ -315,6 +314,12 @@
   [
     (comment)
   ]* @do_nothing
+)
+
+(
+  "->" @append_spaced_soft_ancestor_line
+  .
+  [ (comment) ]* @do_nothing
 )
 
 ; Always put softlines before these:
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -330,10 +330,19 @@
 ;
 
 ; This query guarantees that a break occurs after the first arrow of a multiline type.
-(
-  "->" @append_spaced_softline
-  .
-  [ (comment) ]* @do_nothing
+(function_type
+  (
+    "->" @append_spaced_softline
+    .
+    (comment)* @do_nothing
+  )
+)
+(function_type
+  (
+    "->"
+    .
+    (comment)+ @append_spaced_softline
+  )
 )
 
 ; But the first query alone is not sufficient, because Ocaml arrows are binary operators,
@@ -367,11 +376,38 @@
 (function_type
   "->"
   .
+  (comment)*
+  .
   (function_type
     "->" @append_spaced_soft_ancestor_line
     .
-    [ (comment) ]* @do_nothing
+    (comment)* @do_nothing
   )
+)
+
+(function_type
+  "->"
+  (comment)*
+  .
+  (function_type
+    "->"
+    .
+    (comment)+ @append_spaced_soft_ancestor_line
+  )
+)
+
+; In matches, if a line break is required, we want it to happen after the last comment following the arrow.
+
+(match_case
+  "->"  @append_spaced_softline
+  .
+  (comment)* @do_nothing
+)
+
+(match_case
+  "->"
+  .
+  (comment)+ @append_spaced_softline
 )
 
 ; Always put softlines before these:
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -369,6 +369,8 @@
   .
   (function_type
     "->" @append_spaced_soft_ancestor_line
+    .
+    [ (comment) ]* @do_nothing
   )
 )
 
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -316,12 +316,54 @@
   ]* @do_nothing
 )
 
+; If a type is written on several lines, then the types of all the arguments should be separated.
+;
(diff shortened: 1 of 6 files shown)


GuillaumeGen:
What is the status here? Should we merge it as is, or is multiplying commands something we absolutely want to avoid?

torhovland:
I thought maybe you intended to do some more investigation first. I am a bit uneasy about introducing the new instructions when it is unclear if they will solve the problem well enough. But I am fine with merging this if you promise to clean things up again if this ends up not producing a satisfactory end result 🙂

What do you think, @aspiwack?

GuillaumeGen:
Well, it works well on the examples this PR adds to the test.

languages/ocaml.scm
```diff
@@ -189,6 +188,20 @@
   ]* @do_nothing
 )
 
+(
```
torhovland, line 191: It would be useful with some code examples here that demonstrate why the queries need to be like this.

torhovland: Also, would it make sense to merge these two patterns into one?

torhovland (changes requested):

aspiwack:
Can I have a summary of the design discussion we've had then. What are the various alternatives that we've floated, and what are their pros and cons?

languages/ocaml.scm
GuillaumeGen: I added a quite long comment with a code example. Was it the kind of explanation you expected?

GuillaumeGen: Well, I spoke too fast. The code written in the first bullet point failed the test, but only because the query 
```
(match_case
  "->"  @append_spaced_softline
  .
  (comment)* @do_nothing
)
```
did not exist. If I replace the 2 queries about arrows by those two, the current tests are passing.

aspiwack:
Nevertheless, let's do the exercise (if only because I don't really remember the details of this conversations): what are the various alternatives that we have discussed in this PR? What are their pros and cons? What is your (@GuillaumeGen) recommendation among the alternatives and why?

GuillaumeGen:
What have been discussed in this PR is:
 1. Create the `*_ancestor_*` commands to have commands deciding if a line should be break depending on the number of lines an ancestor node span on. This is what the proposed code do.
  2. Consider that by default nodes of the same nature leaving one under the other should just be seen as if they were flatten, so make the behavior of the `*_ancestor_*` the only possible one.
  3. Implement a `coalesce_same_ancestor` modifier, which would alter the tree obtained after parsing.
  4. Implement a `coalesce_same_ancestor` which would be a local modifier of the commands in the same query.
  
Pros and cons of each
 1. Pros: Already implemented, homogeneous with the rest of the command design,
    Cons: Multiplying commands will hurt maintainability at some point.
 2. Pros: Interface is simple, deciding if nodes should be merged is not on the shoulder of the query writer anymore,
    Cons: Lose flexibility, cannot handle not associative operators.
 3. Pros: A quite natural interface, only one keyword,
    Cons: Require some new tricks since we did not modified the AST yet, it introduces a notion of order between the queries since some match are only true on the (un)modified AST (so probably a preprocessor is required)
 4. Pros: It adds only one keyword,
    Cons: It is essentially the same as 1 but with a different syntax, less homogeneous with the other 8 queries about lines.
    
My opinion:
Since arrow is a non-associative operator and this PR is mainly to address an issue we met with arrows, I would rule out 2.
I have nothing against 3, but I have the feeling that it will add a preprocessor per language, and the more language specific things we have the less generic the formatter is. It does not seem to go in the direction of (my understanding of) the project target.
Regarding 1 and 4, this is only an interface difference. I like the homogeneity of 1, and in the current state this homogeneity is a plus for the user interface, but if we anticipate a multiplication of the modifiers, then a more composable interface would definitely be better (if I have to wonder if I should write `@append_modifierX_spaced_modifierY_newline` or `@append_spaced_modifierY_modifierX_newline`, I might overpopulate #catharsis quite quickly).

languages/ocaml.scm
torhovland: Great, thanks!

torhovland:
@aspiwack I'm OK with this. Please approve unless you have reservations.

aspiwack:
@GuillaumeGen So you are of the opinion that these multiple `soft_ancestor_line` commands are the best in the foreseeable future? Not only as a short term workaround?

I don't really understand the objection against 3


> I have nothing against 3, but I have the feeling that it will add a preprocessor per language, and the more language specific things we have the less generic the formatter is. It does not seem to go in the direction of (my understanding of) the project target.

What would be language specific in this design?

GuillaumeGen:
> So you are of the opinion that these multiple soft_ancestor_line commands are the best in the foreseeable future? Not only as a short term workaround?

In my anticipation of future, yes. But I might lack imagination, if going for the most future-proof solution is key in this decision, option 4 is clearly more future-proof.

> What would be language specific in this design?

It would multiply the number of language-specific files. Plus the purpose of the pre-processor is to tweak the AST to ship away the choices we do not like in the tree-sitter grammars, so the queries would be specific to the micro-decisions each grammar writer did, hence the sharing between languages might be low (at least lower than for formatting queries).

GuillaumeGen:
> It would? I thought there would be just the one query file. What am I missing?

Not sure of what you have in mind, but currently the order in the query file is not relevant. If we start to alter the AST, then order is meaningful. We then probably want a separate pre-processor file with queries altering the AST rather than having a mutable AST potentially altered by any query.

> Do we anticipate other similar queries besides “considering a bunch of identical nodes as a node”? If so, will this be a problem in approach 1 as well (you would have to add {append,prepend}_{empty,spaced}_foobar_lines for all these cases?

My lack of imagination once again. That's precisely why I say that approach 4 is approach 1 with a more future-proof interface.

torhovland:
> A simple way to implement 3 (which is similar to 4, I guess), would be to say that if a node has both @append_spaced_softline and @coalesce_ancestors it behaves as you wrote @append_space_soft_ancestor_line in your implementation. This only requires that we can list the predicates on a node, maybe we can't? If so there may be other implementation.

One way to do it is to give `@coalesce_ancestors` a higher precedence, possibly doing an entire pass where we just collect node id-s that should "coalesce ancestors". Then we do another, proper parsing pass, where we can use this information when processing actions like `@append_spaced_softline`.

aspiwack:
So, do we have a conclusion on this? What do we think is the best way 1/ right now 2/ in the long term?

PR #102, opened 2022-11-15
Rework postprocess

GuillaumeGen:
This new `post_process` function only goes once through the vector of atoms in the `AtomCollection`.

### DIFF
--- a/src/atom_collection.rs
+++ b/src/atom_collection.rs
@@ -1,6 +1,5 @@
 use std::collections::{BTreeSet, HashMap, HashSet};
 
-use itertools::Itertools;
 use tree_sitter::Node;
 
 use crate::{Atom, FormatterError, FormatterResult};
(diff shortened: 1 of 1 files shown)


src/atom_collection.rs
```diff
@@ -284,79 +264,72 @@ impl AtomCollection {
...
-        for i in 0..self.atoms.len() - 1 {
-            if self.atoms[i] == after {
-                for j in i + 1..self.atoms.len() {
-                    if self.atoms[j] != before
-                        && self.atoms[j] != after
-                        && !ignoring.contains(&self.atoms[j])
-                    {
-                        // stop looking
-                        break;
+fn post_process_internal(new_vec: &mut Vec<Atom>, old_vec: &Vec<Atom>) {
```
torhovland, line 274: I think we need a comment that describes the ordering that this function dictates.

torhovland (approved):
Nicely done!

PR #107, opened 2022-12-01
bundle the language directory

figsoda:
this introduces the environment varialbe  `TOPIARY_LANGUAGE_DIR` at build time and allows running topiary outside of the directory

this will make it easier to be packaged in nixpkgs and also allows `nix run`ing without cloning the repository

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -70,7 +70,7 @@ fn run() -> FormatterResult<()> {
     let query_path = if let Some(query) = args.query {
         query
     } else if let Some(language) = args.language {
-        PathBuf::from(option_env!("TOPIARY_LANGUAGE_DIR").unwrap_or("language"))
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -69,7 +70,8 @@ fn run() -> FormatterResult<()> {
     let query_path = if let Some(query) = args.query {
         query
     } else if let Some(language) = args.language {
-        str::to_lowercase(format!("languages/{language:?}.scm").as_str())
+        PathBuf::from(option_env!("TOPIARY_LANGUAGE_DIR").unwrap_or("language"))
```
torhovland, line 73: Should be `unwrap_or("languages")`, right?

torhovland (changes requested):
Nice PR!

src/main.rs
figsoda: fixed it

torhovland (approved):

torhovland:
Thanks!

Issue #90, opened 2022-10-10
Error if unknown capture name is encountered in query


torhovland:
Currently, any capture names are either processed if they are known or otherwise ignored. We don't return an error message if we encounter an unknown capture name, but we should.

Example:

```scheme
"begin" @append_foo
```

This pattern will simply not do anything.

PR #109, opened 2022-12-06
Deal with unknown capture names when resolving query

Xophmeister:
This is a first step wrt solving #90. Panicking is a little extreme, but it's better than nothing; the ultimate goal (in PRs to come) is to make encountering unknown capture groups a parse error and bail-out gracefully.

### DIFF
--- a/src/atom_collection.rs
+++ b/src/atom_collection.rs
@@ -116,9 +116,13 @@ impl AtomCollection {
             "prepend_spaced_softline" => self.prepend(Atom::Softline { spaced: true }, node),
             // Skip over leafs
             "leaf" => {}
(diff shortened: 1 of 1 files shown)


Xophmeister:
Maybe this would be better as a draft, so I can add iterate through subsequent commits...

aspiwack:
I'd ask the question this way: is this PR an improvement in its current state? If it is (I suspect that it is), let's merge, and you can make another PR later to improve further. Otherwise keep it as draft.

GuillaumeGen:
My understanding is also that it improves the current state and should be merged.

torhovland (approved):

Issue #106, opened 2022-12-01
Topiary reports a parsing error on a valid OCaml file


Niols:
## Bug Description

Topiary reports parsing errors on OCaml files manipulating [first-class modules](https://v2.ocaml.org/manual/firstclassmodules.html), even when they are valid OCaml files. The error does not seem to come from the OCaml tree-sitter grammar that parses such constructions just fine.

## Expected Behaviour

Considering that tree-sitter parses such constructions correctly, I would expect Topiary to succeed in parsing, potentially failing elsewhere. As a user, I would expect that running Topiary on my file either succeeds (even if it leads to an odd formatting) or reports an error that does not have to do with parsing. For instance, something about an unsupported construct in the query file?

## Environment

- NixOS unstable with [nixpkgs](https://github.com/NixOS/nixpkgs/) at commit [`6060f12`](https://github.com/NixOS/nixpkgs/commit/6060f1268ca7b44d62304564eadb0b0667ea8b84).
- OCaml and tree-sitter from aforementioned [nixpkgs](https://github.com/NixOS/nixpkgs/).
- Topiary on `main` as of December 1st, 2022, at commit 534b3d0.
- [tree-sitter-ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) at commit [`cc26b1e`](https://github.com/tree-sitter/tree-sitter-ocaml/commit/cc26b1ef111100f26a137bcbcd39fd4e35be9a59).

## Reproduction

My protocol is to create a file using first-class modules and to run OCaml, Topiary and tree-sitter on it, to show that both OCaml and tree-sitter agree that the file is syntactically correct while Topiary reports a syntax error. Each invocation of `nix run github:nixos/nixpkgs/6060f12#<command> --` (or `nix run <flake>#<command> --` in general) can be replaced by a direct invocation of `<command>` if you have it on your machine.

1. Create an ML file named `issue106.ml` with the following content:
   ```ocaml
   module type FOO = sig val foo : string end
   let create foo : (module FOO) = (module struct let foo = foo end)
   module Foo = (val create "Issue #106")
   ```

2. Confirm that it is a valid OCaml file by running OCaml on it and not getting an error:
   ```console
   $ nix run github:nixos/nixpkgs/6060f12#ocaml -- issue106.ml
   ```

2. Run Topiary as of commit 534b3d0 on it. For instance, I would run:
   ```console
   $ commit=534b3d0
[...]

PR #110, opened 2022-12-07
Support first-class module

GuillaumeGen:
This implements the parts missing to support the first-class module example given by Niols in #106.
It includes: 
 - dealing with the line breaks between declaration of different natures, 
 - opening the indentation for module signature which is closed by the `end` keyword 
 - removing the spaces between parentheses and keywords. This last bit is not working as expected, due to a bug in tree-sitter queries: https://github.com/tree-sitter/tree-sitter/issues/1811 . There are 3 options to deal with it: 
     1. Chose that we always want spaces around parentheses (quite unusual in the OCaml world, but the default in many languages), 
     2. Write the queries in a not elegant, but bug-avoiding way,
     3. Spend time working on the tree-sitter repository to fix the issue upstream.

The [golden rule of software quality](https://www.haskellforall.com/2020/07/the-golden-rule-of-software-quality.html) would push toward option 3, but it is the less sure and longest path.

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -14,9 +14,10 @@
 ; Allow blank line before
 [
   (comment)
-  (external)
   (exception_definition)
+  (external)
   (module_definition)
+  (module_type_definition)
   (open_module)
   (type_definition)
   (value_definition)
@@ -75,6 +76,7 @@
   "else"
   "exception"
   "external"
+  "if"
   (infix_operator)
   "let"
   "match"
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -87,7 +100,46 @@
...
+    "|"
+    "||"
+    "->"
+    "<-"
+    "{"
+    "}"
+  ] @prepend_space
+)
 
 ; Prepend spaces
```
torhovland, line 144: I think you can merge all of these with the above that checks for an opening paren.

torhovland (approved):

aspiwack:
> * There are 3 options to deal with it:
>       
>       1. Chose that we always want spaces around parentheses (quite unusual in the OCaml world, but the default in many languages),
>       2. Write the queries in a not elegant, but bug-avoiding way,
>       3. Spend time working on the tree-sitter repository to fix the issue upstream.
> 
> 
> The [golden rule of software quality](https://www.haskellforall.com/2020/07/the-golden-rule-of-software-quality.html) would push toward option 3, but it is the less sure and longest path.

I think you want to do (2) as a stop-gap solution, and see if you can do something to help with (3). Maybe not immediately, but soonish.

GuillaumeGen:
Solution (ii) implemented, but this is quite ugly, as it requires to multiply very similar queries, which would all go in a single alternative otherwise.

Issue #112, opened 2022-12-07
Idempotence check should not generate parse error


GuillaumeGen:
When a query file contains mistakes, it can happen that topiary produces invalid code. That is of course an issue. Currently, to detect it, in the test topiary is ran a second time on the output it produces. Unfortunately, when this second run fails, the reported error is the one of the failing run, without any indication that the error originates from the idempotence checker.

That is the issue in #106 where it was thought for several days that topiary failed to parse a valid Ocaml file, whereas the truth was that topiary generated an invalid Ocaml file when formatting a valid one.

We would like to have error messages mentioning explicitly that it is the idempotence check that failed.

PR #113, opened 2022-12-07
A small error handler for exception raised during the idempotence check

GuillaumeGen:
This tries to solve #112 

When running 
```
cargo install --path .
topiary --query languages/ocaml.scm --input-file tests/issue-106.ml 
```
the new output is 
```
Error: Idempotence found this error when running topiary on its own output:
        Parsing error between line 420, column 26 and line 420, column 37
```


src/error.rs
```diff
@@ -92,6 +104,7 @@ impl Error for FormatterError {
             Self::Writing(WritingError::FromUtf8(source)) => Some(source),
             Self::Writing(WritingError::IntoInner(source)) => Some(source),
             Self::Writing(WritingError::Io(source)) => Some(source),
+            Self::Formatting(err) => err.source(),
```
torhovland, line 107: I think the source should be `Some(err)` (which in turn has `err.source()`).

src/error.rs
```diff
@@ -75,6 +80,13 @@ impl fmt::Display for FormatterError {
             Self::Internal(message, _) | Self::Query(message, _) => {
                 write!(f, "{message}")
             }
+            Self::Formatting(err) => {
+                write!(
+                    f,
+                    "Idempotence found this error when running topiary on its own output:\n\t"
```
torhovland, line 86: You may want to match `Self::Formatting(Idempotence)` and `Self::Formatting(_)` separately, with messages like this:

"The formatter did not produce the same result when invoked twice (idempotence check). This is an error. It would be helpful if you logged this at https://github.com/tweag/topiary"

"The formatter errored when trying to format the code twice (idempotence check). This probably means that the formatter produced invalid code, and is an error. It would be helpful if you logged this at https://github.com/tweag/topiary"

torhovland (approved):

src/error.rs
Niols: +1 to make explicit that this is a bug in Topiary itself that should be reported. The current message is an improvement but still leaves a doubt to the user that maybe it comes from the OCaml code. (For instance if they use Topiary but don't know what the idempotence check is.)

Also +1 to make the difference between various bugs in the idempotence check. That being said, isn't it already the case through the report of the found error? Like in this case I understand that it is a parse error and therefore that the output of the first pass was not parseable, which is different from parseable but different from the output of the second pass.

Niols (approved):

src/error.rs
GuillaumeGen: You are right that `Formatting(Idempotence)` should have the message that currently `Idempotence` has, and the other `Formatting(...)` a message in a different spirit.
Agree about calling to log on GitHub. This was not done in the error messages, but only in the comments previously, so I reproduced it. Will change now.

src/error.rs
```diff
@@ -75,6 +80,13 @@ impl fmt::Display for FormatterError {
             Self::Internal(message, _) | Self::Query(message, _) => {
                 write!(f, "{message}")
             }
+            Self::Formatting(err) => {
```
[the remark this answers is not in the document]
GuillaumeGen: It is associated with idempotence check, but if the output produced by topiary is not parsable, it has nothing to do with idempotence. I used `Formatting` because it was the one you suggested in https://github.com/tweag/topiary/issues/106#issuecomment-1341003683 and it sounded quite accurate to me. No strong opinion on which word should be used.

src/error.rs
GuillaumeGen: After further thought, I think that `Formatting(Idempotence)` does not make sense, so I put the `Idempotence` error outside of the `Formatting` wrapper.

src/error.rs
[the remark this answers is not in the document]
torhovland: I don't have a very strong opinion either. I agree that the error is a formatting error, but the resulting error message is associated with the idempotency check. So if someone decided to throw a `Formatting(err)` in the future, they would get an error message about idempotency.

PR #115, opened 2022-12-08
Detect language from input file extension

### LINKED ISSUE #3: Try to detect language from file extension if not specified as command line argument
This depends on #9 and is only relevant when an input file has been provided.

Xophmeister:
Detect the language by examining the filename extension, if there is one.

:bulb: I have changed the semantics of the command line arguments. Before you were required to supply exactly one of `--language` or `--query`; now you must provide at least one of `--language`, `--query` and `--input-file`. It is not communicated to the user that there is a precedence when it comes to determining the language (i.e., query > language > input file).

:warning: Some of the detection code is influenced by Difftastic. It's not exactly copied verbatim, but it's close enough to make me nervous. To that end, I have included Difftastic's license (MIT) in the relevant source file and highlighted the code that falls under its purview as a first approximation of compliance.

Resolves #3

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -1,4 +1,26 @@
-use std::path::PathBuf;
+// Portions of this code, highlighted below, are influenced by Difftastic (MIT licensed):
+//
+// Copyright (c) 2021-2022 Wilfred Hughes
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use std::path::{Path, PathBuf};
 
 use crate::{FormatterError, FormatterResult};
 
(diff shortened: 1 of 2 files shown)


src/language.rs
```diff
@@ -1,3 +1,27 @@
...
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
```
aspiwack, line 13: I don't think that this count as a “substantial portion of the software”. Feel free to cite Difftastic as an influence, but you probably don't need to include the licence here.

Xophmeister: Replaced with citations in a3276b8

src/language.rs
```diff
@@ -22,4 +51,40 @@ impl Language {
...
+            }
+
+            return Err(FormatterError::Query(
+                format!("Cannot detect language from unknown extension: '{filename}'"),
+                None,
+            ));
+        }
+
+        Err(FormatterError::Query(
+            format!("Cannot detect language without extension: '{filename}'"),
```
torhovland, line 76:
```suggestion
            format!("Cannot detect language of file '{filename}'. Try specifying language explicitly."),
```

src/language.rs
```diff
@@ -22,4 +51,40 @@ impl Language {
             )),
         }
     }
+
+    pub fn detect(filename: &str) -> FormatterResult<&str> {
+        // NOTE This extension search is influenced by Wilfred Hughes' Difftastic
+        // https://github.com/Wilfred/difftastic/blob/master/src/parse/guess_language.rs
+        if let Some(extension) = Path::new(filename).extension() {
```
torhovland, line 58: I think for the errors returned here a new variant `FormatterError::LanguageDetection` is warranted.

src/language.rs
```diff
@@ -22,4 +51,40 @@ impl Language {
...
+            for (language, extensions) in EXTENSIONS {
+                for candidate in extensions.iter() {
+                    if *candidate == extension {
+                        return Ok(*language);
+                    }
+                }
+            }
+
+            return Err(FormatterError::Query(
+                format!("Cannot detect language from unknown extension: '{filename}'"),
```
torhovland, line 70:
```suggestion
                format!("Cannot detect language of file '{filename}' due to unknown extension '{extension}'. Try specifying language explicitly."),
```

src/language.rs
```diff
@@ -22,4 +51,40 @@ impl Language {
...
     }
+
+    pub fn detect(filename: &str) -> FormatterResult<&str> {
+        // NOTE This extension search is influenced by Wilfred Hughes' Difftastic
+        // https://github.com/Wilfred/difftastic/blob/master/src/parse/guess_language.rs
+        if let Some(extension) = Path::new(filename).extension() {
+            let extension = extension.to_str().unwrap();
+
+            for (language, extensions) in EXTENSIONS {
+                for candidate in extensions.iter() {
```
torhovland, line 62: Here you can use https://doc.rust-lang.org/rust-by-example/fn/closures/closure_examples/iter_any.html

torhovland (changes requested):

src/language.rs
Xophmeister: Added in 6d9a1d4171c12e9aea

src/language.rs
Xophmeister: Changed in 6d9a1d4171c12e9aea

src/language.rs
Xophmeister: Changed in 6d9a1d4171c12e9aea

src/language.rs
```diff
@@ -66,16 +66,14 @@ impl Language {
                 }
             }
 
-            return Err(FormatterError::Query(
-                format!("Cannot detect language from unknown extension: '{filename}'"),
-                None,
-            ));
+            return Err(FormatterError::LanguageDetection(format!(
```
torhovland, line 69: Actually now that I think about it, it would be nice if `FormatterError::LanguageDetection` took just the file name (and possibly extension) as parameter, and left it to `FormatterError::fmt` to take care of the error message.

Xophmeister: Good idea :ok_hand:

src/language.rs
Xophmeister: Changed in 40c4d5d467d7d36b8

src/language.rs
Xophmeister: Done in 6d439aabc376df6

PR #127, opened 2022-12-14
Add line break after equal for multiline definition

GuillaumeGen:
Add a line break after = in multiline definition, both to be faithful with the presented example and to have an idempotent formatter.

### DIFF
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -437,7 +437,8 @@ let topological_sort deps =
       raise @@ Dep_error (CircularDependencies (node, path));
     if List.mem node visited then visited
     else
-      let edges = try
+      let edges =
+        try
           List.assoc node graph
         with
           Not_found ->
@@ -483,3 +484,16 @@ type _ variant +=
 
 (* Two times. *)
 type _ variant += WithBuff : t variant
+
+let add_multiline x =
+  let res =
+    x + x
+  in
+  res
+
+let add_one_line x = let res = x + x in res
+
+let add_as_fun_multiline = fun x ->
+  x
+
+let add_as_fun_one_line = fun x -> x
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -395,7 +395,7 @@
   (match_case) @prepend_spaced_softline
 )
 
-; Multi-line definitions must have a linebreak before "in":
+; Multi-line definitions must have a linebreak after "=" and before "in":
 ;
 ; let a =
 ;   expression
@@ -410,6 +410,25 @@
   .
   "in"
 )
+; There are special cases however. We do not want to break lines after "=" when writing
+;
+; let f = function
+;   | Constructor -> expression
+;
+; or
+;
+; let f = fun x ->
+;   expression
+;
+(let_binding
+  "=" @append_spaced_softline
+  .
+  [
+    (function_expression)
+    (fun_expression)
+  ]* @do_nothing
+)
+
 (value_definition
   "and" @prepend_spaced_softline
 )
--- a/tests/samples/input/ocaml.ml
+++ b/tests/samples/input/ocaml.ml
@@ -483,3 +483,15 @@ type _ variant +=
 
 (* Two times. *)
 type _ variant += WithBuff : t variant
+
+let add_multiline x =
+  let res =
+    x + x in
+  res
+
+let add_one_line x = let res = x + x in res
+
+let add_as_fun_multiline = fun x ->
+  x
+
+let add_as_fun_one_line = fun x -> x


GuillaumeGen:
Without it, 
```
let f =
  res
in
```
is formatted as 
```
let f = res
in
```
since there is no line break after the equal,
and this second thing is itself formatted as
```
let f = res in
```
since the definition of `f` now spans on one line.

tests/samples/expected/ocaml.ml
```diff
@@ -222,7 +222,8 @@ let add_channel b ic len =
 let output_buffer oc b =
   output oc b.buffer 0 b.position
 
-let closing = function
+let closing =
```
torhovland, line 225: On this one, I'd like to ask @Niols and @aspiwack for advice.

aspiwack: Are you familiar with [George Orwell's _A Nice Cup of Tea_](https://www.orwellfoundation.com/the-orwell-foundation/orwell/essays-and-other-works/a-nice-cup-of-tea/) essay? I promise that it's on topic.

Anyway, I would keep `function` trailing as in

```ocaml
let closing = function
  | '(' -> ')'
  | '{' -> '}'
  | _ -> assert false
```

(this is the same as before change)

There are few such hanging constructions, I can think of the more complicated

```ocaml
let foo : int -> int -> int = fun x y ->
  fibonacci_is_a_long_name_to_write x + isn't_it_yes_this_is_a_legal_function_name y
````

I usually hang `match` myself, but this looks like a rarer style (so we should probably not go this route):

```ocaml
let bar x y = match x,y with
  | true, false -> "banana"
  | false, true -> "apple"
  | _, _ -> fail_with "I can't think of another fruit"
```

It's probably all I would leave hanging like this.

aspiwack:
> ```
> let f = res
> in
> ```
> 
> since there is no line break after the equal, and this second thing is itself formatted as
> 
> ```
> let f = res in
> ```
> 
> since the definition of `f` now spans on one line.

Independently of the proposed change, this should probably be considered a bug.

tests/samples/expected/ocaml.ml
torhovland: > Are you familiar with [George Orwell's _A Nice Cup of Tea_](https://www.orwellfoundation.com/the-orwell-foundation/orwell/essays-and-other-works/a-nice-cup-of-tea/) essay? I promise that it's on topic.

So is that a hint that I am being pedantic?

By the way, I find it amusing that the British can argue about whether or not to put sugar in tea, but when it comes to milk, the notion of not using it (in tea or coffee) seems completely alien to them.

GuillaumeGen: Since a `match` can contain arbitrary terms, especially the ones for which the formatter prints a like break, I think that putting it on the line of the equal will lead to quite unaesthetic (or inconsistent) result, so I am in favour of breaking the line before it.
For `fun` and `function`, I have no opinion about it. When I manually format, I tend to break line before it, but this is not among the things I am ready to fight for.

torhovland: The reason I asked was that I was unsure if there is an established OCaml style about this. But it sounds like perhaps not.

GuillaumeGen:
> Independently of the proposed change, this should probably be considered a bug.

What do you mean by "a bug"?
Well, this question is quite simple. If there is a valid Ocaml code on which Topiary answers "Error: idempotence check failed" (or whatever the exact error message is), then it is a bug.
More precisely, my question would be: Of which tool is it a bug, according to you?
I considered it as a bug of the query files so changed it. But if read well between the lines of your message, you consider that it is more deep than "The query file missed one case".

GuillaumeGen:
> Independently of the proposed change, this should probably be considered a bug.

Or maybe you simply meant that 
```
let f x = res
in
```
should not be formatted as 
```
let f x = res in
```
and that we should respect the line break the user put in this case.

aspiwack:
> Or maybe you simply meant that

Indeed. Though it can be argued otherwise, I tend to think of `let f x = res in` as a node (which can be either multiline or single line). And in this perspective, we are reformatting a multiline node into a single line node.

Supposing the changes in this PRs (which I approve of), it would mean that

```ocaml
let f x = res
in
```

ought to be formatted as

```ocaml
let f x =
  res
in
```

tests/samples/expected/ocaml.ml
aspiwack: @torhovland 
> So is that a hint that I am being pedantic?

Not at all. I'm saying that a lot of questions regarding formatting have one obvious unquestionable answer to me. And that all these answers are wrong for someone else.

@GuillaumeGen 
> For fun and function, I have no opinion about it. When I manually format, I tend to break line before it, but this is not among the things I am ready to fight for.

Honestly, I don't know what other people do. I'll ask.

GuillaumeGen:
I think both views are relevant. I mean, in your suggestion, you are transforming the `let f x = res` node from a single one to a multiline one. And since `let` can exist without `in` (toplevel `let`) and `let f x = res in` cannot exist without an expression after, the decision taken by the writer of tree-sitter-ocaml was that `let f = res in expr` leads to the tree
```
(let_expression)
  (value_definition)
    "let"
    (let_binding)
      <f>
      "="
      <res>
  "in"
  <expr>
```
Of course, we should not let the formatting we want to obtain be decided by the pre-existing cutting into nodes, however, it has a rationale too.

tests/samples/expected/ocaml.ml
Niols: For `function`, personally, I'd go for:
```ocaml
let closing = function
  | '(' -> ')'
  | '{' -> '}'
  | _ -> assert false
```
and I feel this one is fairly common.

---

For `fun`, I would sometimes do:
```ocaml
let foo : int -> int -> int = fun x y ->
  fibonacci_is_a_long_name_to_write x + isn't_it_yes_this_is_a_legal_function_name y
```
but I think more generally I'd put the `fun` on the next line here; might not be the popular version though:
```ocaml
let foo : int -> int -> int =
  fun x y ->
  fibonacci_is_a_long_name_to_write x + isn't_it_yes_this_is_a_legal_function_name y
```
I don't think there is a clearly established style for that one. If there is one, I'd say it's maybe the first one? Although then I wonder what's best in the case where there is both arguments before the `=` and after the `fun`, eg.:
```ocaml
let foo (x : int) : int -> int = fun y ->
  fibonacci_is_a_long_name_to_write x + isn't_it_yes_this_is_a_legal_function_name y
```
Although actually equivalent, I think the last one really means “`foo` is a function of one argument `x : int` that then returns a function from `int` to `int` defined as `fun y -> ...`”, while the others mean “`foo` is a function of two arguments `x : int` and `y : int`”. In the last case, I do think putting `fun y ->` on a new line brings more clarity.

---

Finally, I tend to put `match` on its own line.
```ocaml
let bar x y =
  match x,y with
  | true, false -> "banana"
  | false, true -> "apple"
  | _, _ -> fail_with "I can't think of another fruit"
```
although I have occasionally done Arnaud's version as well for very short things (like in the given example).

aspiwack:
> I think both views are relevant. I mean, in your suggestion, you are transforming the `let f x = res` node from a single one to a multiline one. And since `let` can exist without `in` (toplevel `let`) and `let f x = res in` cannot exist without an expression after, the decision taken by the writer of tree-sitter-ocaml was that `let f = res in expr` leads to the tree

Arglblflbl! Well, let's continue this discussion in another issue, probably. It's at the very least interesting. I still feel that there is something weird in removing that particular newline. But here is not the place for philosophising.

tests/samples/expected/ocaml.ml
aspiwack: I've got one vote today, so I'll share before I say goodnight: this first vote would let both `function` and `fun` hang (honestly, I'm the only person in the world hanging `match`, so I'm not even asking).

torhovland:
@GuillaumeGen Let us know if/when you are expecting an approval of this.

GuillaumeGen:
What I understood from the previous discussion, is that Arnaud is at least expecting to avoid line breaks before fun and function before merging it.
I can do it right now.

aspiwack:
It can be a separate PR, honestly. Whatever's easier for you, I guess.

torhovland (approved):

torhovland:
Merge when you're happy with this.

aspiwack:
I really enjoy how easy it was to fix `function`  and `fun`.

GuillaumeGen:
Yes, using tree-sitter grammar makes it quite rigid when we are dealing with cut in the code which are not the same as the division in nodes (or at least, we do not have the right command set yet to deal efficiently with it).
But when things are simple to do, the result looks very simple and natural. And "simple things are simple to do" is a property I love!

PR #129, opened 2022-12-15
Ocaml support for bound operators

nbacquey:
Adds spacing and line breaks for `let`-like and `and`-like operators.
This allows the following to be correctly formatted:
```
let greetings =
  let (let*) = Option.bind
  and (and*) a_opt b_opt =
    match (a_opt, b_opt) with
    | (Some a, Some b) -> Some (a, b)
    | _ -> None
  in
  let* msg1 = Some "hello "
  and* msg2 = Some "world"
  in
  Some (msg1 ^ msg2)
```

### DIFF
--- a/tests/samples/input/ocaml.ml
+++ b/tests/samples/input/ocaml.ml
@@ -509,3 +509,17 @@ let is_prime n =
     then no_divisor := false;
   done;
   !no_divisor
+
+(* Showcase the usage of operator bindings *)
+let greetings =
+  let (let*) = Option.bind
+  and (and*) a_opt b_opt =
+    match (a_opt, b_opt) with
+    | (Some a, Some b) -> Some (a, b)
+    | _ -> None
+  in
+  let* msg1 =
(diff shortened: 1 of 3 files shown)


tests/samples/input/ocaml.ml
```diff
@@ -495,3 +495,16 @@ let add_as_fun_multiline = fun x ->
...
+
+(* Showcase the usage of operator bindings *)
+let greetings =
+  let (let*) = Option.bind
+  and (and*) a_opt b_opt =
+    match (a_opt, b_opt) with
+    | (Some a, Some b) -> Some (a, b)
+    | _ -> None
+  in
+  let* msg1 = Some "hello "
```
aspiwack, line 507: I would add an example where the

```ocaml
let* x =
  something or some other
in
```

is multiline like this.

nbacquey: done in latest version

GuillaumeGen (approved):

Issue #80, opened 2022-09-30
Check code coverage on CI


torhovland:

Xophmeister:
I've taken an initial stab at this (446b9f6), but it's far from PR-worthy. That said, let me enumerate what I've done to get to this point:

* [This blog post](https://blog.rng0.io/how-to-do-code-coverage-in-rust) recommends using LLVM profiling data to generate the coverage metrics. I was able to do this locally, running the output through [`grcov`](https://github.com/mozilla/grcov) to get a nice report:

  ![image](https://user-images.githubusercontent.com/384987/207893949-cb17bb49-b3c3-4c0f-a5c6-092a0a00a1e1.png)

  To get here, I needed to install the [`llvm-tools-preview` component](https://internals.rust-lang.org/t/llvm-tools-a-new-rustup-component-for-binary-inspection-objdump-nm-size-and-profiling-profdata/7830) for `rustup`...which meant switching to a `rustup`-mediated toolchain. (`binutils` and `grcov` were available in Nixpkgs.)

* The [Rust page](https://nixos.wiki/wiki/Rust) on the NixOS wiki gives an example of how to use `rustup`, rather than Nixpkgs. However, my Nix skills are very basic and this is a bit beyond me at the moment! To complicate matters further, Topiary uses Nix Flakes, which are further off my radar. That said, [@torhovland's blog post](https://www.tweag.io/blog/2022-09-22-rust-nix/) was useful in pointing me in appropriate directions to continue...

* Topiary is using [Crane](https://github.com/ipetkov/crane). This doesn't expose any API for `rustup`, but it does provide support for [Tarpaulin](https://github.com/xd009642/tarpaulin), for code coverage. My Nix skills were just about enough to get this working, in the aforementioned commit! (It gives a similar result, but I don't know where/if it dumps a nice report, or something useful to push to, e.g., CodeCov. The documentation suggests there should be an XML file somewhere, but I can't find it :confused:)

* Unfortunately, AIUI, Tarpaulin works by instrumenting the code and then using `ptrace` to assess coverage. It's thus x86-64 Linux only, so I've disabled the `macos-latest` CI build target (which is, presumably, `aarch64-darwin`) and would have the same implication for #79.[^1] That's particularly unsatisfying :disappointed:

**tl;dr** What I would like to do is use LLVM profiling, but that would entail changing the Nix derivation outside of my (current) capabilities. What I've been able to do is use Tarpaulin, but that is necessarily limiting.

[...]

PR #130, opened 2022-12-15
Document generation of code coverage report

Xophmeister:
This PR documents how one can generate an HTML code coverage report using LLVM profiling data. It also provides a script, for Nix users, to do this automatically without (hopefully) polluting their environment.

This goes towards #80, but has not been added to the CI as it flaunts the usual Nix Flake-based distribution. (See that issue for discussion.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /result
 /.envrc
 /.direnv
+*.profraw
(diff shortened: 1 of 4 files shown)


torhovland (approved):
Looking great!

PR #131, opened 2022-12-16
Handle for and while loops

GuillaumeGen:
Add "for", "to" and "while" keywords, as well as indentation related to "do" and "done".

### DIFF
--- a/tests/samples/input/ocaml.ml
+++ b/tests/samples/input/ocaml.ml
@@ -495,3 +495,17 @@ let add_as_fun_multiline = fun x ->
   x
 
 let add_as_fun_one_line = fun x -> x
+
+let sum_of_int n =
+  let res = ref 0 in
+  for i = 1 to n do res := !res + i; done
+
+let is_prime n =
+  let no_divisor = ref true in
+  let i = ref 1 in
+  while !no_divisor && !i * !i < n do
+    incr i;
+    if n mod !i = 0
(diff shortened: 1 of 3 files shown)


tests/samples/input/ocaml.ml
```diff
@@ -495,3 +495,17 @@ let add_as_fun_multiline = fun x ->
   x
 
 let add_as_fun_one_line = fun x -> x
+
+let sum_of_int n =
```
torhovland, line 499: @aspiwack I'll approve this, but some multi-line elements are turned into single-line here, so you might want to take a look too.

torhovland (approved):

tests/samples/input/ocaml.ml
aspiwack: Indeed, there is definitely something fishy here. Even with the most generous interpretation of what a node is, this should not become a single line. Can someone open an issue to track that let-ins seem to attract things upward like this?

(in this particular case, I expect `sum_of_int` to be formatted as written)

GuillaumeGen: I can open an issue about it.

aspiwack: :arrow_right: #134 .

torhovland: This can be merged, then.

PR #134, opened 2022-12-16
Allow line break after 'in' and "while"

### LINKED ISSUE #133: [Ocaml] Multiline let-ins are turned into single one
When one writes 
```
let f x = a in
some_expression
```
topiary currently turns it into
```
let f x = a in some_expression
```

This is not the expected behaviour, as it turns the expression node from a multiline node to a single line one.

**Additional context**
The reason for this is that currently we do allow line breaks after the "in" keyword only for some very specific constructs https://github.com/tweag/topiary/blob/d7f746dd24b15ec47506f9df3a811646d31f148f/languages/ocaml.scm#L437-L453 . This is way too restrictive.

GuillaumeGen:
This addresses #133 by adding the new line after the "in" in let definition rather than having it before some very specific constructs.
Same for "=" and "with", there are already some line breaks added depending of the context (ie ancestor node) in which these symbols occur, and it seems more principled to have it this way and made the incriminated query redundant.

@aspiwack I know that you do not like the formatting of `add_three_lines`, but as I mentioned in https://github.com/tweag/topiary/pull/127#issuecomment-1351770298 I do not see how to format it differently with the current set of command.
Of course, adding a new command would not be hard at all, but I do not see it as a priority.

Closes #133 .

### DIFF
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -446,10 +446,11 @@ let topological_sort deps =
             else
               raise
               @@ Files_legacy.Files_error (ObjectFileNotFound (mk_mident node))
-          | _ -> assert false
-      in node::List.fold_left (explore (node::path))
+      in
+      node::List.fold_left (explore (node::path))
         visited (List.map Files_legacy.get_file edges)
-  in List.rev @@ List.fold_left (fun visited (n, _) -> explore [] visited n) [] graph
+  in
+  List.rev @@ List.fold_left (fun visited (n, _) -> explore [] visited n) [] graph
 
 (* The even and odd functions assume that their argument is non-negative. *)
 let rec odd = function
@@ -495,6 +496,14 @@ let add_multiline x =
   res
 
 let add_one_line x = let res = x + x in res
+
+let add_two_lines x =
+  let res = x + x in
+  res
+
+let add_three_lines x =
+  let res = x + x in
+  res
 
 let add_as_fun_multiline = fun x ->
   x
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -446,6 +446,7 @@ let topological_sort deps =
             else
               raise
               @@ Files_legacy.Files_error (ObjectFileNotFound (mk_mident node))
+          | _ -> assert false
       in
       node::List.fold_left (explore (node::path))
         visited (List.map Files_legacy.get_file edges)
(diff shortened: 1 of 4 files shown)


aspiwack:
(edited the initial message to add an auto-close)

Looks like no disaster happened by making the query more general. Good. Maybe this was an artefact of a time where we didn't have `@do_nothing`.

>  Of course, adding a new command would not be hard at all, but I do not see it as a priority.

Probably not. Though just for the sake of the discussion, what would this new command be? (I kind of what to have an idea of how ad hoc it would be)

GuillaumeGen:
It has to be discussed. A simple thing I can think of is to have a query like
```
(let_expression
  (value_definition
    (let_binding
      "=" @append_spaced_anchoredline
   ) @append_spaced_anchoredline
  )
 .
 'in" @anchor
)
```
where `spaced_anchoredline` would be a new line if the node is multiline, or if it is not on the same line as the anchor and would be a space otherwise.

GuillaumeGen:
After https://github.com/tweag/topiary/issues/133#issuecomment-1355133182 I indeed think that it makes more sense to add a soft linebreak after every "in" and "with" rather than trying to move from an over-restrictive list of next elements to an over-restrictive list of ancestor nodes.
The last commit goes in this direction.

tests/samples/expected/ocaml.ml
```diff
@@ -446,9 +446,11 @@ let topological_sort deps =
             else
               raise
               @@ Files_legacy.Files_error (ObjectFileNotFound (mk_mident node))
-      in node::List.fold_left (explore (node::path))
+      in
+      node::List.fold_left (explore (node::path))
         visited (List.map Files_legacy.get_file edges)
-  in List.rev @@ List.fold_left (fun visited (n, _) -> explore [] visited n) [] graph
+  in
+  List.rev @@ List.fold_left (fun visited (n, _) -> explore [] visited n) [] graph
```
torhovland, line 453: @aspiwack I'm fine with this PR if you're OK with this formatting (line break and no indenting).

aspiwack: Yes, it's the most common style (there are many though. But this is the one I use, so it's also the best style).

aspiwack (approved):

tests/samples/expected/ocaml.ml
yannham: Oh no, it's not the one I'm using :cry: out of curiosity, what is your criterion for keeping the `in` at the end (as below in `let res = x + x in`) versus introducing a new line? Do you do that as soon as the let-binding spans more than one line?

aspiwack: @yannham 

The design is that each form has two different shapes: single-line or multi-line. The selection between single-line and multi-line depends on the programmer's input: if I wrote something on several line, Topiary will use the multi-line shape, if I wrote it on one line, it will use the single-line shape.

The current style for `let` is

```ocaml
let x =
  …
in
```

and

```ocaml
let x = … in
```

The fact that

```ocaml
let x = foo
in
```

is considered single-line rather than multiline is a technical limitation due to details in the grammar.

yannham: I see, this makes sense. Thanks!

Issue #135, opened 2022-12-18
[OCaml] Missing space (or newline) after `type ... and`


Niols:
Last version on main (d7f746dd24b15ec47506f9df3a811646d31f148f) will format
```
type t = int
and s = float
```
as
```
type t = intand s = float
```
and then crash during the idempotence check. It should be
```
type t = int and s = float
```
or as it was initially.

nbacquey:
This issue has been closed by #136

PR #136, opened 2022-12-19
Handle mutually recursive types

GuillaumeGen:
Solves #135

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -406,6 +406,7 @@
 
 ; Always put softlines before these:
 [
+  "and"
   "done"
   "end"
   (else_clause)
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -433,6 +433,10 @@
   "and" @prepend_spaced_softline
 )
 
+(type_definition
+  "and" @prepend_spaced_softline
```
torhovland, line 437: Can't we simply always prepend a softline before `and`?

torhovland (approved):

languages/ocaml.scm
GuillaumeGen: I think we can. Let me have a check in the grammar of Ocaml before doing this modification.

GuillaumeGen: A search for `and` in https://github.com/tree-sitter/tree-sitter-ocaml/blob/master/ocaml/grammar.js tells me that and can appear in mutually recursive `value_definition`, `type_definition`, `module_definition`, `module_type_definition` and `class_definition`. I thunk we want a soft break in all those cases.

PR #142, opened 2022-12-20
[OCaml]: Support for quoted strings

nbacquey:
Allows the following to be formatted correctly:
```
let quoted_string =
  {|Hello|}
print_endline quoted_string
```

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -7,6 +7,7 @@
 (
   [
     (character)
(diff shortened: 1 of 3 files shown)


GuillaumeGen (approved):
Looks fine.
I would love to understand why adding `(quoted_string)` in the leaves does not prevent us from explore the inner node, but this question should not prevent this fix to be integrated.

PR #146, opened 2022-12-21
[OCaml] Support object-oriented programming

nbacquey:
Allows correct formatting of object-oriented code, like the following:
```
module Inner1 = struct
  class ['a] pushable =
    object (self)
      val mutable list = ([]: 'a list)

      method push x =
        list <- x::list
    end

  class virtual ['a] poppable =
    object (self)
      inherit ['a] pushable

      (* unsafe implementation *)
      method pop =
        let result::rest = list in
        list <- rest;
        Some result
    end
end

module Inner2 = struct
  class type ['a] stack =
    object
      inherit ['a] Inner1.poppable
    end

  class ['a] stack_impl : ['a] stack =
    object
      inherit ['a] Inner1.poppable as super

      (* safe implementation *)
      method pop =
        match list with
        | _::_ -> super#pop
        | _ -> None
    end
end

let (Some 2) =
  let my_stack = new Inner2.stack_impl in
  my_stack#push 1;
  my_stack#push 2;
  my_stack#pop
```
Note that I'm not very familiar with OCaml's object-oriented capabilities. Feel free to point out if there's some oop constructions that I've missed.

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -14,9 +14,15 @@
 
 ; Allow blank line before
 [
+  (class_definition)
+  (class_type_definition)
   (comment)
   (exception_definition)
   (external)
+  (inheritance_definition)
+  (inheritance_specification)
+  (instance_variable_definition)
+  (method_definition)
   (module_definition)
   (module_type_definition)
   (open_module)
@@ -83,6 +89,7 @@
   "and"
   "as"
   "assert"
+  "class"
   "downto"
   "else"
   "exception"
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -218,6 +246,21 @@
...
+  "("* @do_nothing
+  .
+  "new" @prepend_space
+)
+(
+  "("* @do_nothing
+  .
+  "object" @prepend_space
+)
+(
```
torhovland, line 259: Duplicate.

torhovland (approved):

languages/ocaml.scm
nbacquey: Well spotted, thanks

PR #148, opened 2022-12-21
Bash support (POC)

Xophmeister:
This PR adds vacuous Bash support. That is, it brings in all the dependencies and contains the necessary changes to the codebase to support Bash, but the query file for Bash contains no formatting directives and the sample files (input and expected) just have their shebang lines; these will be expanded in subsequent PRs.

(I also took the opportunity to update the input vs. expected test runner to use the language detection code.)

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -10,6 +10,7 @@ pub enum Language {
     Ocaml,
     Rust,
     Toml,
+    Bash,
 }
 
 // NOTE This list of extension mappings is influenced by Wilfred Hughes' Difftastic
@@ -37,6 +38,7 @@ const EXTENSIONS: &[(&str, &[&str])] = &[
     ("ocaml", &["ml"]),
     ("rust", &["rs"]),
     ("toml", &["toml"]),
+    ("bash", &["sh", "bash"]),
 ];
 
 impl Language {
@@ -46,6 +48,8 @@ impl Language {
             "ocaml" => Ok(Language::Ocaml),
             "rust" => Ok(Language::Rust),
             "toml" => Ok(Language::Toml),
+            "bash" => Ok(Language::Bash),
+
             _ => Err(FormatterError::Query(
                 format!("Unsupported language specified: '{s}'"),
                 None,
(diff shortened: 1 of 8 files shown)


src/language.rs
```diff
@@ -10,6 +10,7 @@ pub enum Language {
     Ocaml,
     Rust,
     Toml,
+    Bash,
```
torhovland, line 13: You can keep all these lists of languages ordered.

tests/sample-tester.rs
```diff
@@ -12,22 +12,17 @@ fn input_output_tester() {
...
-        let language = match extension {
-            "json" => Language::Json,
-            "ml" => Language::Ocaml,
-            "rs" => Language::Rust,
-            "toml" => Language::Toml,
-            _ => panic!("File extension {} not supported.", extension),
-        };
+        let input_path = file.path();
+        let language =
+            Language::detect(input_path.to_str().unwrap()).unwrap_or_else(|err| panic!("{err}"));
```
torhovland, line 18: This `unwrap_or_else` could have been a simple `unwrap`, right?

torhovland (approved):

tests/sample-tester.rs
Xophmeister: Probably :sweat_smile:

I'm probably wrong, but I thought `unwrap`'ing an arbitrary `Err` would just give a generic panic. Whereas, as in this case, when the error type implements `Display`, I can use `unwrap_or_else` to get a nicer error message... Or am I wrong and `unwrap` is also `Display`-aware?

src/language.rs
Xophmeister: Reordered in 9eb72856e5694

tests/sample-tester.rs
Xophmeister: It looks like the `unwrap` panic uses the error type's `Debug` implementation, rather than `Display`. For example:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error', src/main.rs:14:7
```

PR #149, opened 2022-12-22
[OCaml] miscellaneous improvements on module/functors formatting

nbacquey:
This will allow the following to be formatted correctly:
```
module type T1 = sig
  type t = private int64
end

module type T2 = sig end

module M: T2 = struct end

module F (X: T1) (Y: T1 with type t := X.t) = struct
  module type S = sig
    type t = X.t

    val zero : t

    val succ : t -> t

    include T2
  end

  let zero = 0

  let succ n = n + 1

  include T2
end
```

### DIFF
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -606,6 +606,9 @@ let (Some 2) =
   my_stack#push 2;
   my_stack#pop
 
-(* Module with a private type *)
+(* Some modules and functors *)
 module type M = sig type t = private int64
+end
+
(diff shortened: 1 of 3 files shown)


tests/samples/expected/ocaml.ml
```diff
@@ -471,6 +471,7 @@ module type Printer = sig
     element printer and [sep] as separator between elements. *)
   val print_list : string -> 'a printer -> 'a list printer
   val print_name : name printer
+
```
nbacquey, line 474: This blank line appears in the `input` file, but was deleted in the previous version

torhovland (approved):

Issue #138, opened 2022-12-19
Bash formatting support


Xophmeister:
**Is your feature request related to a problem? Please describe.**
AFAIK, there is no de facto Bash/shell script formatter. It is _very_ common to come across horrendously formatted shell scripts!

**Describe the solution you'd like**
Bash support in Topiary. (Bash, for now; perhaps POSIX `sh`, `zsh`, etc. in the future.)

**Describe alternatives you've considered**
n/a

**Additional context**
Having Bash support, alongside OCaml, JSON and TOML, in the first -- or an early -- release would be a good demonstration of both the general-ness of the engine and the potential for the project.

(Grammar: https://github.com/tree-sitter/tree-sitter-bash)

---

* [x] Update Rust code
* [x] Commands, command lists and command pipelines
* [x] Conditionals
* [x] Switch statements
* [x] Loops
* [x] Redirection and heredocs
* [x] Compound statements
* [x] Variables (declaration and expansion)
* [x] Command and process substitution
* [x] Subshells
* [x] Functions
* [x] (Opinionated) line spacing
* [x] Comments

Niols:
I'm not sure what you mean by “de facto” Bash formatters, but there do exist some Shell formatters, including for instance [shfmt](https://github.com/mvdan/sh). Don't get me wrong, I'm super excited to see it included in Topiary, I just wouldn't want someone somewhere to claim there were no other Shell formatters!

Xophmeister:
> I'm not sure what you mean by “de facto” Bash formatters, but there do exist some Shell formatters, including for instance [shfmt](https://github.com/mvdan/sh).

Good point. What I mean is that -- in my experience, at least -- it's not common for formatters to be applied to Bash scripts and no particular formatter has a significant mindshare, not that they don't exist. (Compared with, e.g., linters, where many would immediately jump to Shellcheck.)

Niols:
That makes sense and I completely agree. Hopefully, that is something Topiary could bring to the table: if it's already a dependency of your project, why not make it format your Shell files as well?

Xophmeister:
n.b., Line spacing is a bit messed up, but it's "good enough"; see #203.

PR #155, opened 2022-12-23
[Bash] Step 1: Basic commands and conditionals

Xophmeister:
This PR introduces support for basic Bash commands and conditional blocks, with appropriate samples for integration testing. This is in aid of #138.

#### To Do :bulb:
* [x] One command per line, in the simple case (i.e., not sequences, pipelines, etc.)
* [ ] ~Correct spacing for binary expressions in test commands.~
* [x] Generalisation (if possible) of `if_statement` and `elif_clause` nodes.

#### Questions :question:
* [x] Can capture names be applied to field queries? (Yes :heavy_check_mark:)

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -21,24 +21,29 @@
   "else"
   "fi"
   (string)
-  (command)
   (test_command)
 ] @prepend_space
 
 ;; Commands
 
-; Space between command line arguments
-(command
-  argument: _ @append_space @prepend_space
-)
-
-; FIXME One command per line, modulo many exceptions:
+; One command per line, modulo many exceptions:
 ; * Strings of && and ||
 ; * Pipelines
 ; * Subshells
 ; * As arguments to if, while, for, etc.
 ; * Others?...
-;(command) @append_input_softline
+;
+; FIXME Can this be generalised, or does *every* context need to be
+; individually enumerated?...
+(program (command) @prepend_spaced_softline @append_hardline)
+(if_statement _ "then"(command) @prepend_spaced_softline)
+(elif_clause _ "then" (command) @prepend_spaced_softline)
+(else_clause (command) @prepend_spaced_softline)
+
+; Space between command line arguments
+(command
+  argument: _ @append_space @prepend_space
+)
 
 ;; Operators
 
@@ -68,7 +73,7 @@
 ; Keep the "if" and the "then" on the same line,
 ; inserting a delimiter when necessary
 (if_statement
-  (_) @append_delimiter
+  (_) @prepend_space @append_delimiter
   ";"* @do_nothing
   "then"
 
@@ -77,7 +82,7 @@
 
 ; FIXME Can this be generalised from the above?
 (elif_clause
-  (_) @append_delimiter
+  (_) @prepend_space @append_delimiter
   ";"* @do_nothing
   "then"
 
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -36,7 +36,7 @@
 ; FIXME Can this be generalised, or does *every* context need to be
 ; individually enumerated?...
 (program (command) @prepend_spaced_softline @append_hardline)
-(if_statement _ "then"(command) @prepend_spaced_softline)
+(if_statement _ "then" (command) @prepend_spaced_softline)
 (elif_clause _ "then" (command) @prepend_spaced_softline)
 (else_clause (command) @prepend_spaced_softline)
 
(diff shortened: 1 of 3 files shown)


Xophmeister:
RE Correct spacing for binary expressions: It turns out `binary_expression` is not in my CST, which explains why it isn't working as expected.

```
{Node test_command (13, 6) - (13, 14)} - Named: true
  {Node (( (13, 6) - (13, 8)} - Named: false
  {Node word (13, 8) - (13, 12)} - Named: true
  {Node )) (13, 12) - (13, 14)} - Named: false
```

I'm not sure why this is the case; it's [clearly in the grammar](https://github.com/tree-sitter/tree-sitter-bash/blob/4488aa41406547e478636a4fcfd24f5bbc3f2f74/grammar.js#L372-L389).

Xophmeister:
253680a is _almost_ correct, apart from the extra hardline after the final `fi`. I can solve that by removing the `@append_hardline` from: https://github.com/tweag/topiary/blob/253680a20228a8465ca90bd43efa9faa146a993b/languages/bash.scm#L92-L99

However, then there's a _missing_ hardline between the two `if`-blocks.

Nonetheless, opening for review to assess the validity of my queries and approach...

torhovland:
Repeated hardlines _should_ be OK, because post-processing will clean them out. But it is possible that fails because something unexpected has found its way in-between the hardlines.

torhovland:
And it certainly makes sense to me that the `if` statement needs to end with a hardline.

languages/bash.scm
```diff
@@ -1,2 +1,114 @@
...
+
+; Allow blank line before
+; TODO Add directives here...
+[
+  (comment)
+  (if_statement)
+] @allow_blank_line_before
+
+; Prepend spaces
+; (Not "surround", to avoid extraneous spaces between delimiters)
```
torhovland, line 15: You should not worry about that here. That is the reponsibility of the post-processing.

languages/bash.scm
```diff
@@ -1,2 +1,114 @@
...
+;; Commands
+
+; One command per line, modulo many exceptions:
+; * Strings of && and ||
+; * Pipelines
+; * Subshells
+; * As arguments to if, while, for, etc.
+; * Others?...
+;
+; FIXME Can this be generalised, or does *every* context need to be
```
torhovland, line 36: Maybe simply `(command) @prepend_spaced_softline`?

languages/bash.scm
```diff
@@ -1,2 +1,114 @@
...
+; individually enumerated?...
+(program (command) @prepend_spaced_softline @append_hardline)
+(if_statement _ "then"(command) @prepend_spaced_softline)
```
torhovland, line 39:
```suggestion
(if_statement _ "then" (command) @prepend_spaced_softline)
```

languages/bash.scm
```diff
@@ -1,2 +1,114 @@
...
+; inserting a delimiter when necessary
+(if_statement
+  (_) @prepend_space @append_delimiter
+  ";"* @do_nothing
+  "then"
+
+  (#delimiter! ";")
+)
+
+; FIXME Can this be generalised from the above?
```
torhovland, line 83: Maybe if you say "anything that can have `then` in it":

```
(_
  (_) @prepend_space @append_delimiter
  ";"* @do_nothing
  "then"

  (#delimiter! ";")
)
```

torhovland (approved):

languages/bash.scm
Xophmeister: Nice :+1:

languages/bash.scm
Xophmeister: This breaks, e.g., `if`-clauses. It will turn:

```bash
if something
```

into:

```bash
if
something
```

The same will happen in the (many) other contexts where commands are embedded into some clause.

languages/bash.scm
Xophmeister: There's/there'll be post-processing? To what degree is it envisioned to operate (e.g., simple whitespace trimming would deal with the extra `\n`s, but targetting delimiters would need to be grammar-aware, etc.)?

ErinvanderVeen (approved):

languages/bash.scm
Xophmeister: This?

https://github.com/tweag/topiary/blob/c36d4a2253f337e1a28d497826a84754b8d833f6/README.md?plain=1#L116-L123

What I'm trying to achieve by just prepending spaces is `if foo; then`, rather than `if foo ; then`, for example.

torhovland: Understood. But `if`, `then`, and many more keywords can still be surrounded by spaces.

languages/bash.scm
torhovland: Got it.

PR #158, opened 2023-01-02
[OCaml] spacing before `::` when preceded by labeled argument

nbacquey:
```
let into_list ~value = my_id ~value :: []
```

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -529,6 +529,24 @@
   ">" @prepend_space
 )
 
+; Keep spacing around the list constructor if it is preceded by
+; a labeled argument, avoiding syntax errors.
+;
+; This is syntactically correct:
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -473,6 +473,25 @@
...
+; This is syntactically correct:
+;   foo ~arg :: []
+; This is not:
+;   foo ~arg::[]
+;
+; While both are correct if the argument is not labeled
+(
+  (application_expression
+    (_)*
+    .
```
torhovland, line 488: Aren't these two lines redundant?

torhovland (approved):

languages/ocaml.scm
nbacquey: They are indeed ; it's fixed in the new version

PR #159, opened 2023-01-02
[OCaml] spacing in object signatures

nbacquey:
```
let obj_id (obj : < .. >) = obj
let obj_with_unit_id (obj : < nothing: unit; .. >) = obj
let obj_with_only_unit_id (obj : < nothing: unit >) = obj
```

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -473,6 +473,18 @@
   (_)
 )
 
+; Keep spacing between pointy brackets when used in object types.
+;
+; This is syntactically correct:
+;   let obj_id (obj : < .. >) = obj
+; This is not:
+;   let obj_id (obj : <..>) = obj
+(object_type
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -473,6 +473,20 @@
...
+;
+; This is syntactically correct:
+;   let obj_id (obj : < .. >) = obj
+; This is not:
+;   let obj_id (obj : <..>) = obj
+(object_type
+  "<" @append_space
+  .
+  (_)*
+  .
```
torhovland, line 486: Aren't these lines redundant?

torhovland (approved):

languages/ocaml.scm
nbacquey: This is redundant indeed. Fixed in latest commit

PR #160, opened 2023-01-02
draft:[OCaml] support of ppx syntax

nbacquey:
```
let _ = [%ext fun () -> ()]

let _ = 12 [@attr pl]

let _ = "some string" [@@attr pl]

[@@@attr pl]

let%ext _ = ()
```

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -89,64 +89,69 @@
 )
 
 ; Surround spaces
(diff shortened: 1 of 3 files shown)


nbacquey:
Some of this pr may be redundant with #144 
Some work may be needed to merge both PRs

torhovland:
How do you suggest we proceed with these?

Issue #162, opened 2023-01-04
[OCaml] Cubic time complexity when processing pattern matching with large number of cases


nbacquey:
**Describe the bug**
Topiary takes an unreasonable time to process the following function:
```
let big_match = function
  | 0 -> 0
  ...
  [repeat n times]
  ...
  | 0 -> 0
```
I've run topiary on a few files named `ocaml_n.ml`, which contain `n` repetitions of the pattern `| 0 -> 0`, as well as a real-life, 1200 lines long OCaml file containing a large pattern matching.
```
$ time ( topiary -s -i ocaml_real.ml > /dev/null 2>&1 )
real    0m17,029s
user    0m17,000s
sys     0m0,028s

$ time ( topiary -s -i ocaml_512.ml > /dev/null 2>&1 )
real    0m6,669s
user    0m6,661s
sys     0m0,008s

$ time ( topiary -s -i ocaml_1024.ml > /dev/null 2>&1 )
real    0m56,513s
user    0m56,489s
sys     0m0,021s

$ time ( topiary -s -i ocaml_2048.ml > /dev/null 2>&1 )
real    8m2,791s
user    8m2,752s
sys     0m0,009s
```
The results suggest there is an cubic behavior somewhere: doubling the number of cases multiplies the time by 8. Moreover, activating `RUST_LOG=debug` on `ocaml_2048.ml` produced a 540MB log file, containing over 2M lines.

**To Reproduce**
- retrieve the `ocaml_n.ml` files from [my branch](https://github.com/tweag/topiary/tree/nb/ocaml_time_experiments) , or produce them independently,
- run `time ( topiary -s -i ocaml_n.ml > /dev/null 2>&1 )`
- observe the unreasonable times

**Expected behavior**
We shouldn't take more than 0.1s to process a real-life, 1200 lines long OCaml file.

**Environment**
 - Ubuntu 22.04 + main on 2023-01-04 (https://github.com/tweag/topiary/commit/3a21fb5d982b77c75a7a4d4fdd859bccdac2fa34)

**Additional context**
I'm still trying to figure out if we have a query that produces an unreasonable number of nodes, a process that is unreasonably slow, or both.

torhovland:
It would be nice to see the flamegraph. If you can't get that working (by installing `perf`), you can send me a sample file.

nbacquey:
Here is the flamegraph for `ocaml_1024.ml`:
[...]

PR #163, opened 2023-01-04
[Ocaml] Add an anchor for performance

GuillaumeGen:
This seems to solve the issue mentioned in #162

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -623,6 +623,7 @@
 ;
 (
   "|"* @do_nothing
+  .
   (match_case) @prepend_spaced_softline
 )
 


torhovland (approved):
Smallest PR ever 😆

GuillaumeGen:
On my computer, times drop from 
real    1m12,987s
user    1m12,946s
sys     0m0,036s

to
real    0m0,329s
user    0m0,313s
sys     0m0,018s

for the 1024 line case.

aspiwack:
Out of curiosity: why does this PR work?

aspiwack:
Thank you, it's very clear now. It's an easy trap to fall into, I think. A gotcha to keep in mind.

GuillaumeGen:
Just for completeness if anyone finds this issue because it got strike by a similar problem.
Once Nicolas understood that the performance issue stroke us when dealing with a huge pattern matching, I looked for the queries related to pattern matching and modified them to make them print a different number each. For instance:
```
(
  "|"* @do_nothing
  (match_case) @prepend_spaced_softline @append_delimiter
  (#delimiter! "2")
)
```
and then one can run `topiary --input-file=match_1024.ml --skip-idempotence` and see that there are hundreds of 2's, so that the issue comes from this query.

torhovland:
I have on occasion also misused `@append_delimiter` to debug which pattern is getting applied in a particular situation 😀

PR #167, opened 2023-01-09
Draft Nickel formatter

GuillaumeGen:
This is a very drafty version of the formatter for Nickel.
The main purpose of it is to have people complaining about the output to know what the formatter should do.

The current version tends to add new lines and spaces quite often (and is not always super consistent about spaces).

### DIFF
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -1,0 +1,193 @@
+; Configuration
+(#language! nickel)
+
+[
+  (static_string)
+  (str_chunks_single)
+] @leaf
+
+; Allow blank line before
+[
+  (comment)
+] @allow_blank_line_before
+
+(comment) @prepend_input_softline
+
+(
+  (comment) @append_input_softline
+  .
+  [ "," ";" ]* @do_nothing
+)
+
+; Surround spaces
+; A space is put after, and before keywords.
+; It is also put before and after "|", ":" and "?" separating annotation from the annotated object.
+(
+  [
+    "if"
+    "then"
+    "else"
+    "forall"
+    "in"
+    "let"
+    "switch"
+    "null"
+    "true"
+    "false"
+    "fun"
+    "import"
+    "default"
+    "doc"
+    "|"
+    ":"
+    "?"
+    "="
+    "rec"
+    "{"
+    "}"
+    "Array"
+    "Dyn"
+    "Num"
+    "Bool"
+    "Str"
+    "->"
+  ]
+) @prepend_space @append_space
+
+(let_expr
+  "=" @append_spaced_softline @append_indent_start
+  .
+  t1: (_) @append_indent_end @append_spaced_softline
+  .
+  "in" @prepend_spaced_softline @append_spaced_softline
+)
+
+(fun_expr
+  "=>" @append_spaced_softline @append_indent_start
+  (_) @append_indent_end
+  .
+)
+
+(fun_expr
+  (pattern) @append_space
+)
+
+(switch_expr
+  "{" @append_spaced_softline @append_indent_start
+  "}" @prepend_indent_end @prepend_spaced_softline @append_spaced_softline
+)
+
+(switch_expr
+  "," @append_spaced_softline
+)
+
+(ite_expr
+  "then" @prepend_spaced_softline @append_spaced_softline @append_indent_start
+  t1 : (term) @append_spaced_softline @append_indent_end
+  "else" @append_indent_start @append_spaced_softline
+  t2: (term) @append_indent_end
+)
+
+(forall
+  "." @append_spaced_softline @append_indent_start
+  (_) @append_indent_end
+  .
+)
+
+(forall
+  (ident) @append_space
+  .
+  (ident)
+)
+
+(uni_record
+  "{" @append_spaced_softline @append_indent_start
+  "}" @prepend_indent_end @prepend_spaced_softline
+)
+
+(uni_record
+  "," @append_spaced_softline
+)
+
+(uni_record
+  ";" @append_spaced_softline
+)
+
+; We want the same rule for arrays as for records,
+; but array does not have a specific syntactic class, so we use the one of atoms.
+(atom
+  .
+  "[" @append_spaced_softline @append_indent_start
+  (term)
+  "]" @prepend_indent_end @prepend_spaced_softline
+  .
+)
+
+(atom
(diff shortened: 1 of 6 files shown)


thufschmitt:
> See #167

Unless you're trying to trick us into an infinite recursion, I guess you meant https://github.com/tweag/topiary/issues/165 :stuck_out_tongue:

aspiwack:
Indeed.

languages/nickel.scm
```diff
@@ -0,0 +1,193 @@
+; Configuration
+(#language! nickel)
+
+[
+  (static_string)
+  (str_chunks_single)
+] @leaf
```
torhovland, line 7: For consistency we should add the same comment here about `leaf` that all the other query files include.

languages/nickel.scm
```diff
@@ -0,0 +1,193 @@
...
+  "}" @prepend_indent_end @prepend_spaced_softline @append_spaced_softline
+)
+
+(switch_expr
+  "," @append_spaced_softline
+)
+
+(ite_expr
+  "then" @prepend_spaced_softline @append_spaced_softline @append_indent_start
+  t1 : (term) @append_spaced_softline @append_indent_end
```
torhovland, line 86:
```suggestion
  t1: (term) @append_spaced_softline @append_indent_end
```

languages/nickel.scm
```diff
@@ -0,0 +1,193 @@
...
+)
+
+(destruct
+  "," @append_spaced_softline
+)
+
+[
+  (interpolation_start)
+  (interpolation_end)
+] @prepend_space @append_space
```
torhovland, line 142: This should go in the surround space section near the top.

languages/nickel.scm
```diff
@@ -0,0 +1,193 @@
...
+  "|>"
+  "<"
+  "<="
+  ">"
+  ">="
+  "=="
+  "!="
+  "&&"
+  "||"
+] @prepend_space @append_space
```
torhovland, line 170: This should go in the surround space section near the top.

tests/samples/expected/nickel.ncl
```diff
@@ -0,0 +1,130 @@
...
+  in
+  let security =
+    import "security.ncl"
+  in
+  server & security & { firewall.enabled = false } ,
+  merge_server = {
+    server.host.ip = "182.168.1.1",
+    server.host.port = 80,
+    server.host.name = "hello-world.net",
+  } ,
```
torhovland, line 62: Do we want space before comma here?

torhovland (approved):
The query file is readable enough for now, but when it becomes more complicated, we'll need comments with sample code that illustrate what the patterns do.

tests/samples/expected/nickel.ncl
GuillaumeGen: No

GuillaumeGen:
[bla1.txt](https://github.com/tweag/topiary/files/10413266/bla1.txt)

Here is the result of the formatting on the "benches" folder.
There is a file which is failing to parse, due to the fact that an identifier is called `match` which is also a keyword.
I do not know if this is a feature of Nickel or if it is intended to fail to parse and this file was created before the renaming of `switch` in `match`.

Regarding the formatting, I think that the main issue is the let-ins which are all formatted as 
```
let x = 
  ...
in
t
```
The reason for it is that the `let_expr` nodes have a huge number of children, all at the same level, making it not practical to define "single line" and "multiline" situation.
Since we are in a quite different position compared to the other grammar, since Tweag is maintaining the grammar, I would suggest to change slightly the grammar to make it easier to handle.

yannham:
> There is a file which is failing to parse, due to the fact that an identifier is called match which is also a keyword.
I do not know if this is a feature of Nickel or if it is intended to fail to parse and this file was created before the renaming of switch in match.

Ah, yes, this must have been done before the change, but it's sad that the CI doesn't pick on such errors (benches are too costly, but we could at least check for syntax errors). Let me fix that.

Otherwise, it's looking good for a first draft :slightly_smiling_face: there seems to be issues with indenting of `if-else` inside the mantis benchmark though. Typically (L963 of bla.txt):

```
{
            Name = tName,
            Driver = t.driver,
            Config = t.config,
            Env = t.env,
            KillSignal = t.kill_signal,
          } & ( if t.kill_timeout != null
          then
            {
              KillTimeout = time.ParseDuration (t.kill_timeout)
            }
          else
            { } ) & {
            Affinities = array.map ( fun { attribute, value, operator, weight, .. } =>
              {
                LTarget = attribute,
                RTarget = value,
                Operand = operator,
                Weight = weight,
              } ) t.affinities,
# etc..
```

I find this part confusing for example:

```nickel
       else
            { } ) & {
            Affinities = array.map ( fun { attribute, value, operator, weight, .. } =>
              {
                LTarget = attribute,
```

I guess it's correct in the sense that the `else` ends, we merge with a new record, and the indentation of the content of the record turns out to be 2 spaces, but it looks like the merging and the definition of `Affinities` are part of the `else` branch. Maybe just inserting a new line after a if-then-else (unless it is on one line only) could be sufficient?

yannham:
@ErinvanderVeen is someone else picking up the work of @GuillaumeGen ? It seems to be in a reasonable state to be merged (I can't judge of the code, but regarding the announced scope of the PR and what it does provides).

ErinvanderVeen:
@yannham Yes! We have two eager people excited to start working on the Nickel formatter when they finish their current formatters (Bash and OCaml). We've added Nickel support to milestone v0.1 and so we won't release until it is added.

yannham:
@ErinvanderVeen glad to hear that! Topiary started as your investigation of a formatter for Nickel, and now the circle is complete :smile:

GuillaumeGen:
I implemented a workaround for @yannham suggestion.
Yann suggested to break a line after the `if_then_else` expression. This solution works but it creates a quite annoying situation putting quite often a comma alone on its line. There are currently no solution to add a line break but before a comma, since comma is not a direct sibling of the `if_then_else` expression, it is the next leaf, but it is less deep in the AST.
The solution I propose is to add a softline after the infix operation `&`. This makes the output more readable, but I do not know the nickel syntax (outside of the fact that I have read the tree-sitter grammar) so it might very well be the case that another operator should induce line break just like the `&`.

Current output is 
[bla1.txt](https://github.com/tweag/topiary/files/10549934/bla1.txt)

PR #169, opened 2023-01-11
[OCaml] support for `.mli` files

### LINKED ISSUE #164: [OCaml] Discussion about proper integration of `.mli` files
There is almost nothing to do to make topiary able to process `.mli` (OCaml interface) files.

Almost nothing, but still something: [tree-sitter-ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) has a separate grammar for `.mli` files, which we don't load at the moment. It is a small task to add support for `.mli` files as a completely separate language, which I have done in [this branch](https://github.com/tweag/topiary/tree/nb/ocaml_mli). I've called this "new" language `ocamli`.

However, there is a caveat: the query file we use for `.ml` files works perfectly fine for `.mli` files as well, except that the configuration line in the header must be `(#language! ocamli)` instead of `(#language! ocaml)` (to work properly with the `--query` cli option).

Currently in my branch, I've just copied `ocaml.scm` into `ocamli.scm`, and changed the header. I don't like this solution though, because it introduces a lot of code duplication. So far, I can see two alternatives:
1. make `ocamli.scm` a symlink to `ocaml.scm`.
2. introduce a bit of logic in the Rust code, don't make a separate `ocamli` language, allow to choose between the two parsers depending on whether the file is `.ml` or `.mli`, but always use the same query file.

Neither option are perfectly satisfactory, as they both forego the support of the `--query` option for `.mli` files, unless we introduce some logic of the form "if the first parser fails, try the second one".

What are your opinions on the matter?

**Environment**
- Ubuntu 22.04 + `main` on 2023-01-04 (https://github.com/tweag/topiary/commit/f6608f3e4308d67a4c7cdc32a216e353b4c40ada)

nbacquey:
### Modifications to language processing

This PR changes the way languages are processed in Topiary.

To properly process an input stream, we need two things:
- a grammar, to produce a syntax tree from the input,
- a query file, to format this syntax tree.

Additionally, we use the intermediate notion of `Language`, which is represented by an enum type internally.

The CLI for Topiary ensures we get at least one of the following arguments:
- `--language`, to specify the language of the input stream,
- `--query`, to specify which query file to use,
- `--input-file`, to specify a file from which to read the input.

If `--input-file` is missing, Topiary reads from `stdin`.

We follow the following steps to determine the grammar and the query:

First, we determine the language:
- if the `--language` argument is present, use it,
- else if the `--input-file` is present, guess the language from the file extension,
- else, use the language specified in the query file.

Then the language determines the grammar(s). Indeed, the same language can have multiple grammars: for instance OCaml has separate grammars for the interface (`.mli`) and the implementation (`.ml`) files. We test all grammars associated with the language, until one succeeds or all fail.

The query file is then determined as follows:
- if the `--query` argument is present, use it,
- else automatically deduce the query file from the language.

This PR consolidates all operations regarding languages and their transformation into grammars or query files in the `language.rs` file.

### Support for OCaml interface files

This PR adds support for interface files in OCaml: there are now 3 languages: `Ocaml`, `OcamlInterface`, and `OcamlImplementation`.

`OcamlInterface` is the language used for `.mli` files, its associated grammar is the OCaml interface grammar.
`OcamlImplementation` is the language used for `.ml` files, its associated grammar is the OCaml implementation grammar.
`Ocaml` isn't associated with any file extension, and is associated with both implementaiton and interface grammars, in that order.

All three languages are associated with the `ocaml.scm` query file. The query file itself specifies that it uses the `Ocaml` language.

In practice, it means that if a `.ml` or `.mli` file is provided with the `--input-file` argument, or if the `--language` argument is set to `ocaml-implementation` or `ocaml-interface`, then the correct grammar will be used immediately.

However, if the input is provided via `stdin` and the `--language` argument is `ocaml`, or the `--query` argument is `languages/ocaml.scm`, then Topiary will try the implementation grammar, then the interface grammar if the first fails. It will then format the resulting tree with `ocaml.scm`.

### The special case of ambiguous interface files

Some interface files parse correctly with both the interface grammar and the implementation grammar, but produce syntax trees in which some node types have changed. For instance, the following piece of code:
```
include Type_intf.Type
```
produces this tree when parsed with the interface grammar:
```
{Node include_module_type (27, 0) - (27, 22)} - Named: true
  {Node include (27, 0) - (27, 7)} - Named: false
  {Node module_type_path (27, 8) - (27, 22)} - Named: true
    {Node extended_module_path (27, 8) - (27, 17)} - Named: true
      {Node module_name (27, 8) - (27, 17)} - Named: true
    {Node . (27, 17) - (27, 18)} - Named: false
    {Node module_type_name (27, 18) - (27, 22)} - Named: true
```
but produces this tree when parsed with the implementation grammar:
```
{Node include_module (27, 0) - (27, 22)} - Named: true
  {Node include (27, 0) - (27, 7)} - Named: false
  {Node module_path (27, 8) - (27, 22)} - Named: true
    {Node module_path (27, 8) - (27, 17)} - Named: true
      {Node module_name (27, 8) - (27, 17)} - Named: true
    {Node . (27, 17) - (27, 18)} - Named: false
    {Node module_name (27, 18) - (27, 22)} - Named: true
```

However, we want both trees to be formatted the same way: we don't want the way we've chosen to input the file in Topiary to change the output result.

To ensure this property, this PR introduces the notion of equivalence classes over node types in a query file. Simply put, if two node types belong to the same equivalence class, then they must appear in the same queries. This guarantees that the syntax trees produced by two different grammars will be formatted the same way.

So far, the following equivalence classes have been identified for OCaml:
- `include_module`, `include_module_type`
- `module_type_path`, `module_path`, `extended_module_path`
- `module_type_name`, `module_name`

The equivalence classes are not enforced yet, it's up to the maintainers to respect them. This could change in the future.

Closes #164

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1,6 +1,27 @@
 ; Configuration
 (#language! ocaml)
 
+; This query file is used to format trees produced by two different grammars:
+; - the grammar for OCaml interface files `tree_sitter_ocaml::language_ocaml_interface()`
+; - the grammar for OCaml implementation files `tree_sitter_ocaml::language_ocaml()`
+
+; On certain cases, some files may correctly parse with both grammars, but produce
+; different syntax trees. We want those files to be formatted the same way,
+; independently of the grammar used to parse them.
+
+; To ensure this property, we introduce equivalence classes on syntax node types.
+; Simply put, if two node types are in the same equivalence class, then they must
+; appear in the same queries.
+
+; If you add or modify a query containing a node type in one of the equivalence classes,
+; please make sure to add or modify the corresponding queries for all other node types
+; in the same equivalence class.
+
+; The equivalence classes are the following:
+; - include_module, include_module_type
+; - module_path, extended_module_path, module_type_path
+; - module_name, module_type_name
+
 ; Sometimes we want to indicate that certain parts of our source text should
 ; not be formatted, but taken as is. We use the leaf capture name to inform the
 ; tool of this.
@@ -21,8 +42,9 @@
   (exception_definition)
   (external)
   (floating_attribute)
-  (include_module)
-  (include_module_type)
+  ; equivalence class
+    (include_module)
+    (include_module_type)
   (inheritance_definition)
   (inheritance_specification)
   (instance_variable_definition)
@@ -63,7 +85,9 @@
   [
     (exception_definition)
     (external)
-    (include_module)
+    ; equivalence class
+      (include_module)
+      (include_module_type)
     (module_definition)
     (module_type_definition)
     (open_module)
@@ -390,10 +414,12 @@
     (class_type_path)
     (constructed_type)
     (constructor_path)
-    (extended_module_path)
     (field_get_expression)
     (labeled_argument)
-    (module_path)
+    ; equivalence class
+      (extended_module_path)
+      (module_path)
+      (module_type_path)
     (number)
     (parenthesized_expression)
     (parenthesized_pattern)
@@ -429,11 +455,13 @@
     (constructed_type)
     (constructor_path)
     (constructor_pattern)
-    (extended_module_path)
     (field_get_expression)
     (local_open_pattern)
     (labeled_argument)
-    (module_path)
+    ; equivalence class
+      (extended_module_path)
+      (module_path)
+      (module_type_path)
     (number)
     (parenthesized_expression)
     (parenthesized_pattern)
@@ -700,7 +728,9 @@
   [
     (value_specification)
     (type_definition)
-    (include_module_type)
+    ; equivalence class
+      (include_module)
+      (include_module_type)
   ] @append_spaced_softline
 )
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,7 +72,7 @@ pub type FormatterResult<T> = std::result::Result<T, FormatterError>;
 /// let mut output = Vec::new();
 /// let mut query = BufReader::new(File::open("languages/json.scm").expect("query file"));
 ///
-/// match formatter(&mut input, &mut output, &mut query, false) {
+/// match formatter(&mut input, &mut output, &mut query, None, false) {
 ///   Ok(()) => {
 ///     let formatted = String::from_utf8(output).expect("valid utf-8");
 ///   }
@@ -88,6 +88,7 @@ pub fn formatter(
     input: &mut dyn io::Read,
     output: &mut dyn io::Write,
     query: &mut dyn io::Read,
+    language: Option<Language>,
     skip_idempotence: bool,
 ) -> FormatterResult<()> {
     let content = read_input(input).map_err(|e| {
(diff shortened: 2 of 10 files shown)


Xophmeister (approved):
LGTM :ok_hand: ...although the CI is failing with a `File Not Found` error in the integratio tests.

languages/ocaml.scm
```diff
@@ -1,6 +1,27 @@
 ; Configuration
 (#language! ocaml)
 
+; This query file is used to format trees produced by two different grammars:
+; - the grammar for OCaml interface files `tree_sitter_ocaml::language_ocaml_interface()`
+; - the grammar for OCaml implementation files `tree_sitter_ocaml::language_ocaml()`
+
+; On certain cases, some files may correctly parse with both grammars, but produce
+; different syntax trees. We want those files to be formatted the same way,
+; independantly of the grammar used to parse them.
```
torhovland, line 10:
```suggestion
; independently of the grammar used to parse them.
```

languages/ocaml.scm
```diff
@@ -394,6 +416,7 @@
     (field_get_expression)
     (labeled_argument)
     (module_path)
+    (module_type_path)
```
torhovland, line 419: Shouldn't `extended_module_path` also be here, then?

languages/ocaml.scm
```diff
@@ -434,6 +457,7 @@
     (local_open_pattern)
     (labeled_argument)
     (module_path)
+    (module_type_path)
```
torhovland, line 460: Same as above.

src/lib.rs
```diff
@@ -97,7 +98,12 @@ pub fn formatter(
         FormatterError::Reading(ReadingError::Io("Failed to read query content".into(), e))
     })?;
 
-    let configuration = Configuration::parse(&query)?;
+    let mut configuration = Configuration::parse(&query)?;
+    // Replace the language deduced from the query file by the one from the CLI, if any
+    match language {
```
torhovland, line 103: I think `if let Some(...) = ...` reads nicer here.

src/tree_sitter.rs
```diff
@@ -81,25 +78,37 @@ fn capture_name<'a, 'b>(query: &'a Query, capture: &'b QueryCapture) -> &'a str
...
-        Language::Ocaml => tree_sitter_ocaml::language_ocaml(),
-        Language::Rust => tree_sitter_rust::language(),
-        Language::Toml => tree_sitter_toml::language(),
-    }
-}
-
-fn parse(content: &str, grammar: tree_sitter::Language) -> FormatterResult<Tree> {
+// A single "language" can correspond to multiple grammars.
+// For instance, we have separate grammars for interfaces and implementation in OCaml.
+// When the proper grammar cannot be infered from the extension of the input file,
```
torhovland, line 83:
```suggestion
// When the proper grammar cannot be inferred from the extension of the input file,
```

torhovland (approved):

languages/ocaml.scm
nbacquey: fixed

languages/ocaml.scm
nbacquey: It was just above in the list. I made the equivalence classes more explicit in the new version

languages/ocaml.scm
nbacquey: Same as above as well.

src/lib.rs
nbacquey: Indeed, thanks. I did that in the new version.

src/tree_sitter.rs
nbacquey: fixed

Issue #172, opened 2023-01-13
Ability to add line continuations in multi-line contexts


Xophmeister:
**Is your feature request related to a problem? Please describe.**
In some languages (e.g., Python, Bash), multi-line contexts require a line continuation character; usually an escaped newline.

I don't believe -- but I'm not certain! -- that it's possible to distinguish between a single-line and multi-line context, in the query language. That is, in the sense of performing different actions in different contexts. Therefore, line continuations can't be inserted as though they were delimiters, because they'd be inserted in all contexts.

**Describe the solution you'd like**
As this would apply to the source as a whole, perhaps a new directive:

```scm
(#multiline! STRING)
```

The default would be `(#multiline! "\n")`. However, this could be overridden to something like `(#multiline! "\\\n")` -- or whatever the language's escape character is -- when needed.

Any newlines that are added by Topiary in a multi-line context will use this `STRING`, rather than a simple `\n`.

**Describe alternatives you've considered**
One could _assume_ the programmer knew what they were doing and thus look for manually added line continuation nodes. Then queries could be written that target these characters: if they exist, we're in a multi-line context; otherwise, single-line.

My gut feeling is that this may incorrectly target non-line continuation nodes. However, even if that's not the case, it would double the number of queries required (i.e., an explicit multi-line and single-line version, for each).

Xophmeister:
A rule that explicitly targets the line-continuation and appends a hardline was suggested, so multi-line Bash commands preserve their formatting. i.e.,:

```scheme
(command
  "\\" @append_hardline
)
```

Unfortunately, the Bash grammar rejects this outright with an `Invalid node type` error :disappointed:

Niols:
I'm actually quite curious to see how that will unfold. I don't know about bash, but in POSIX Shell, the line continuation is not handled in the parser but in the lexer. If the tree-sitter grammar follows the official one, I suppose it would not appear there. (Again, bash is very much not POSIX Shell so it might be doing its own thing there.)

Xophmeister:
On further thought, having this apply to the whole source will be too general; even if it's limited to just softlines. For example, I've specified the following queries to format Bash compound statements:

```scheme
(compound_statement
  .
  "{" @append_spaced_softline @append_indent_start
)

(compound_statement
  "}" @prepend_spaced_softline @prepend_indent_end
  .
)
```

With the proposal in the OP issue, that would format a multi-line compound statement like so:

```bash
{ \
  whatever \
}
```

[...]

Issue #171, opened 2023-01-12
topiary.tweag.io Interactive Playground


Xophmeister:
**Is your feature request related to a problem? Please describe.**
To prototype new queries and to play with Topiary as a formatting engine (e.g., to drive adoption), it would be really useful to have an online, interactive playground at, say, topiary.tweag.io (or similar). These are very common in this space: e.g., [Tree Sitter](https://tree-sitter.github.io/tree-sitter/playground), [Shellcheck](https://www.shellcheck.net), [Rust](https://play.rust-lang.org), [Black (third-party playground)](https://black.vercel.app), etc.

**Describe the solution you'd like**
Provide a simple playground website (frontend and backend), with the following widgets:
* Source code input
* Source code grammar selector
* Topiary query input (defaults to the bundled language queries, based on grammar selector)
* Output panel

The output panel would update interactively, on changes to the input widgets.

**Describe alternatives you've considered**
It's quite easy to rig this up in the terminal, for prototyping and playing, but there's a non-zero amount of friction.

**Additional context**
Many playgrounds allow you to save/share state. This would be a "nice-to-have", but not strictly necessary, IMO.

Xophmeister:
Reporting findings, so far, so they don't get lost:

### WASM vs. REST

A REST API would be very easy to write. However, that would then require a long-term commitment on server infrastructure and all that entails.

A WASM target is harder to write, presuming it can be done, but then it can be statically hosted with a trivial maintenance burden.

### Is WASM possible?

The [Tree-sitter playground](https://tree-sitter.github.io/tree-sitter/playground) is using WASM. Also, the [Rust Sitter project](https://github.com/hydro-project/rust-sitter) contains this quote in its `README`:

> _Note: By default, Rust Sitter uses a fork of Tree Sitter with a pure-Rust runtime to support `wasm32-unknown-unknown`. To use the standard C runtime instead, disable default features and enable the `tree-sitter-standard` feature_

That highly suggests that, not only is it possible, but gives two departure points to see how it's done.

### Notes and Resources

* [Rust and WebAssembly](https://rustwasm.github.io) \
  This contains a bunch of resources, including a [book](https://rustwasm.github.io/docs/book/). It was a struggle to get the book exercises to work as described -- things have moved on, since it was written -- but it's possible, albeit leaving one with a somewhat unsatisfied feeling from the superficial understanding.

* Nix\
[...]

PR #173, opened 2023-01-13
[Bash] More command formatting

Xophmeister:
This PR introduces queries that improve support for command, command lists and command pipelines, in aid of #138. It is outputting valid code, but suffers from some (non-blocking, imo) shortfalls:

* The line-spacing is not perfect;
* Multi-line "commands" are collapsed into a single line (see #172)

I have also cleaned-up and checked-in the script -- with documentation -- I created to support interactive query development in the terminal (as a quick-and-dirty alternative to #171); it seems generally useful.

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -5,15 +5,17 @@
 (string) @leaf
 
 ; Allow blank line before
-; TODO Add directives here...
 [
   (comment)
   (if_statement)
+  (command)
+  (list)
+  (pipeline)
+  ; TODO: etc.
 ] @allow_blank_line_before
 
 ; Prepend spaces
 ; (Not "surround", to avoid extraneous spaces between delimiters)
-; TODO Add keywords, etc., here...
 [
   "if"
   "then"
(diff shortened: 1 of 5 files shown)


languages/bash.scm
```diff
@@ -5,15 +5,17 @@
...
   (comment)
   (if_statement)
+  (command)
+  (list)
+  (pipeline)
+  ; TODO: etc.
 ] @allow_blank_line_before
 
 ; Prepend spaces
 ; (Not "surround", to avoid extraneous spaces between delimiters)
```
nbacquey, line 18: I think we can surround, and let the post-processing take care of extraneous spaces

languages/bash.scm
```diff
@@ -22,23 +24,43 @@
...
+; collapsed on to a single line for now... See Issue #172
+
+; One command per line in the following contexts:
+; * Top-level
+; * In any branch of a conditional
+; * <TODO: etc.>
+;
+; NOTE Because "command" is such a pervasive and general concept, each
+; context needs to be individually enumerated to account for exceptions.
+(program [(command) (list) (pipeline)] @prepend_hardline)
```
nbacquey, line 51: Even if it's more space-consuming, I think I prefer the heavily-indented way to write queries, even simple ones.
At least for consistency within the query files, e.g.
```
(program
  [
    (command)
    (list)
    (pipeline)
  ] @prepend_hardline
)
```
Or at least
```
(program
  [(command) (list) (pipeline)] @prepend_hardline
)
```
if the `[(command) (list) (pipeline)]` syntagm is expected to be everywhere

nbacquey (approved):

languages/bash.scm
Xophmeister: The reason this is in here is to protect the `<CLAUSE>; <CLAUSE>` case; i.e., no space before the delimiter. I don't believe the post-processing deals with this.

That said, I've fixed this in #174 by making the "surround with spaces" list much more restrictive.

languages/bash.scm
Xophmeister: Good idea :+1: I'll fix this in #174 to avoid making merge conflict problems for myself

languages/bash.scm
```diff
@@ -22,23 +24,43 @@
...
 ;
-; FIXME Can this be generalised, or does *every* context need to be
-; individually enumerated?...
-(program (command) @prepend_spaced_softline @append_hardline)
-(if_statement _ "then" (command) @prepend_spaced_softline)
-(elif_clause _ "then" (command) @prepend_spaced_softline)
-(else_clause (command) @prepend_spaced_softline)
+; That is: [(command) (list) (pipeline)], per the grammar
+
+; FIXME I don't think it's possible to insert the necessary line
```
torhovland, line 39: Should `command` be treated as a leaf (so we leave it entirely alone)?

languages/bash.scm
```diff
@@ -22,23 +24,43 @@
...
+;
+; NOTE Because "command" is such a pervasive and general concept, each
+; context needs to be individually enumerated to account for exceptions.
+(program [(command) (list) (pipeline)] @prepend_hardline)
+(if_statement _ "then" [(command) (list) (pipeline)] @prepend_hardline)
+(elif_clause _ "then" [(command) (list) (pipeline)] @prepend_hardline)
+(else_clause [(command) (list) (pipeline)] @prepend_hardline)
+
+; Surround command list and pipeline delimiters with spaces
+(list ["&&" "||"] @append_space @prepend_space)
```
torhovland, line 57: Could we just put these in the list of symbols to surround with spaces? Do they really need to come within `list` and `pipeline`?

Xophmeister: I guess they don't. I've done it like this to make the context explicit, as I build the queries up incrementally; but you're right that it's redundant.

I'll add a note next to these rules (in #174) to ultimate generalise them into the list of symbols to surround with spaces.

Xophmeister: Note added in 6b4c69e8641193

languages/bash.scm
Xophmeister: I'm not fully clear about the consequences of `@leaf`. I understand that Topiary just skips over these nodes, but to what end/what effect would that have on `(command)`?

torhovland: It would keep leafs formatted as is. I would think that might be better than forcing all commands to be single-line.

Xophmeister: Ah, I see...

So the following:

```bash
foo \
--bar
```

Will be formatted as `foo --bar`, without making it a leaf. Whereas, as a leaf, like you say, no formatting changes are applied. That's certainly better, assuming that's what the author intended :+1:

Changed in cfbd8cb2cd64568

Xophmeister: ...Oh, but then that obviates this rule:

```scheme
; Space between command line arguments
(command
  argument: _* @append_space @prepend_space
)
```

Which formats, e.g., `foo     --bar  --quux` into `foo --bar --quux`... I quite like this rule.

torhovland: Yeah, I see.

PR #179, opened 2023-01-17
[Bash] Functions and variables

Xophmeister:
This PR introduces formatting queries for function definitions, variable declaration and variable assignment, in aid of #138.

Note: It would be _really_ nice to be able to write a query that can convert "simple expansions" (e.g., `$foo`) into full, unambiguous expansions (i.e., `${foo}`) as a stylistic choice; similar to how I'm forcibly removing the redundant `function` keyword. I played around with inserting delimiters, but couldn't make anything work :cry:

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -1,8 +1,12 @@
 ; Configuration
 (#language! bash)
 
-; Don't modify string literals
-(string) @leaf
+; Don't modify string literals or variable expansions
+[
+  (expansion)
+  (simple_expansion)
+  (string)
+] @leaf
 
 ; Allow blank line before
 ; FIXME Blank line spacing around major syntactic blocks is not correct.
@@ -13,12 +17,14 @@
   (command)
   (comment)
   (compound_statement)
+  (declaration_command)
   (for_statement)
   (function_definition)
   (if_statement)
   (list)
   (pipeline)
   (subshell)
+  (variable_assignment)
   (while_statement)
   ; TODO: etc.
 ] @allow_blank_line_before
@@ -26,17 +32,22 @@
 ; Surround with spaces
 [
   "case"
+  "declare"
   "do"
   "done"
   "elif"
   "else"
   "esac"
+  "export"
   "fi"
   "for"
   "if"
   "in"
+  "local"
+  "readonly"
   "select"
   "then"
+  "typeset"
   "until"
   "while"
   (string)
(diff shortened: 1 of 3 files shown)


languages/bash.scm
```diff
@@ -328,3 +340,45 @@
...
+)
+
+; NOTE The (simple_expansion), for `$foo`, and (expansion), for `${foo}`
+; and friends, node types exist. We consider them as leaves (see above).
+; However, it would be _really_ nice if we could write a query that
+; converts all (simple_expansions) into (expansions). It can almost be
+; done with delimiters, but it doesn't quite work :( For example:
+;
+; (simple_expansion (variable_name) @prepend_delimiter (#delimiter! "{"))
+; (simple_expansion (variable_name) @append_delimiter (#delimiter! "}"))
```
torhovland, line 384: Why doesn't it?

languages/bash.scm
```diff
@@ -328,3 +340,45 @@
...
+  body: _ @prepend_space
+)
+
+; NOTE The "function" keyword in function definitions is optional and
+; thus usually considered redundant. Therefore we delete it, if it's
+; present in the input.
+(function_definition
+  .
+  "function" @delete
+)
```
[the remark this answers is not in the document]
aspiwack: I'm certainly not one to have an opinion on how Bash should be written. I notice that in my scripts I usually don't write `function`, I'll trust @Xophmeister that this is typical.

I agree that this is the sort of things that we probably want to normalise.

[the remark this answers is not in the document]
torhovland: OK. I thought you might have felt that this is a case where we need to respect the input of the programmer. This gets approved, then.

torhovland (approved):

languages/bash.scm
Xophmeister: A `(simple_expansion)` is made up of `$` and `(variable_name)`. Adding the closing `}` works fine, but I cannot write a query that interposes the `{` between the sigil and name.

The above, commented out query, renders `{$foo}`. I tried being really specific, with anchors, like:

```scm
(simple_expansion
  .
  "$" @append_delimiter
  .
  (variable_name) ; OR: @prepend_delimiter
  .

  (#delimiter! "{")
)
```

...but that gives `$foo}{` and `{$foo}`, respectively.

torhovland: Hmm, that is odd. Does it make a difference if you put `"$"` in parens?

Xophmeister: No luck, I'm afraid

torhovland: 🤷

PR #182, opened 2023-01-18
[Bash] Redirection, herestrings, heredocs, command and process substitution

Xophmeister:
This PR introduces formatting support for the following Bash constructs, in aid of #138:
* [x] Redirection statements
* [x] Herestrings and heredocs
* [x] Command and process substitution
* [x] Inline environment variable assignment (e.g., `FOO=bar quux`)
* [x] Opinionated line spacing
  * New line after multi-line syntactic blocks (e.g., `if` statements, subshells, etc.)
  * New line after a run of "units of execution" (± a few others) before a multi-line syntactic block or variable
  * New line after a run of variable declarations or assignments.

Notes:
* Process substitution "just works", without requiring special rules.
* There is an (unassailable) problem with heredocs having their first line affected by the indent level.
* This PR fixes new lines for variable assignment outside of declaration contexts. This further complicates the "unit of execution" alternation and speciation, given the lack of negative anchors in the Tree Sitter query language.

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -74,6 +74,7 @@
   [
     (c_style_for_statement)
     (case_statement)
+    (comment)
     (compound_statement)
     (declaration_command)
     (for_statement)
@@ -97,6 +98,7 @@
     (c_style_for_statement)
     (case_statement)
     (command)
+    (comment)
     (compound_statement)
     (compound_statement)
     (for_statement)
@@ -133,6 +135,8 @@
   "until"
   "while"
 ] @append_space @prepend_space
+
+(comment) @prepend_hardline
 
 ;; Compound Statements and Subshells
 
@@ -316,7 +320,8 @@
 ; Ensure heredocs start on a new line, after their start marker, and
 ; there is a new line after their end marker
 ; NOTE This is a syntactic requirement
-(heredoc_body) @prepend_hardline @append_input_softline
+(heredoc_start) @append_hardline
+(heredoc_body) @append_input_softline
 
 ;; Conditionals
 
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -5,11 +5,14 @@
 ; any which are encountered by Topiary will be forcibly collapsed on to
 ; a single line. (See Issue #172)
 
-; Don't modify string literals or variable expansions
+; Don't modify string literals, heredocs, atomic "words" or variable
+; expansions (simple or otherwise)
 [
   (expansion)
+  (heredoc_body)
   (simple_expansion)
   (string)
+  (word)
 ] @leaf
 
 ; Allow blank line before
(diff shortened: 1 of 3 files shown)


Xophmeister:
Comments are quite right. Let me fix that...

Xophmeister:
...Hmm, comments are harder than I anticipated. I'll try to get them working in their own PR.

So this one is ready for review.

languages/bash.scm
```diff
@@ -23,13 +32,88 @@
...
+    (for_statement)
+    (function_definition)
+    (if_statement)
+    (list)
+    (pipeline)
+    (redirected_statement)
+    (subshell)
+    (subshell)
+    (while_statement)
+  ]
```
torhovland, line 113: Do you need this? Shouldn't the above always be followed by a hardline?

languages/bash.scm
```diff
@@ -23,13 +32,88 @@
...
+    (comment)
+    (compound_statement)
+    (declaration_command)
+    (for_statement)
+    (function_definition)
+    (if_statement)
+    (subshell)
+    (variable_assignment)
+    (while_statement)
+  ]
```
torhovland, line 86: What would go wrong if you simply dropped this part?

torhovland (approved):

languages/bash.scm
Xophmeister: Variable declarations and assignments get a `@prepend_hardline` elsewhere, so they appear on a new line that way. This block also adds a hardline after a run of variables followed by one of the nodes in that alternation, which are also `@prepend_hardline`. This formats as a blank link between a run of variables and those nodes.

If I comment out this anchor + alternation, it has two effects:
1. Every variable declaration and assignment is separated by blank lines (redundant, but syntactically OK)
2. There are cases where declarations and assignments can be mixed and this will split them (which is syntactically invalid). For example:

    ```bash
    export A B=1 C
    ```

    ...gets formatted as:

    ```bash
    export A B=1
    C
    ```

So this somewhat complicated rule is largely from my choice of prepending hardlines, rather than appending them (see #186; specifically the commit message in 968852323). The choice of prepending was because I couldn't get "indent blocks" to work correctly using an append strategy without either an errant new line or spurious indentation:

1. ```bash
   while true; do
     foo

   done
   ```

2. ```bash
   while true; do
     foo
     done
   ```

However, my hunch is that this choice is what's making comments particularly difficult (impossible?) to format; hence #186. (It's also "weird" in the sense that it breaks expectations, which is also something worth correcting.)

languages/bash.scm
Xophmeister: (Same as https://github.com/tweag/topiary/pull/182#discussion_r1083931668: Doing so puts blanks lines where they're unnecessary, including places where they're syntactically invalid.)

PR #186, opened 2023-01-20
[Bash] Comments

### LINKED ISSUE #138: Bash formatting support

Xophmeister:
This PR introduces:
* [x] Formatting queries for comments
* [x] Idempotence (except heredocs)

Notes
* [x] Depends on #193 
* Resolves #138

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -451,13 +451,18 @@
 
 ; Ensure heredocs start on a new line, after their start marker, and
 ; there is a new line after their end marker, when followed by any named
-; node. (NOTE This may need some refinement...)
+; node. (NOTE This may still need refinement...)
 ; NOTE These are a syntactic requirements
 (heredoc_start) @append_hardline
+
+; NOTE Anecdotally, queries that target the (heredoc_body) node act
+; differently, dependant upon expansions existing in the heredoc. Hence
+; this query, that targets any following named node, rather than simply
+; appending a new line to any heredoc (which doesn't work).
 (
-  (heredoc_body) @append_hardline
+  (heredoc_body)
   .
-  (_)
+  (_) @prepend_hardline
 )
 
 ;; Conditionals
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -149,14 +149,42 @@
 ; 2. All comments must end with a new line.
 ; 3. Comments can be interposed by blank lines, if they exist in the
 ;    input (i.e., blank lines shouldn't be engineered elsewhere).
-; 4. A run of standalone comments (i.e., without anything, including
-;    blank lines, interposing) should be kept together.
-; 5. Trailing comments should only appear after "units of execution" or
-;    variable declarations/assignment. (This is despite it being
-;    syntactically valid to put them elsewhere.)
+; 4. A comment can never change flavour (i.e., standalone to trailing,
+;    or vice versa).
+; 5. Trailing comments should only appear after "units of execution"
+;    (see Commands section, below) or variable declarations/assignment.
+;    (This is despite it being syntactically valid to put them
+;    elsewhere.)
 
-; FIXME
+; Rule 1: See @leaf rule, above
+
+; Rule 2
 (comment) @append_hardline
+
+; Rule 3: See @allow_blank_line_before rule, above, plus the below to
+; account for standalone comments following trailing comments
+(
+  ; Context: Unit of execution or variable
+  [(command) (list) (pipeline) (subshell) (compound_statement) (redirected_statement) (declaration_command) (variable_assignment)]
+  .
+  (comment)
+  .
+  (comment) @prepend_hardline
+)
+
+; Rule 4: We only have to protect against the case of a standalone
+; comment, after a statement, being slurped on to that statement's line
+; and becoming a trailing comment.
+; FIXME This case cannot be distinguished by the grammar. There may be
+; another way to express it...
+
+; Rule 5
+(
+  ; Context: Unit of execution or variable
+  [(command) (list) (pipeline) (subshell) (compound_statement) (redirected_statement) (declaration_command) (variable_assignment)]
+  .
+  (comment) @prepend_space
+)
 
 ;; Compound Statements and Subshells
 
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -151,39 +151,26 @@
 ;    input (i.e., blank lines shouldn't be engineered elsewhere).
 ; 4. A comment can never change flavour (i.e., standalone to trailing,
 ;    or vice versa).
-; 5. Trailing comments should only appear after "units of execution"
-;    (see Commands section, below) or variable declarations/assignment.
-;    (This is despite it being syntactically valid to put them
-;    elsewhere.)
+; 5. Trailing comments should be interposed by a space.
 
 ; Rule 1: See @leaf rule, above
 
 ; Rule 2
 (comment) @append_hardline
 
-; Rule 3: See @allow_blank_line_before rule, above, plus the below to
-; account for standalone comments following trailing comments
-(
-  ; Context: Unit of execution or variable
-  [(command) (list) (pipeline) (subshell) (compound_statement) (redirected_statement) (declaration_command) (variable_assignment)]
-  .
-  (comment)
-  .
-  (comment) @prepend_hardline
-)
+; Rule 3: See @allow_blank_line_before rule, above.
+; FIXME This doesn't quite get us what we want. It's close, but blank
+; lines between comments can get consumed.
 
 ; Rule 4: We only have to protect against the case of a standalone
 ; comment, after a statement, being slurped on to that statement's line
-; and becoming a trailing comment.
-; FIXME This case cannot be distinguished by the grammar. There may be
-; another way to express it...
+; and becoming a trailing comment. That case is satisfied by Rule 5.
 
 ; Rule 5
 (
-  ; Context: Unit of execution or variable
-  [(command) (list) (pipeline) (subshell) (compound_statement) (redirected_statement) (declaration_command) (variable_assignment)]
-  .
   (comment) @prepend_space
+  .
+  "\n"
 )
 
 ;; Compound Statements and Subshells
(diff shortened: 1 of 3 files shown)


Xophmeister:
Status update:
* I have reversed the new line strategy to append and reached parity with the previous branch (modulo errant trailing blank lines, which should be easily fixable).
* Reversing the strategy has paid off and comments are now easier to format. There's one outstanding issue -- standalone comments being slurped into trailing comments -- which I've yet to work out. (The Rust queries can handle this, but the Bash grammar may not be pinned down enough to emulate its approach...)
* There's something wrong with the heredoc formatting (besides the indentation issue): they're not getting a hardline after the end marker in some contexts.

aspiwack:
> I have reversed the new line strategy to append and reached parity with the previous branch (modulo errant trailing blank lines, which should be easily fixable).

I think you probably want to make this its own PR.

Xophmeister:
> I think you probably want to make this its own PR.

Good idea. I've factored this out into #193.

Xophmeister:
### Status Update

The Bash grammar considers comments to be sibling nodes to those around it. For example,

```bash
# comment
command #comment
```

...has a tree that looks like this:

* `program`
  * `comment`
  * `command`
  * `comment`

The tree structure alone cannot distinguish between a comment that trails a command on the same line, and a "nude" command with a comment interposed by new line(s). That is, the following have equivalent structures:

```bash
command # comment

command
# comment
```

Normally, an input softline would resolve this problem. However, it doesn't seem to be possible with the Bash grammar.

For instance, the Rust grammar also considers comments to be sibling nodes. Yet the following minimal set of queries will format comments as expected:

```scm
(#language! rust)
"mod" @prepend_space @append_space

; Comment formatting

(line_comment) @allow_blank_line_before

(line_comment) @append_hardline

(
  ";" @append_input_softline
  .
  (line_comment)* @do_nothing
)
```

This makes no change to the following minimal example:

```rust
// standalone comment

// standalone comment
// standalone comment
mod a;

// standalone comment
mod b; // trailing comment

// standalone comment
// standalone comment
mod c; // trailing comment
mod d;
mod e;
```

If we try to mechanically convert this to the Bash grammar, we get:

```scm
(#language! bash)

; Comment formatting

(comment) @allow_blank_line_before

(comment) @append_hardline

(
  (command) @append_input_softline
  .
  (comment)* @do_nothing
)
```

This formats the following (correctly formatted) input:

```bash
# standalone comment

# standalone comment
# standalone comment
command

# standalone comment
command # trailing comment

# standalone comment
# standalone comment
command # trailing comment
command
command
```

...as this monstrosity:

```bash
# standalone comment

# standalone comment
# standalone comment
command# standalone comment
command# trailing comment
# standalone comment
# standalone comment
command# trailing comment
command command
```

You'll notice the following problems:
1. The standalone comment that follows the first command has been slurped into the trailing position.
2. Trailing comments are not interposed with a space.
3. The blank line before the last set of standalone comments has been consumed.
4. A sequence of commands is collapsed on to a single line, separated by spaces. (This introduces an egregious syntactic change from the input.)

There are things we can do to improve this:
* Commands that are followed by commands should have a new line interposed.
* All comments should be prepended with an input softline, rather than having commands prepended with one when a comment follows.
* When a command is followed by _two_ comment nodes, then that second comment _must_ be a standalone comment, so it can have a hardline prepended. (You'll note that this is a bit of a hack! The first comment could also be a standalone comment, so this may insert an errant blank line in that context.)

We thus arrive at this:
```scm
(#language! bash)

; Command formatting

(
  (command) @append_hardline
  .
  (command)
)

; Comment formatting

(comment) @allow_blank_line_before

(comment) @prepend_input_softline @append_hardline

(
  (command)
  .
  (comment)
  .
  (comment) @prepend_hardline
)
```

Which formats the minimal example like so:
```bash
# standalone comment

# standalone comment
# standalone comment
command # standalone comment
command # trailing comment

# standalone comment
# standalone comment
command # trailing comment
command
command
```

This is _almost_ correct...except for that standalone comment that it again slurps into the trailing position.

### To Do

Use these minimal examples to understand why the Bash comment formatting does not behave as we'd expect (and is demonstrably proven by the Rust counterexample).

Xophmeister:
Rust's tree:
* `source_file`
  * `line_comment`
  * `line_comment`
  * `line_comment`
  * `mod_item`
    * `mod`
    * `identifier`
    * `;`
  * `line_comment`
  * `mod_item`
    * `mod`
    * `identifier`
    * `;`
  * `line_comment`
  * `line_comment`
  * `line_comment`
  * `mod_item`
    * `mod`
    * `identifier`
    * `;`
  * `line_comment`
  * `mod_item`
    * `mod`
    * `identifier`
    * `;`
  * `mod_item`
    * `mod`
    * `identifier`
    * `;`

Bash's tree:
* `program`
  * `comment`
  * `comment`
  * `comment`
  * `command`
    * `command_name`
      * `word`
  * `\n`
  * `comment`
  * `command`
    * `command_name`
      * `word`
  * `comment`
  * `\n`
  * `comment`
  * `comment`
  * `command`
    * `command_name`
      * `word`
  * `comment`
  * `\n`
  * `command`
    * `command_name`
      * `word`
  * `\n`
  * `command`
    * `command_name`
      * `word`
  * `\n`

...Those anonymous `\n` nodes are suspicious, methinks!

Xophmeister:
This "works":
```scm
(#language! bash)

; Command formatting

; ONLY append a hard line after a command when it is followed by an
; anonymous new line node
(
  (command) @append_hardline
  .
  "\n" ; @delete: This doesn't do anything
)

; Comment formatting

; Allow blank line to interpose comments
(comment) @allow_blank_line_before

; ALWAYS append a hard line after a comment
(comment) @append_hardline

; When a comment is followed by an anonymous new line node, it MUST be a
; trailing comment, so prepend a space
(
  (comment) @prepend_space
  .
  "\n" ; @delete: This doesn't do anything
)
```

...rendering the following output:
```bash
# standalone comment

# standalone comment
# standalone comment
command
# standalone comment
command # trailing comment
# standalone comment
# standalone comment
command # trailing comment
command
command

```

It's still not as well-behaved as the Rust code, but it at least gets most of what we want.

Xophmeister:
Targetting the anonymous `\n` nodes has allowed us to preserve the semantics of comment formatting. It also simplifies the queries for line spacing of "units of execution".

However, it comes at a cost: The final command in a block is followed by an errant blank line. For example:

```bash
while true; do
  something

done
```

We can remove this by applying the `@delete` capture name against the `\n` node, but that in turn breaks other things. (Namely: blank line spacing between "blocks" and the correct indentation of case statements when the final branch is missing a terminator.) On balance, this is the least bad option and can probably be fixed in a future PR.

There is also a bug with variable assignments involving command substitutions; an invalid space is inserted, resulting in a syntax error in the final output. I'll attempt to fix that before finalising this PR.

torhovland:
There is something wrong with this premise:

> For instance, the Rust grammar also considers comments to be sibling nodes. Yet the following minimal set of queries will format comments as expected:

You are missing these from the Rust queries:

```scheme
; Input softlines before and after all comments. This means that the input
; decides if a comment should have line breaks before or after. A line comment
; always ends with a line break.
[
  (block_comment)
  (line_comment)
] @prepend_input_softline

; Input softline after block comments unless followed by comma or semicolon, as
; they are always put directly after.
(
  (block_comment) @append_input_softline
  .
  [ "," ";" ]* @do_nothing
)
```

In your case, only the prepend before line comment is relevant.

Can you take another look with this in mind?

Xophmeister:
I'm only interested in line comments, because Bash doesn't have block comments.[^1] So this rule, in the Rust queries:
```scm
; Input softlines before and after all comments. This means that the input
; decides if a comment should have line breaks before or after. A line comment
; always ends with a line break.
[
  (block_comment)
  (line_comment)
] @prepend_input_softline
```
...can be simplified -- for the sake of a minimal exemplar -- to just `(line_comment) @prepend_input_softline`. That correctly deals with trailing line comments, but in vitro, it causes adjacent statements to wrap on to a single line:
```rust
// ...
mod d; mod e;
```

Hence I changed it to the following, to exhibit the expected behaviour:
```scm
(
  ";" @append_input_softline
  .
  (line_comment)* @do_nothing
)
```

Arguably, the false premise is asserting that comment rules _can_ be considered in isolation, when in reality, they -- and the queries that govern them -- interact with everything that's around them. In that sense, the above query is a "hack" to that end. A more faithful minimal exemplar would therefore look something like:
```scm
(#language! rust)
"mod" @prepend_space @append_space

(line_comment) @allow_blank_line_before

(line_comment) @append_hardline

(line_comment) @prepend_input_softline

(
  (mod_item) @append_hardline
  .
  (line_comment)* @do_nothing
)
```

Mechanically converting this to target Bash does a reasonable job:
```scm
(#language! bash)

(comment) @allow_blank_line_before

(comment) @append_hardline

(comment) @prepend_input_softline

(
  (command) @append_hardline
  .
  (comment)* @do_nothing
)
```
Output:
```bash
# standalone comment

# standalone comment
# standalone comment
command # standalone comment
command # trailing comment
 # standalone comment
# standalone comment
command # trailing comment
command
command
␤
```

There are a few problems:
1. The standalone comment after the first command is slurped into the trailing position.
2. More generally, the blank lines before standalone comments have been consumed.
3. The unusual spacing is due to the Bash grammar inserting anonymous `\n` nodes. These appear in the following contexts:
   * After a `(command)` that is immediately followed by a new line.
   * After a `(comment)` that follows a sibling node and is immediately followed by a new line (i.e., a trailing comment).

The anonymous `\n` nodes is a divergence from the Rust grammar in two senses: we have to deal with them, to fix the unusual spacing; but they allow us to target more specific nodes. For instance, standalone commands can be targetted without a `@do_nothing` exception for trailing comments. Likewise, trailing comments can be explicitly targetted, so we can simplify `@prepend_input_softline` into `@prepend_space`:

```scm
(#language! bash)

(comment) @allow_blank_line_before

(comment) @append_hardline

(
  (command) @append_hardline
  .
  "\n" @delete
)

(
  (comment) @prepend_space
  .
  "\n" @delete
)
```

Output:

```bash
# standalone comment

# standalone comment
# standalone comment
command
# standalone comment
command # trailing comment
# standalone comment
# standalone comment
command # trailing comment
command
command
```

The only remaining problem is the consumption of new lines before standalone comments (see #202). Otherwise it's a good approximation of what we ultimately want. As such, these rules are close to the final rules defined in this PR, as of writing. The only difference is that the anonymous `\n` nodes are not `@delete`'d. This, IIRC, was to preserve correct indentation within blocks.

[^1]:
    As such, the following rule is simply not relevant, so it was omitted from the minimal exemplar:
    ```scm
    ; Input softline after block comments unless followed by comma or semicolon, as
    ; they are always put directly after.
    (
      (block_comment) @append_input_softline
      .
      [ "," ";" ]* @do_nothing
    )
    ```

PR #198, opened 2023-01-30
[OCaml] List formatting

### LINKED ISSUE #189: [OCaml] Bad formatting of multiline lists
**Describe the bug**
Formatting of multiline lists is quite poor. This is the output of topiary:
```
let _ =
  [1 ;
  2 ;
  3]
```
**Expected behavior**
I would expect the example above to be formatted in either one of those:
```
let _ = [
  1 ;
  2 ;
  3
]
```
```
let _ =
  [ 1
  ; 2
  ; 3
  ]
```
```
let _ =
  [
    1 ;
    2 ;
    3
  ]
```
```
let _ =
  [ 1 ;
    2 ;
    3
  ]
```

**Environment**
 - Ubuntu 22.04
 - main on 2023-01-24 (https://github.com/tweag/topiary/commit/3eefa610bab865a25adaf93a1046937b85836657)

nbacquey:
Allows proper formatting of
```ocaml
let _ =
  [
    1;
    2;
    3
  ]
```

Ideally, we would want to add a final `;` to multiline lists, but we don't have the capacity to do that yet: see #199.

Closes #189

### DIFF
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -573,6 +573,13 @@ let large_const =
     (long_argument_4 : int) ->
     val
 
+let _ =
+  [
+    1;
+    2;
+    3
+  ]
+
 (* Showcase the usage of operator bindings *)
 let greetings =
   let (let*) = Option.bind
@@ -588,11 +595,11 @@ let greetings =
   Some (msg1 ^ msg2)
 
 (* Some pattern-matching *)
-let hd::_ = [1, 2, 3]
+let hd::_ = [1; 2; 3]
 
 let Some message = Some "message"
 
-let [1, snd] = [1, 2]
+let [1; snd] = [1; 2]
 
 type a = int
 and b = float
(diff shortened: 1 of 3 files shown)


tests/samples/expected/ocaml.ml
```diff
@@ -573,6 +573,13 @@ let large_const =
     (long_argument_4 : int) ->
     val
 
+let _ =
+  [
+    1;
+    2;
+    3
```
aspiwack, line 580: To the risk of belabouring a point. What I _actually_ expect is:

```suggestion
    1;
    2;
    3;
```

nbacquey: I agree that's the ideal outcome, but we can't have it yet: see #199 
Should I add the issue to the v0.1 milestone?

aspiwack: I added it to the milestone. We can remove it if the conclusion is that we shall not.

For the record: this should definitely be tackled in a separate PR either way. I was making that point because the PR is titled “List formatting”, which would seem to imply that there is no further issue with lists.

PR #209, opened 2023-02-01
Open output file in write mode

### LINKED ISSUE #207: The `-o` option seems broken in many ways
## Describe the bug

It looks to me like the `-o` option is broken in many ways. It makes Topiary crash when targeting a non-existing file and makes Topiary succeeds silently without writing anything when targeting an existing file.

## To Reproduce

Last commit on `main`: 327540d429148ee6be280844b9dc8a88584bbeb4

- Go in an empty directory.
  ```
  mkdir test
  cd test
  ```

- Get yourself an OCaml test file. (Or any language, really.)
  ```
  echo 'open Foo' > test.ml
  ```

- Run Topiary on this test file, with `-o` targeting a non-existing file name.
  ```
  nix run github:tweag/topiary/327540d429148ee6be280844b9dc8a88584bbeb4 -- -i test.ml -o test2.ml
  ```
  You will get an error of the kind:
  ```
  Error: Writing error
    Caused by: No such file or directory (os error 2)
  ```
  and the return code will be `1`.

- Check that no file named `test2.ml` was created in the process.

- Now create a file named `test2.ml`.
  ```
  touch test2.ml
  ```

- Re-run the same command. Topiary now targets an existing file name.
  ```
  nix run github:tweag/topiary/327540d429148ee6be280844b9dc8a88584bbeb4 -- -i test.ml -o test2.ml
  ```
  You will get no output and the return code will be `0`.

- Now check the content of `test2.ml` and see that it has not changed a single bit.

I have also tried the same protocol in a `nix shell github:tweag/topiary` in case `nix run` was doing weird stuff and it gave the same result.

## Expected behavior

When targeting a non-existing file name, I would expect it to be created. When targeting an existing file name, I would expect it to be overridden, or I would expect Topiary to fail with a message explaining that it doesn't want to override an existing file.

## Environment

 - NixOS unstable with nixpkgs as of https://github.com/NixOS/nixpkgs/commit/1b1f50645af2a70dc93eae18bfd88d330bfbcf7f
 - Version of the code: Last commit on `main` when I write this: 327540d429148ee6be280844b9dc8a88584bbeb4.

nbacquey:
Closes #207

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -79,7 +79,7 @@ fn run() -> FormatterResult<()> {
 
     let mut output: Box<dyn std::io::Write> = match args.output_file.as_deref() {
         Some("-") | None => Box::new(stdout()),
-        Some(file) => Box::new(BufWriter::new(File::open(file)?)),
+        Some(file) => Box::new(BufWriter::new(File::create(file)?)),
     };
 
     let language: Option<Language> = if let Some(language) = args.language {


Xophmeister (approved):
LGTM :+1:

PR #212, opened 2023-02-01
Don't enforce line break after ":" in module bindings

### LINKED ISSUE #191: [OCaml] Line breaks in module types
**Describe the bug**
In interface files, we are too eager to put line breaks before the `sig` keyword. For instance, this common piece of code:
```ocaml
module Foo: sig
  val foo : int
end
```
gets formatted as
```ocaml
module Foo:
sig
  val foo : int
end

```

**Expected behavior**
I expect the original formatting to be kept as is.

**Environment**
 - Ubuntu 22.04
 - main on 2023-01-24 (https://github.com/tweag/topiary/commit/3eefa610bab865a25adaf93a1046937b85836657)

nbacquey:
Allows the following to be kept as it is:
```ocaml
module Foo : sig
  val foo: int
end
```
Closes #191

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -630,7 +630,7 @@
   ":" @append_spaced_softline
 )
 (module_binding
-  ":" @append_spaced_softline
+  ":" @append_input_softline
 )
 (field_declaration
   ":" @append_spaced_softline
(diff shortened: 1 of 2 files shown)


Xophmeister (approved):
:+1:

languages/ocaml.scm
```diff
@@ -630,7 +630,7 @@
   ":" @append_spaced_softline
 )
 (module_binding
-  ":" @append_spaced_softline
+  ":" @append_input_softline
```
torhovland, line 633: This means we're not opinionated on this. Is there any way we could format using our customary single/multi line logic instead?

nbacquey: We could be opinionated, but I think this warrants another PR. Also, if I'm not mistaken it relates to #217

PR #213, opened 2023-02-01
Format input file "in place"

### LINKED ISSUE #208: A CLI option to modify files in place
## Is your feature request related to a problem? Please describe.

I tend to use my formatters in a git context where I am happy to modify the files and let `git` show me the difference that it made. Our other formatter, Ormolu, does support an option `--mode inplace` (`-i` for short) to do that, and I use it all the time.

## Describe the solution you'd like

I would like to be able to run `topiary --in-place my_file.ml` (or even `topiary -i my_file.ml`) so that it reads from and writes to `my_file.ml`.

Alternatively, I would be happy if `-i` and `-o` could target the same file and things would just work as expected. Maybe it does; hard to see at the moment with https://github.com/tweag/topiary/issues/207.

## Describe alternatives you've considered

Common workarounds would be to first write to a file then move it:
```
$ topiary -i my_file.ml -o my_file.ml.out
$ mv my_file.ml.out my_file.ml
```
or to use something like [Sponge](https://linux.die.net/man/1/sponge).

Xophmeister:
This PR changes the way output is persisted to disk, when needs be. Rather than writing directly to the specified output file, it instead writes to a temporary file, which is then copied to the originally-specified output file after formatting. This allows for "in place" formatting -- or the impression of which -- when the input and output refer to the same file.

Notes:
* This is almost certainly not the nicest/most idiomatic way of doing things! There was a large period of me almost-blindly tweaking type signatures until the compiler and linter stopped complaining :sweat_smile: Particularly:
  * As per the [comment here](https://github.com/tweag/topiary/blob/41a14c0b406d357c875485917d5eda9a40b1e681/src/main.rs#L55-L62), it would improve the API if the persistence was automatic. AIUI, this _could_ be implemented as part of my struct's `Drop` trait... Thoughts?
  * My original plan was to implement `AsMut<dyn std::io::Write>` for `OutputFile`. I _believe_ this is an idiomatic way of doing things -- but I'm not sure -- but, either way, I couldn't get it to type check. Hence the [`as_writable` function](https://github.com/tweag/topiary/blob/41a14c0b406d357c875485917d5eda9a40b1e681/src/main.rs#L64-L69).
  * Remember the blind tweaking I mentioned? That explains [this](https://github.com/tweag/topiary/blob/41a14c0b406d357c875485917d5eda9a40b1e681/src/main.rs#L64). I don't get why the explicit lifetime is needed here.

  Any hints or guidance on the above, before this gets merged, would be appreciated :pray:

* There's no harm in performing this switcheroo in all cases; the IO performance hit of writing twice will be trivial. Thus obviating the need for an explicit command line switch.

* Resolves #208


Xophmeister:
Maybe writing into a `Vec<u8>` (i.e., memory) would be better than a temporary file...

Cargo.toml
```diff
@@ -14,6 +14,7 @@ log = "0.4"
 pretty = "0.11"
 pretty_assertions = "1.3"
 regex = "1.6.0"
+tempfile = "3.3.0"
```
torhovland, line 17: As you suggest, we should be able to manage without this.

src/main.rs
```diff
@@ -31,6 +33,42 @@ impl From<SupportedLanguage> for Language {
...
+            Some("-") | None => Ok(Self::Stdout),
+            Some(file) => Ok(Self::Disk {
+                staged: NamedTempFile::new()?,
+                output: file.into(),
+            }),
+        }
+    }
+
+    fn persist(self) -> FormatterResult<()> {
+        // NOTE Maybe it would be better to persist on Drop...or is that too magical?
```
torhovland, line 56: I don't think this is obvious Drop behaviour. What if you just decided you wanted to discard the output file entirely?

torhovland (changes requested):
Could you also add some tests on file outputting?

Cargo.toml
torhovland: By writing to memory instead, I mean.

torhovland: OK, that's a fair point.

torhovland (approved):

torhovland (changes requested):

README.md
```diff
@@ -29,33 +29,37 @@ a development environment.
...
+  Print help information
+
 * `-V`, `--version`\
   Print version information
 
 Language selection is based on precedence, in the following order:
+* A specified query file
 * A specified language
 * Detected from the input file's extension
-* A specified query file
```
nbacquey, line 58: I don't think changing the order in which arguments are parsed changes the priority on language selection. Besides, I think it's better if we keep the current order (for performance reasons for ocaml files)

nbacquey (changes requested):

README.md
Xophmeister: It doesn't. I was just changing the order in the documentation (the README and the `--help`) to match. The priority order hasn't changed in this PR.

src/main.rs
```diff
@@ -31,32 +33,73 @@ impl From<SupportedLanguage> for Language {
...
 
     /// Path to an output file. If omitted, or equal to "-", write to standard
     /// output.
-    #[clap(short, long)]
+    #[clap(short, long, display_order = 2)]
     output_file: Option<String>,
+
+    /// Format the input file in place.
+    #[clap(short, long, display_order = 3)]
+    in_place: bool,
```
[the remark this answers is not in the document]
Xophmeister: Of course; just pushing intermediary commits. The commit message for 879754f mentions it's not yet implemented. I'll make the PR a draft to make it clearer...

Xophmeister:
> Could you also add some tests on file outputting?

@torhovland What kind of tests do you have in mind? Would a simple integration test of the `OutputFile` struct suffice? (I'd need to make it public, but I guess that's fine.)

torhovland:
Something that _would have_ caught that problems in #207 😄

torhovland (approved):

torhovland:
And that `--in-place` actually works.

README.md
nbacquey: I agree that the PR doesn't change the priority order, yet it changes the *documentation* about the priority order (the list just above).

Xophmeister: Because the documentation was wrong. The documentation had it as `language > extension > query`, whereas it should be `query > language > extension`. Right?

Xophmeister:
> Something that _would have_ caught the problems in #207 :smile:

> And that `--in-place` actually works.

Done and done :smile_cat:

Xophmeister:
Oh, FFS, it breaks the CI :roll_eyes:

tests/cli-tester.rs
```diff
@@ -0,0 +1,75 @@
...
+fn read_state(path: &PathBuf) -> String {
+    let mut file = File::open(path).unwrap();
+    let mut contents = String::new();
+    file.read_to_string(&mut contents).unwrap();
+
+    contents
+}
+
+#[test]
+fn test_no_clobber() {
```
torhovland, line 26: Add one more test about being able to create a new output file, and I suppose it's good.

torhovland (approved):
All good as long as you and Nicolas agree about the querying order.

tests/cli-tester.rs
Xophmeister: Will do :+1:

Hopefully I won't generate too much noise trying to track down the CI problem...

README.md
nbacquey: If I'm not mistaken, `query > language > extension` used to be the priority order, but I changed it in #169 when I refactored the way languages are processed. Now query files have the lowest priority.
The reason for it is that `ocaml.scm` now covers two languages (ocaml interface and ocaml implementation), so its `#language!` predicate is interpreted as "try implementation first, then interface if it fails". Using that logic with the highest priority would be a performance hit, because most of the time we can actually know if we're processing implementation or interface by looking at the file extension, hence the priority change.

We can discuss whether this choice is correct of course, but I think it should be done in a separate issue

Xophmeister: Ah, my mistake. I remember you changed it, but I noticed [this comment](https://github.com/tweag/topiary/blob/e21c293acd905d11eb5472400b312a26047823a1/src/main.rs#L75) was inconsistent. I thought that was the correct order and the documentation was out-of-date... I'll fix it in the morning :+1:

Xophmeister: Changed in 33622f0bedc4

nbacquey (approved):

PR #223, opened 2023-02-02
[OCaml] Integrate ppx attributes in field declarations

### LINKED ISSUE #220: [OCaml] Idempotency errors on PPX tags in records
As of 6f71f36

### Snippet 1

The following
```ocaml
type t =
  { foo : bool [@default false];
    bar : float }
```
yields an idempotency check failure caused by a parsing error in the second round. With `--skip-idempotency`, the result is
```ocaml
type t = {
  foo: bool; [@default false] ;
  bar: float;
}
```
Note the double `;` before and after the tag. This is indeed problematic and not valid OCaml.

### Snippets 2 and 2'

Now, when the tag is the last element of the record, consider the following
```ocaml
type t =
  { foo : bool;
    bar : float [@default false] }
```
and its version with an additional `;`
```ocaml
type t =
  { foo : bool;
    bar : float; [@default false] }
```
(GitHub's syntaxic coloration dies a bit here.) I am not entirely sure that the semantics is the same with the tag before and after the semi-colon, one would have to check. Anyways.

The latter works fine and gets formatted as
```ocaml
type t = {
  foo: bool;
  bar: float;
  [@default false]
}
```
I'm not necessarily convinced by the tag going on its own line, but it is valid OCaml AFAIK.

Now the former yields another idempotency check failure, but not because of a parsing error. It first gets formatted as
```ocaml
type t = {
  foo: bool;
  bar: float; [@default false]
}
```
which then gets formatted as
```ocaml
type t = {
  foo: bool;
  bar: float;
  [@default false]
}
```
I personally prefer the first option with the tag on the same line.

nbacquey:
Allows the following:
```ocaml
type t = {
  foo: bool [@default false];
  bar: int
}

type t = {
  foo: bool;
  bar: int [@default 0]
}

type t = {
  foo: bool;
  bar: int [@default 0] (* comment *);
}
```
to be formatted as:
```ocaml
type t = {
  foo: bool [@default false];
  bar: int;
}

type t = {
  foo: bool;
  bar: int [@default 0];
}

type t = {
  foo: bool;
  bar: int [@default 0]; (* comment *)
}
```

Closes #220

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -115,7 +115,6 @@
     "and"
     "as"
     "assert"
-    (attribute)
     "class"
     "downto"
     "else"
@@ -784,22 +783,22 @@
   ] @append_spaced_softline @prepend_spaced_softline
 )
 
-; Put a semicolon delimiter after field declarations, unless they already have
-; one, in which case we do nothing.
-; The semicolon always comes right after the declaration of the new field,
-; before any comment.
+; Put a semicolon delimiter after field declarations and potential ppx attributes,
+; unless they already have one, in which case we do nothing.
+; The semicolon always comes right after the declaration of the new field
+; and attributes, before any comment.
 ; Hence, if there is a comment between the field declaration and the associated ";",
 ; the semicolon is moved before the comment.
 ;
 ; type t =
-;   { mutable position : int (* End-of-line comment *);
+;   { mutable position : int [@default 0] (* End-of-line comment *);
 ;   ...
 ;
 ; is turned into
 ;
 ; type t =
 ;   {
-;     mutable position : int; (* End-of-line comment *)
+;     mutable position : int [@default 0]; (* End-of-line comment *)
 ;     ...
 ;
 (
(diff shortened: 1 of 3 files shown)


Xophmeister (approved):

languages/ocaml.scm
```diff
@@ -784,42 +783,49 @@
...
-    (comment)
+    (attribute)
   ]* @do_nothing
   (#delimiter! ";")
 )
 (
-  (field_declaration) @append_delimiter
+  (field_declaration)
+  .
+  (attribute)+ @append_delimiter
```
torhovland, line 816: Doesn't this break the case where there is no attribute?

torhovland (changes requested):

languages/ocaml.scm
nbacquey: This is handled by the case above:
```scheme
(
  (field_declaration) @append_delimiter
  .
  [
    ";"
    (comment)
    (attribute)
  ]* @do_nothing
  (#delimiter! ";")
)
```

torhovland: What about the comment? I am only seeing:

```
(
  (field_declaration) @append_delimiter
  .
  [
    ";"
    (attribute)
  ]* @do_nothing
  (#delimiter! ";")
)
```

nbacquey: You're right, I copy/pasted the diff too quickly.
The query in your message takes care of the case when there is no attribute.
Do you think there is an example that would get mis-formatted with the new queries?

torhovland: I was worried that `(comment)` on line 810 got lost, but that actually seems like the right thing to do.

PR #234, opened 2023-02-03
Surround :: with spaces in all instances

### LINKED ISSUE #232: [OCaml] Spacing around list concatenation
As of c22f8d5, the following
```ocaml
let rec map f = function
  | [] -> []
  | x :: xs -> f x :: map f xs
```
gets formatted as
```ocaml
let rec map f = function
  | [] -> []
  | x::xs -> f x::map f xs
```
I would have expected it to remain the same, with spacing around `::`. Some people like the absence of spacing, in particular in short patterns such as `x::xs` (or when matching on a pair of lists, `x::xs, y::ys`). However, I believe we should add spacing around `::` for two reasons:
- by consistency with the other operators, for which Topiary adds spacing (eg. arithmetic operators).
- because it makes some things less readable otherwise: my brain would parse `f x::map f xs` instinctively as four tokens, `f`, `x::map`, `f` and `xs`.

Xophmeister:
Resolves #232, modulo OCaml people's approval

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -558,25 +558,6 @@
   ">" @prepend_space
 )
 
-; NOTE This rule is redundant now that spacing is added around ::
-; Keep spacing around the list constructor if it is preceded by
-; a labeled argument, avoiding syntax errors.
-;
-; This is syntactically correct:
-;   foo ~arg :: []
-; This is not:
-;   foo ~arg::[]
-;
-; While both are correct if the argument is not labeled
-;(
-;  (application_expression
-;    (labeled_argument)
-;    .
-;  )
-;  .
-;  "::" @prepend_space @append_space
-;)
-
 ; Softlines. These become either a space or a newline, depending on whether we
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -557,6 +558,7 @@
   ">" @prepend_space
 )
 
+; NOTE This rule is redundant now that spacing is added around ::
```
nbacquey, line 561: I think we should remove this rule altogether, and add a test case which breaks if we remove the spacing (i.e. add `foo ~arg :: []` somewhere in the OCaml input.

nbacquey (changes requested):

languages/ocaml.scm
Xophmeister: That's a good idea :+1: I just left it commented out in this draft, to make the consequences clear.

Xophmeister: Changed in d14e6b05b1a7f4

nbacquey (approved):

PR #237, opened 2023-02-06
Write out the diff to the error log, rather than panicking

### LINKED ISSUE #221: Very different reports of idempotency check failures depending on the reason
The issue https://github.com/tweag/topiary/issues/220 shows two different ways of getting Topiary to reach an idempotency check failure on OCaml code, one by making it yield un-parseable code in the first round, and another one by making it yield two different outputs in the two rounds.

However, those two idempotency check failures get reported very differently, and one of them has an output that looks more like unhandled exceptions and looks a bit scare, or might look scary for random users. See for yourself:

### Snippet 1

```ocaml
type t =
  { foo : bool [@default false];
    bar : float }
```
gives the output
```
Error: The formatter errored when trying to format the code twice (idempotence check).
This probably means that the formatter produced invalid code.
It would be helpful if you logged this error at https://github.com/tweag/topiary/issues/new?assignees=&labels=type%3A+bug&template=bug_report.md
  Caused by: Parsing error between line 2, column 12 and line 2, column 13
```

### Snippet 2

```ocaml
type t =
  { foo : bool;
    bar : float [@default false] }
```
gives the output
```
[2023-02-01T19:06:28Z ERROR topiary] Failed idempotence check
thread 'main' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 type t = {
   foo: bool;
<  bar: float; [@default false]
>  bar: float;
>  [@default false]
 }
 

', src/lib.rs:180:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Actually, it has colours in my terminal:
image

I think this one could be improved. The diff is not a bad thing, but `left` and `right` are not very informative and people might not know what the idempotency check is about. I would expect something closer to the first one, maybe:
```
Error: The formatter errored when trying to format the code twice (idempotence check).
This probably means that the formatter produced invalid code.
It would be helpful if you logged this error at https://github.com/tweag/topiary/issues/new?assignees=&labels=type%3A+bug&template=bug_report.md
  Caused by: Difference between first and second rounds:
  Diff < first / second > :
   type t = {
     foo: bool;
  <  bar: float; [@default false]
  >  bar: float;
  >  [@default false]
   }
```
(The line stating that “this probably means that the formatter produced invalid code” should probably go away in that case.)

Xophmeister:
This PR changes the error message when there is an idempotency failure by outputting the colourised diff, as before, but without panicking.

Error style before change:

image

Error style after change:

image

Resolves #221

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -103,7 +103,7 @@ impl fmt::Display for FormatterError {
             Self::Formatting(_err) => {
                 write!(
                     f,
-                    "The formatter errored when trying to format the code twice (idempotence check).\nThis probably means that the formatter produced invalid code.\n{please_log_message}"
+                    "The formatter failed when trying to format the code twice (idempotence check).\nThis probably means that the formatter produced invalid code.\n{please_log_message}"
                 )
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@ use configuration::Configuration;
 pub use error::{FormatterError, ReadingError, WritingError};
 use itertools::Itertools;
 pub use language::Language;
-use pretty_assertions::assert_eq;
+use pretty_assertions::StrComparison;
 use std::io;
 
 mod atom_collection;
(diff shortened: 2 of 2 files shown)


Niols (approved):
Neat improvement indeed! https://github.com/tweag/topiary/issues/221 is also about consistency between various ways to report errors, I suppose everything is going towards this “log-style” colorised output?

PR #240, opened 2023-02-06
[OCaml] indendation and softlines in parenthesized expressions

### LINKED ISSUE #218: [OCaml] Odd behaviour of parenthesized multi-line expressions
As of 6f71f36,

### Snippet 1

The following
```ocaml
let warnings s =
  let%lwt tunes = List.sort_lwt Tune.compare tunes in
  (match tunes with
   | [] -> add_warning Empty
   | tune::tunes ->
     let _ = foo in
     ());
  Lwt.return true
```
gets formatted as
```ocaml
let warnings s =
  let%lwt tunes = List.sort_lwt Tune.compare tunes in
  (
  match tunes with
  | [] -> add_warning Empty
  | tune::tunes ->
    let _ = foo in
    ());
  Lwt.return true
```
Note how the parenthesis before `match` ends up on its own line while its closing counterpart stays attached to the code of the `match` in question. Also not how the match is aligned on the parenthesis and not indented. I would expect to keep my original formatting or maybe go for:
```ocaml
let warnings s =
  let%lwt tunes = List.sort_lwt Tune.compare tunes in
  (
    match tunes with
    | [] -> add_warning Empty
    | tune::tunes ->
      let _ = foo in
      ()
  );
  Lwt.return true
```

### Snippet 2

The following
```ocaml
let x =
  if true then
    (
      let path = ahaah in
      do_something path
    )
  else
    bar
```
gets formatted as
```ocaml
let x =
  if true then
    ( let path = ahaah in
    do_something path)
  else
    bar
```
Note in particular the space after the opening parenthesis. This space goes away if my first line is `do_something_else;`  and not a `let ... in`. Additionally, note that `do_something` is aligned with the opening parenthesis while I would have expected it indented by one. Actually, I think my initial formatting makes some more sense and looks like how Topiary handles multi-line tuples and lists.

nbacquey:
Formats
```ocaml
let _ = (let x = 42 in
  x)
```
as
```ocaml
let _ =
  (
    let x = 42 in
    x
  )
```
Closes #218

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1080,9 +1080,8 @@
 ;   )
 (parenthesized_expression
   .
-  "(" @append_empty_softline
-  (product_expression) @prepend_indent_start @append_indent_end
-  ")" @prepend_empty_softline
+  "(" @append_empty_softline @append_indent_start
+  ")" @prepend_indent_end @prepend_empty_softline
   .
 )
 ; Parenthesis are optional when using tuples, so scopes must be tied
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -587,6 +587,12 @@ let [|a; _; _|] =
     3;
   |]
 
+let _ =
+  (
+    let x = 42 in
+    x
+  )
+
 (* Showcase the usage of operator bindings *)
 let greetings =
   let (let*) = Option.bind
--- a/tests/samples/input/ocaml.ml
+++ b/tests/samples/input/ocaml.ml
@@ -575,6 +575,9 @@ let [a; _; _] = [1; 2;
 let [|a; _; _|] = [|1; 2;
   3|]
 
+let _ = (let x = 42 in
+  x)
+
 (* Showcase the usage of operator bindings *)
 let greetings =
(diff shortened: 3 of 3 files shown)


languages/ocaml.scm
```diff
@@ -1081,7 +1081,7 @@
 (parenthesized_expression
   .
   "(" @append_empty_softline
-  (product_expression) @prepend_indent_start @append_indent_end
+  _ @prepend_indent_start @append_indent_end
   ")" @prepend_empty_softline
   .
 )
```
Xophmeister, line 1087: The way the indentation is done (i.e., set and unset on each node between the parentheses) is a little unusual. It may be simpler to understand as:
```scm
(parenthesized_expression
  .
  "(" @append_empty_softline @append_indent_start
  ")" @prepend_empty_softline @prepend_indent_end
  .
)
```

Although, admittedly, this doesn't always do what you'd expect...so maybe you tried this already.

nbacquey: I don't remember why I had it that way, but your solution works, and is more elegant. I've amended the PR

Xophmeister (approved):

PR #241, opened 2023-02-06
Correct parenthetical type spacing

### LINKED ISSUE #215: [OCaml] Odd spaces in parenthesized types
As of 6f71f36:

### Snippet 1

The following
```ocaml
type query = (string * string list) list
```
gets formatted as
```ocaml
type query = (string * string list ) list
```
(This also happens without `string *`, even if I would probably then not use parentheses.)
(This also happens without the first `list`.)

### Snippet 2

The following
```ocaml
type x = ('any Slug.t -> bool) -> float
```
gets formatted as
```ocaml
type x = ( 'any Slug.t -> bool) -> float
```
(This does not happen without `'any`.)

Xophmeister:
Resolves #215

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -350,13 +350,16 @@
   "type" @prepend_space
 )
 (
-  [
-    "("
-    "["
-  ]* @do_nothing
+  (_)
   .
   (type_variable) @prepend_space
 )
+(
+  (type_variable)
+  .
+  "." @append_space
+)
+
 (
   "("* @do_nothing
   .
@@ -536,6 +539,8 @@
 )
 (constructed_type
   (_) @append_space
+  .
+  ")"? @do_nothing
   .
   (_)
 )
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -828,3 +828,7 @@ end
 (* Ensure labelled arguments are correctly spaced *)
 let _ =
   foo ~arg :: []
+
+(* Correct spacing inside parenthetical type definitions *)
+type query = (string * string list) list
+type x = ('any Slug.t -> bool) -> float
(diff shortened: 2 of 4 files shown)


Xophmeister:
Yikes... Let me fix the merge conflicts

languages/ocaml.scm
```diff
@@ -350,13 +350,16 @@
...
+  (_)
   .
   (type_variable) @prepend_space
 )
+(
+  (type_variable)
+  .
+  "." @append_space
+)
+
```
nbacquey, line 362: I do really understand the reason for this modification, and the one above. Could you clarify?

Xophmeister: (Disclaimer: I don't know the OCaml terminology, so I'll have to spell things out verbosely :sweat_smile:)

The original rule:
```scm
(
  [
    "("
    "["
  ]* @do_nothing
  .
  (type_variable) @prepend_space
)
```
..._should_ have been enough to address the spacing in the `type t: ('x -> x)`-case. However, it clearly wasn't working. As such, I changed it to this:

```scm
(
  (_)
  .
  (type_variable) @prepend_space
)
```

That worked, but it broke the spacing of type variables followed by a full stop, which is an anonymous node. i.e.:
```ocaml
(* This... *)
let my_const : 'a 'b. 'a -> 'b -> 'a = Fun.const

(* Got formatted as... *)
let my_const : 'a 'b.'a -> 'b -> 'a = Fun.const
```

I tried changing the first half of the tuple pattern to `[(_) "."]`, but that didn't work either. As such, I had to write an explicit rule for it to target type variables followed by a full stop, hence:
```scm
(
  (type_variable)
  .
  "." @append_space
)
```

There may be a better solution, but this worked.

nbacquey: I think that, for the sake of the maintainability of the query file, it's better to remove from existing rules than add new rules, and I think the problem can be addressed by the former. Let me draft something

nbacquey: My attempt is at #258

Xophmeister: That looks great :100: I'll close this PR

Xophmeister:
Subsumed by #258

PR #258, opened 2023-02-07
[OCaml] fix and simplify spacing policy

### LINKED ISSUE #215: [OCaml] Odd spaces in parenthesized types

### LINKED ISSUE #245: Missing space after parenthesis in types
As of f97a163, the following
```ocaml
type ('a) foo = 'a
```
gets formatted as:
```ocaml
type ('a )foo = 'a
```
The space after `'a` probably comes from #215, and #241 should fix it. However, note the absence of space between the closing parenthesis and `foo`.

The example is rather stupid, but think of types such as:
```ocaml
type ('a, 'b) result =
  | Ok of 'a
  | Error of 'b
```
getting formatted as:
```ocaml
type ('a , 'b )result =
  | Ok of 'a
  | Error of 'b
```
with no space after the parenthesis and before `result`. (There is also an additional space before the comma and I'm reporting it in another issue.)

### LINKED ISSUE #246: Additional space before comma in types
As of f97a163, the following
```ocaml
type foo = (int, int) result
```
gets formatted as
```ocaml
type foo = (int , int ) result
```
The space before the closing parenthesis is probably #215. However, there is an additional space before the comma. Note that it also happens on the left-hand side of the type definition. For instance
```ocaml
type ('a, 'b) result =
  | Ok of 'a
  | Error of 'b
```
gets formatted as 
```ocaml
type ('a , 'b )result =
  | Ok of 'a
  | Error of 'b
```
that includes this extra space. (The lack of space before result has been reported in #245.)

### LINKED ISSUE #247: No space between function name and explicit type quantification
As of f97a163, the following
```ocaml
let id (type s) (x : s) : s = x
```
gets formatted as
```ocaml
let id(type s) (x : s) : s = x
```
without space between `id` and `(type s)`. This is still valid OCaml and everything (even if GitHub's coloration gets confused), but there should probably be a space.

### LINKED ISSUE #250: No space after universal quantification in records
As of f97a163, the following
```ocaml
type foo = { a : 'a. ('a, mandatory) arg -> 'a; }
```
gets formatted as
```ocaml
type foo = { a: 'a.('a , mandatory ) arg -> 'a; }
```
I think there should be a space after the universal quantification introduced by `'a.`, because as it is now it gives a wrong idea of the priority of quantification over the arrow, imo.

nbacquey:
Fixes several spacing policies, mainly by removing specific spacing rules, and adding items to the big "list of stuff that must be separated by spaces" query.

Among others, allows correct formatting of
```ocaml
type query = (string * string list) list
type x = ('any Slug.t -> bool) -> float
let id (type s) (x : s) : s = x
type foo = { a: 'a. ('a, mandatory) arg -> 'a; }
type foo = (int, int) result
```

Closes #215
Closes #245
Closes #246
Closes #247
Closes #250

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -350,14 +350,6 @@
   "type" @prepend_space
 )
 (
-  [
-    "("
-    "["
-  ]* @do_nothing
-  .
-  (type_variable) @prepend_space
-)
-(
   "("* @do_nothing
   .
   "val" @prepend_space
@@ -401,8 +393,6 @@
 ;   type a b. a: a -> b: b -> a =
 ;   fun ~a ~b -> a
 (polymorphic_type
-  (abstract_type)
-  .
   "." @append_space
 )
 
@@ -424,12 +414,15 @@
     (number)
     (parenthesized_expression)
     (parenthesized_pattern)
+    (polymorphic_variant_type)
     (quoted_string)
     (string)
     (tag)
+    (type_constructor)
     (type_constructor_path)
     (typed_expression)
     (type_variable)
+    (value_name)
     (value_path)
     (value_pattern)
     ")"
@@ -447,6 +440,7 @@
   ]* @do_nothing
   .
   [
+    (abstract_type)
     (boolean)
     (character)
     (class_name)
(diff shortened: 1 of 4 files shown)


Xophmeister (approved):
Nice :+1:

PR #259, opened 2023-02-07
[OCaml] Scoped softlines for sequences and infix expressions

### LINKED ISSUE #251: Weird behaviour of multiline mode in sequence of operators.
As of f97a163, the following
```ocaml
let x = foo || foo || foo
        || bar || bar || bar || bar
```
gets formatted as
```ocaml
let x =
  foo
  || foo
  || foo
  || bar || bar || bar || bar
```
I think it's not the first time I've seen behaviours like this one, like it happened in lists and maybe even records before.

Also happens with sequence:
```ocaml
let x = foo; foo; foo;
        bar; bar; bar; bar
```
becomes
```ocaml
let x =
  foo;
  foo;
  foo;
  bar; bar; bar; bar
```

nbacquey:
Formats
```ocaml
let x = foo || foo
  || bar || bar

let x = foo; foo;
  bar; bar
```
as
```ocaml
let x =
  foo
  || foo
  || bar
  || bar

let x =
  foo;
  foo;
  bar;
  bar
```
Closes #251

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1112,6 +1112,54 @@
   (#scope_id! "function_type")
 )
 
+; Allow softlines in infix expressions, such as
+; let b =
+;   foo
+;   || bar
+;   || baz
+
+; As above, infix expressions are nested grammar elements, so we must identify the
+; top-level one: it is the one that is not preceded by an infix operator.
+; We only consider the common logic operators, as not to mess with arithmetic expressions
+(
+  (infix_operator
+    [
+      "||"
+      "&&"
+    ]
+  )? @do_nothing
+  .
+  (infix_expression) @begin_scope @end_scope
+  (#scope_id! "infix_expression")
+)
+(infix_expression
+  (infix_operator
+    [
+      "||"
+      "&&"
+    ]
+  ) @prepend_spaced_scoped_softline
+  (#scope_id! "infix_expression")
+)
+
+; Allow softlines in sequences, such as
+; let b =
+;   foo;
+;   bar;
+;   baz
+; As above, sequences are nested grammar elements, so we must identify the
+; top-level one: it is the one that is not preceded by a ";".
+(
+  ";"? @do_nothing
+  .
+  (sequence_expression) @begin_scope @end_scope
+  (#scope_id! "sequence_expression")
+)
+(sequence_expression
+  ";" @append_spaced_scoped_softline
+  (#scope_id! "sequence_expression")
+)
+
 ; Indent and add softlines in lists and arrays, such as
 ; let _ =
 ;   [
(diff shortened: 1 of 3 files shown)


Xophmeister (approved):

languages/ocaml.scm
```diff
@@ -1112,6 +1112,54 @@
...
+
+; As above, infix expressions are nested grammar elements, so we must identify the
+; top-level one: it is the one that is not preceded by an infix operator.
+; We only consider the common logic operators, as not to mess with arithmetic expressions
+(
+  (infix_operator
+    [
+      "||"
+      "&&"
+    ]
```
aspiwack, line 1129: There are more than two infix operators in Ocaml (in fact there are infinitely many). Can we avoid having to list them all?

nbacquey: This is deliberate, see the comment above: `We only consider the common logic operators, as not to mess with arithmetic expressions`.
There is a nice illustration in the test files:
```ocaml
let _ = len < 0 || srcoff < 0 || srcoff > src.position - len
  || dstoff < 0 || dstoff > (Bytes.length dst) - len
```
If we add a softline on every infix operator, it gets formatted as
```ocaml
let _ =
  len < 0
  || srcoff < 0
  || srcoff > src.position - len
  || dstoff < 0
  || dstoff
  > (Bytes.length dst)
  - len
```
Which is not the desired output.

aspiwack: I see. But the question still stands. There are infinitely many operators in Ocaml, and we want to format them all. They have known priorities, can we use these priorities (or something in the syntax tree's structure) to format properly? (like each priority, or nesting level, or something should be a different scope, shouldn't it?).

Maybe this is too hard for 0.1. But this is definitely something that needs figuring out.

nbacquey: I agree that each operator priority should define its own scope, ideally. However, I think we don't have access to the priorities, at the level at which we operate. We only have access to the syntax tree, which, for `a || b || c > d - e`, is
```
{ Node expression_item }
  { Node infix_expression }
    { Node value_path }
      { Node value_name }
    { Node infix_operator }
      { Node || }
    { Node infix_expression }
      { Node value_path }
        { Node value_name }
      { Node infix_operator }
        { Node || }
      { Node infix_expression }
        { Node value_path }
          { Node value_name }
        { Node infix_operator }
        { Node infix_expression }
          { Node value_path }
            { Node value_name }
          { Node infix_operator }
            { Node - }
          { Node value_path }
            { Node value_name }
```
The fact that the operator priorities are not exposed (or even known?) by the tree-sitter grammar is quite hindering.

aspiwack: I thought it was just the priority of the first character in the symbol, which would be simpler for us (though not necessarily simple).

> We only have access to the syntax tree, which, for `a || b || c > d - e`, is

I guess that the best solution will look like making a scope per list of operators of the same name. We would have one scope for `||`, then one (nested) for `>` and one nested for `-`. We will probably want to build logic for this. Which is kind of what  #98 was proposing to do, I believe.

PR #260, opened 2023-02-07
Improvements to error handling and determining the location of query files

### LINKED ISSUE #256: Writing error when reading from standard input
### To Reproduce

```console
$ echo 'open Foo' | nix run github:tweag/topiary/fe8b859 -- --language ocaml
Error: Writing error
  Caused by: No such file or directory (os error 2)
```

Same behaviour with `nix shell github:tweag/topiary/fe8b859` followed by `echo 'open Foo' | topiary -l ocaml`.

Same behaviour with `--output-file test.ml`.

Same behaviour with `--input-file test.ml` (with an existing `test.ml`).

### Expected behavior

```console
$ echo 'open Foo' | nix run github:tweag/topiary/fe8b859 -- --language ocaml
open Foo
```

### Environment
 - OS name + version: NixOS with https://github.com/NixOS/nixpkgs/commit/2caf4ef5005ecc68141ecb4aac271079f7371c44
 - Version of the code: fe8b859

Xophmeister:
This PR:

* [x] Overhauls the I/O error handling

  We cannot distinguish, from `std::io::Error`, between a read error or a write error. This was returning invalid `WritingError`s, e.g., when attempting to read a query file that did not exist.

  As such, `ReadingError` and `WritingError` have been consolidated into `IoError`, which has two subtypes: `Filesystem` (which correlates directly to `std::io::Error`), and `Generic` (which is a "catch all" for everything else).

  Error outputting is now done with logging, rather than `eprintln!`.

* [x] Updates `Language::query_path`

  Using the same priority order as before (runtime envvar > build time envvar > `./languages`), also check if the query file exists. If it does, at whatever level, then all is well; otherwise, return an error.

* [x] Miscellany
  * [x] The `From<io::IntoInnerError>` implementation has been genericised to accept _any_ inner `Write`r. This allows us to properly handle errors when we persist the output to disk.
  * [x] The CLI integration tests no longer need to explicitly set the `TOPIARY_LANGUAGE_DIR` environment variable.
  * [x] Updated the Nix derivation to correct the `TOPIARY_LANGUAGE_DIR` setting and the documentation to match.

Resolves #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,13 @@ The project can be built and installed with Cargo from the repository
 directory:
 
 ```bash
-export TOPIARY_LANGUAGE_DIR="$(pwd)/languages"
 cargo install --path .
 ```
 
-The `TOPIARY_LANGUAGE_DIR` variable can be set either at compile time or runtime. It should point to the directory where Topiary's language query files (`.scm`) are located.
+The `TOPIARY_LANGUAGE_DIR` variable can be set either at build time or
+runtime. It should point to the directory where Topiary's language query
+files (`.scm`) are located. Otherwise, Topiary will fall back to the
+`languages` directory in the current working directory.
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for details on setting up
 a development environment.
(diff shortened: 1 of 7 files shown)


nbacquey (approved):
Looks good, thanks for taking care of this!

Issue #203, opened 2023-01-30
Errant blank lines added after runs of statements


Xophmeister:
**Describe the bug**
New lines are inserted after "units of execution". After the last of which, in a run, this will be rendered as a blank line. This is particularly acute in indented sub-blocks.

**To Reproduce**
Run the following through the Bash formatting rules:

```bash
while true; do
  command
done
```

**Expected behavior**
The above code example should not be changed by the formatter. Currently, it is formatted as follows:

```bash
while true; do
  command

done
```

**Environment**
 - OS name + version: NixOS 22.05
 - Version of the code: 12f1894

**Additional context**
Not for v0.1

The [Bash Tree Sitter grammar](https://github.com/tree-sitter/tree-sitter-bash) inserts anonymous `\n` nodes into the tree after items that should be followed by a new line (see discussion in #186). This is _probably_ the best place to start looking as the root of the cause...

Xophmeister:
(See https://github.com/tree-sitter/tree-sitter-bash/issues/155 for analysis)

patrislav1:
> The [Bash Tree Sitter grammar](https://github.com/tree-sitter/tree-sitter-bash) inserts anonymous `\n` nodes into the tree after items that should be followed by a new line (see discussion in #186). This is _probably_ the best place to start looking as the root of the cause...

the spurious LF was apparently [fixed half a year ago](https://github.com/tree-sitter/tree-sitter-bash/pull/198), is there a special reason why the bash grammar is [pinned to an older revision](https://github.com/tweag/topiary/blob/main/Cargo.toml#L77)?

torhovland:
I think the reason we pinned some versions was to have all the grammars depend on the same version of Tree-sitter. It's possible this can all be reviewed and bumped by now.

Xophmeister:
@patrislav1 the Bash Tree-sitter grammar has also changed significantly enough, in the meantime, that the formatting queries written for Bash are no longer valid with the newer grammar (i.e., it will immediately fail with syntax errors). In general, formatting queries need to be pinned to compatible grammars.

patrislav1:
Yep, I also found after un-pinning the old `tree-sitter-bash` revision and using the most recent one, topiary would abort with query errors.

PR #264, opened 2023-02-08
Remove leading and trailing blank lines

### LINKED ISSUE #48: Remove blank lines at the start or end
We may get them as a result of appending or prepending softlines at the root level, e.g. before any initial comments.

Xophmeister:
This PR changes the whitespace trimmer such that it:

1. Trims whitespace from the end of each formatted line (as before)
2. Trims any leading and/or trailing new lines
3. Reinstates the new line at EOF.

Note: The formatted output assumes Unix-style new lines (`\n`). That's no change from before this PR, but this could probably be improved to accommodate the different platforms'/input file's new line expectations.

Resolves #48

### DIFF
--- a/tests/samples/expected/bash.sh
+++ b/tests/samples/expected/bash.sh
@@ -150,4 +150,3 @@ export xyzzy=$(
   something
   another_thing --foo
 )
-
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,7 +145,7 @@ pub fn formatter(
     // Pretty-print atoms
     log::info!("Pretty-print output");
     let rendered = pretty::render(&atoms[..], configuration.indent_level)?;
-    let trimmed = trim_trailing_spaces(&rendered);
+    let trimmed = trim_whitespace(&rendered);
 
     if !skip_idempotence {
         idempotence_check(&trimmed, &query, language)?
@@ -162,8 +162,11 @@ fn read_input(input: &mut dyn io::Read) -> Result<String, io::Error> {
     Ok(content)
 }
 
-fn trim_trailing_spaces(s: &str) -> String {
-    Itertools::intersperse(s.split('\n').map(|line| line.trim_end()), "\n").collect::<String>()
+fn trim_whitespace(s: &str) -> String {
+    // Trim whitespace from the end of each line,
+    // then trim any leading/trailing new lines,
+    // finally reinstate the new line at EOF.
+    format!("{}\n", s.lines().map(str::trim_end).join("\n").trim())
 }
 
 fn idempotence_check(


Xophmeister:
> Note: The formatted output assumes Unix-style new lines (`\n`). That's no change from before this PR, but this could probably be improved to accommodate the different platforms'/input file's new line expectations.

Looks like this probably doesn't matter: https://internals.rust-lang.org/t/formatln-format-platform-specific-newline/13798/9

tests/samples/expected/bash.sh
```diff
@@ -150,4 +150,3 @@ export xyzzy=$(
   something
   another_thing --foo
 )
-
```
nbacquey, line 153: Why is the EOF newline deleted here?

Xophmeister: That's not the EOF new line; it's an additional, trailing new line that was added to make the tests pass:

```console
$  xxd tests/samples/expected/bash.sh | tail -3
00000610: 736f 6d65 7468 696e 670a 2020 616e 6f74  something.  anot
00000620: 6865 725f 7468 696e 6720 2d2d 666f 6f0a  her_thing --foo.
00000630: 290a                                     ).
```

nbacquey: I fear I don't understand. I think we want the newline here, in the expected file, don't we?
If anything, we could remove the trailing newline from the input file, and check that it has been added in the expected file.

Xophmeister: My understanding of #48 is that some rules end up adding leading or trailing blank lines in the output, which are undesirable but difficult to remove using just the query language (e.g., you can't anchor against the start/end of a file, AFAIK), thus post-processing should be used to remove them instead.

In the Bash queries, the output gets a _lot_ of unnecessary blank lines (see #203), including at EOF (not including the final, necessary `\n`). Thus it's common -- as in the integration test case -- to get an output that ends with two `\n`s. This PR, under my (potentially incorrect :sweat_smile:) understanding of #48, strips them off and then adds the final `\n`; hence the change to the expected output to match.

In general, if my source file looks something like this:
```
.
...
XX..
XXX.
X
XX...
.
```
where `X` is content and `.` is trailing whitespace -- either trailing spaces/etc. on a line, or leading/trailing blank lines -- then I think I would _always_ want it transformed into:
```
XX
XXX
X
XX
```
That's what this PR achieves. Previously, the output would have looked like this:
```
.
.
XX
XXX
X
XX
.
```
(i.e., Just the trailing spaces on each line are removed.)

nbacquey: I think that, in your example, I would like the output to be
```
XX
XXX
X
XX
.
```
I thought trailing blank lines were desirable (`git` berates me when I forget mine).

Xophmeister: Here I'm making a distinction between "blank lines" and "new lines". The previous trimming function outputs this:

```
\n
\n
XX\n
XXX\n
X\n
XX\n
\n
```

The change in this PR outputs this:

```
XX\n
XXX\n
X\n
XX\n
```

i.e., The trailing blank line is removed, but the EOF new line is preserved. (Apologies for the confusion!)

nbacquey: Alright, everything is fine then!

nbacquey (approved):

tests/samples/expected/bash.sh
Xophmeister: :tada:

PR #274, opened 2023-02-09
[OCaml] Support type constraints and polymorphic syntax

nbacquey:
Add correct support for the following:
```ocaml
type (+'meth, 'prefix, 'params, 'query, 'input, 'output) service =
  ('meth, 'prefix, 'params, 'query, 'input, 'output, error) raw
  constraint 'meth = [< meth]
```

### DIFF
--- a/tests/samples/expected/ocaml.mli
+++ b/tests/samples/expected/ocaml.mli
@@ -4444,9 +4444,11 @@ module Kind: sig
   type 'a double_consensus_operation_evidence =
     | Double_consensus_operation_evidence
 
-  type double_endorsement_evidence = endorsement_consensus_kind double_consensus_operation_evidence
+  type double_endorsement_evidence =
+    endorsement_consensus_kind double_consensus_operation_evidence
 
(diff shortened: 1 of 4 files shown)


tests/samples/expected/ocaml.mli
```diff
@@ -5372,15 +5375,15 @@ module Token: sig
   val transfer_n :
     ?origin: Receipt.update_origin ->
     context ->
-    ([<source] * Tez.t) list ->
-    [<sink] ->
+    ( [< source] * Tez.t) list ->
```
Xophmeister, line 5378: Should there be a space here, after the open paren?
```suggestion
    ([< source] * Tez.t) list ->
```

Xophmeister (approved):

tests/samples/expected/ocaml.mli
Xophmeister: (By "should", I mean, "is this what we want", rather than "is this what we expect".)

nbacquey: There shouldn't, well spotted

Issue #219, opened 2023-02-01
[OCaml] Handling of multiline function arguments


Niols:
As of 6f71f36,

```ocaml
let consolidate_query_parameters
    (uri : Uri.t)
    (body : Cohttp_lwt.Body.t)
    : QueryParameters.t Lwt.t
  =
  let%lwt high = QueryParameters.from_body body in
  let low = QueryParameters.from_uri uri in
  Lwt.return (QueryParameters.append ~high ~low)
```
gets formatted as
```ocaml
let consolidate_query_parameters
  (uri : Uri.t)
  (body : Cohttp_lwt.Body.t) :
  QueryParameters.t Lwt.t =
  let%lwt high = QueryParameters.from_body body in
  let low = QueryParameters.from_uri uri in
  Lwt.return (QueryParameters.append ~high ~low)
```
This makes very unclear the separation between the arguments and the body of the function, especially because my type is qualified and `QueryParameters.t Lwt.t =` requires the eye to go quite far (either `=` if we want to be sure or `.t` if we guess that it's a type, then) to understand that this line is the return type of the function.

As a simpler-for-the-eye example (the previous was mostly here to make my point about readability), one can consider:
```ocaml
let f
  foo
  bar :
  int =
  do_things
```

I think several things contribute to the reason why this is confusing:
- the `:` at the end of the line is quite discreet and does not give us a clear cue that we've now reached a return type and not yet another argument
- the `=` at the end of the line is quite discreet and does not give us a clear cue that we've now reached the end of the arguments and are starting to read the body of the function
- the indentation remaining the same does not suggest in any way a change of context.

I think that at least one of those points should be taken care of, and maybe even all of them. At least:
- `:` could be at the beginning of the last line, giving a clear cue that this one has a different nature
- `=` could be on its own line, effectively separating the arguments from the body
- my Emacs mode indents the arguments by one more level, making it more obvious that some context is changing when suddenly the indentation level drops.

PR #275, opened 2023-02-09
[OCaml] Fix failing idempotency in mono-line function

nbacquey:
The problem was that
```ocaml
let foo =
  fun x -> x
in
foo
```
would get formatted as
```ocaml
let foo = fun x -> x
in
foo
```
which would itself be formatted as
```ocaml
let foo = fun x -> x in
foo
```

My fix is not completely satisfactory, since it relies on an unopinionated `@append_input_softline`.
Feel free to discuss better solutions.


Xophmeister:
I think this is a reasonable compromise, but I'm not sure about the look of this as a non-(and therefore unreliable source)-OCaml person:

```ocaml
let my_const
    : type a b. a: a -> b: b -> a
  =
  fun ~a ~b -> a
```

aspiwack:
@nbacquey 

I think that a more typical formatting would be

```ocaml
let f = fun x ->
  y
in
```

But I assume we don't have quite the right infrastructure to make this sort of things happen? There is a whole logic on hanging bits in Ormolu to handle this sort of cases I think.

@Xophmeister 



>  I'm not sure about the look of this[…]
> 
> ```ocaml
> let my_const
>     : type a b. a: a -> b: b -> a
>   =
>   fun ~a ~b -> a
> ```

Well, it's weird that the function doesn't get indented after the `=`, what's happening there?

nbacquey:
@aspiwack I think we can achieve your proposed formatting with custom scopes, let me try something.

@Xophmeister This choice of formatting is what was discussed in #219. It looks alright to me, could you pinpoint what's bothering you?

Xophmeister:
> @Xophmeister This choice of formatting is what was discussed in #219. It looks alright to me, could you pinpoint what's bothering you?

If that was the decided-upon format, then that's fine by me :slightly_smiling_face:

nbacquey:
The new version has an opinionated formatting, and also solves an additional similar issue with single-lined `function`

tests/samples/expected/ocaml.ml
```diff
@@ -596,6 +597,19 @@ let _ =
...
 
+let _ =
+  let foo = fun x ->
+      x
+  in
+  foo
+
+let _ =
+  let foo = function
+    true -> false
```
aspiwack, line 608: Ugh. Let's add the preceding `|` . Probably not needed for v0.1 (we'll have the same story as with list terminators, I imagine: we only want the `|` for multi-line only, and remove it for single line. Probably.

(this is definitely not a comment to take into account for this PR.)

aspiwack (approved):

PR #280, opened 2023-02-10
Handling of indentation for modules with a `with`

### LINKED ISSUE #248: Handling of indentation for modules with a `with`
As of f97a163, the following
```ocaml
module MFloat : SERIALISABLE
  with type t = float =
struct
  type t = float [@@deriving yojson]
  let _key = "float"
end
```
gets formatted as
```ocaml
module MFloat: SERIALISABLE with
  type t = float = struct
  type t = float [@@deriving yojson]
  let _key = "float"
end
```
I think this yields the same considerations as #219 in term of readability: the `struct` gets lost, and the `type t = float` from the `with` statement can get confused for a declaration in the module.

In a similar vain, I am not very convinced by 
```ocaml
module MOption
  (A: SERIALISABLE)
  : SERIALISABLE = struct
  type t = A.t option [@@deriving yojson]
  let _key = A._key ^ "-option"
end
```
with no visual clue about when module argument/types end and when the module definition starts.

Xophmeister:
* Resolves #248 
* [x] Based on #282 (merge first to avoid conflicts)

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1386,35 +1386,25 @@
 ; end
 (module_binding
   (module_name) @append_indent_start @begin_scope
-  "=" @prepend_indent_end @end_scope
+  "=" @prepend_empty_scoped_softline @prepend_indent_end @end_scope
   (#scope_id! "module_binding_before_equal")
 )
 ; if a module binding has no equal sign, everything enters the scope
 (module_binding
   (#scope_id! "module_binding_before_equal")
   (module_name) @append_indent_start @begin_scope
-  "="? @do_nothing
   [
-    (signature)? @do_nothing
     [
       (functor_type)
       (module_type_constraint)
     ] @append_indent_end @end_scope
   ]
-  .
+  "="? @do_nothing
 )
 (module_binding
   (module_name) @append_empty_scoped_softline
   (module_parameter) @append_spaced_scoped_softline
   (#scope_id! "module_binding_before_equal")
-)
-
-; Input softline after an equals sign that is followed by a structure
-; in a module binding
-(module_binding
-  "=" @append_input_softline
-  .
-  (structure)
 )
 
 ; Try block formatting
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1409,6 +1409,14 @@
   (#scope_id! "module_binding_before_equal")
 )
 
+; Input softline after an equals sign that is followed by a structure
+; in a module binding
+(module_binding
+  "=" @append_input_softline
+  .
+  (structure)
+)
+
 ; Try block formatting
 ; A soft linebreak after the "try" (potentially "try%ppx") and one after the "with".
 (try_expression
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -923,3 +923,11 @@ let x =
     | false -> 0.
     | true -> 1.
   ]
+
+(* New line for structures in module definitions if they appear in the input *)
+module MFloat: SERIALISABLE with
+    type t = float =
+struct
+  type t = float [@@deriving yojson]
+  let _key = "float"
+end
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -926,8 +926,8 @@ let x =
 
 (* New line for structures in module definitions if they appear in the input *)
 module MFloat: SERIALISABLE with
-    type t = float =
-struct
+  type t = float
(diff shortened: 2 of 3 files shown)


Xophmeister:
This is a work in progress.

I have added the input softline between the `=` and a structure, in a module definition. However, somewhere between f97a163...f7d1004, a rule was introduced that breaks the indentation. (Amongst some more subtle differences.).

Expected:
```ocaml
module MFloat : SERIALISABLE
  with type t = float =
struct
  type t = float [@@deriving yojson]
  let _key = "float"
end
```

Actual (as of ad0b2b6):
```ocaml
module MFloat: SERIALISABLE with
    type t = float =
struct
  type t = float [@@deriving yojson]
  let _key = "float"
end
```

An extra pair of eyes would be appreciated on this one :eyes:

Xophmeister:
The indentation bug was introduced in 12bf707e78e4, with this rule:

https://github.com/tweag/topiary/blob/12bf707e78e47497f1269d4d96a77477df2315fe/languages/ocaml.scm#L1290-L1303

Xophmeister:
@nbacquey I can fix the indentation if I add a `@do_nothing` rule against any `with` anonymous node that appears:

```scm
(module_binding
  (#scope_id! "module_binding_before_equal")
  (module_name) @append_indent_start @begin_scope
  "="? @do_nothing
  [
    (signature)? @do_nothing
    [
      (functor_type)
      (module_type_constraint
        "with"? @do_nothing
      )
    ] @append_indent_end @end_scope
  ]
  .
)
```

However, that then breaks the indentation on other examples with a `with`:
```diff
 module MUnit: SERIALISABLE with
-  type t = unit
+type t = unit

     (* ...snip... *)
     module Protocol_implementation:
-      PVM.S with
-      type context = Context.t
-      and type state = Context.tree
-      and type proof = Context.Proof.tree Context.Proof.t
+    PVM.S with
+    type context = Context.t
+    and type state = Context.tree
+    and type proof = Context.Proof.tree Context.Proof.t
```

I _think_, therefore, that there was nothing wrong with this rule, as such, but the double indentation is getting applied elsewhere. However, the tree for the `type t = float` is:

* `(constrain_type)`
  * Anonymous `type`
  * `(type_constructor_path)`
    * `(type_constructor)`
  * Anonymous `=`
  * `(type_constructor_path)`
    * `(type_constructor)`

I don't see any rule that would match any part of this subtree... Ideas?

nbacquey:
The problem is indeed that the rule you identified applies, where it should not (it's designed to apply only for module bindings without an `=` sign, but it obviously fails at that. I'm thinking about a way to fix it.

On another note, it allowed me to identify yet another problem with tree-sitter query grammar, which I'll be documenting

nbacquey:
This new version of the rule fixes the problem:
```scheme
(module_binding
  (#scope_id! "module_binding_before_equal")
  (module_name) @append_indent_start @begin_scope
  [
    (functor_type)
    (module_type_constraint)
  ] @append_indent_end @end_scope
  "="? @do_nothing
)
```
I removed the `(signature)` part, because it did nothing to the tests. The real modification comes from pushing the `=` at the end.

Also, instead of adding a new unopinionated rule to keep input line breaks, I would prefer to be opinionated and add a scoped line break using the query just above the one I modified. Do you want to handle it @Xophmeister ? Otherwise I can draft something.

languages/ocaml.scm
```diff
@@ -1386,22 +1386,20 @@
...
   (#scope_id! "module_binding_before_equal")
   (module_name) @append_indent_start @begin_scope
-  "="? @do_nothing
   [
-    (signature)? @do_nothing
     [
       (functor_type)
       (module_type_constraint)
     ] @append_indent_end @end_scope
   ]
```
nbacquey, line 1401: We can remove double brackets here

Xophmeister:
Thanks, @nbacquey. That updated query certainly worked :+1: and I put in the scoped softline in 37c5b47; thus un-drafting this PR. Note that I could only prepend the scoped softline (appending didn't do anything), so the final output looks like this:

```ocaml
module MFloat: SERIALISABLE with
  type t = float
= struct
  type t = float [@@deriving yojson]
  let _key = "float"
end
```

languages/ocaml.scm
Xophmeister: Done in 37c5b47 :ok_hand:

nbacquey:
Prepending is fine with me. Appending doesn't work because the custom scope ends just after the `=`, so the scoped softline would be out of scope (which should trigger a warning message).

PR #278, opened 2023-02-10
Indentation of multi-line types in PPX syntax

Xophmeister:
* Resolves #231 
* [x] Based on #275 (merge first to avoid conflicts)

PR #282, opened 2023-02-13
Fix various PPX indentation issues

### LINKED ISSUE #231: [OCaml] Indentation of multi-line types in PPX syntax
As of c22f8d5, the following
```ocaml
let h =
  [%madcast: float ->
    bool]
```
gets formatted as
```ocaml
let h =
  [%madcast: float ->
  bool]
```
losing the indentation. I would basically expect it to remain the same.

### LINKED ISSUE #233: [OCaml] Indentation of `function` and `match` cases in PPX syntax
As of c22f8d5, the following
```ocaml
let x =
  [%expr function
    | false -> 0.
    | true -> 1.]
```
gets formatted as
```ocaml
let x =
  [%expr function
  | false -> 0.
  | true -> 1.]
```
losing the indentation. I would have expected it to remain the same, probably.

---

As for `match`, the following
```ocaml
let x =
  [%expr match x with
    | false -> 0.
    | true -> 1.]
```
gets formatted as
```ocaml
let x =
  [%expr
  match x with
  | false -> 0.
  | true -> 1.]
```
I would have expected it to remain the same or, if one insists to have `match` not go on the line of the `[%expr`, the I would expect this
```ocaml
let x =
  [%expr
    match x with
    | false -> 0.
    | true -> 1.]
```
with indentation anyways.

Xophmeister:
This PR supersedes #278 and #279, to simplify the commit history after bad rebases:

* Resolves #231 
* Resolves #233
  
Notes:
* The `function`/`match` rule looks a little bit too general and may "over-format". It works on the minimal examples, but I cannot speak to how it affects other `(expression_item)` nodes.
* The history is still a bit of a mess :confounded: Best to just review the diff from `main`.

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1086,13 +1086,25 @@
 )
 
 ; Make an indented block where a function type arrow starts. Only for the root
-; level, not for each arrow.
+; level of constructed types and PPX extensions, not for each arrow.
 ;
 ; (?used_slot:bool ref ->
 ;   Longident.t loc ->
 ;   Path.t * Env.t)
 ;
+; let h =
+;   [%madcast: float ->
+;     bool]
+;
 (constructed_type
+  (function_type
+    "->" @append_indent_start
+    (_) @append_indent_end
+    .
+  )
+)
+
+(attribute_payload
   (function_type
     "->" @append_indent_start
     (_) @append_indent_end
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1086,25 +1086,13 @@
 )
 
 ; Make an indented block where a function type arrow starts. Only for the root
-; level of constructed types and PPX extensions, not for each arrow.
+; level, not for each arrow.
 ;
 ; (?used_slot:bool ref ->
 ;   Longident.t loc ->
 ;   Path.t * Env.t)
 ;
-; let h =
-;   [%madcast: float ->
-;     bool]
-;
 (constructed_type
-  (function_type
-    "->" @append_indent_start
-    (_) @append_indent_end
-    .
-  )
-)
-
-(attribute_payload
   (function_type
     "->" @append_indent_start
     (_) @append_indent_end
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -1113,11 +1113,10 @@
 )
 
 ; Make an indented block where a function/match starts in PPX syntax.
-; NOTE This is probably a bit of a hack...
-(expression_item
-  .
-  (_) @prepend_indent_start
-) @append_indent_end
+(extension
+  "[%" @append_indent_start
+  "]" @prepend_indent_end @prepend_empty_softline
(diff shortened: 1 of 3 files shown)


languages/ocaml.scm
```diff
@@ -1100,6 +1104,21 @@
...
+  )
+)
+
+; Make an indented block where a function/match starts in PPX syntax.
+; NOTE This is probably a bit of a hack...
+(expression_item
+  .
+  (_) @prepend_indent_start
+) @append_indent_end
+
```
nbacquey, line 1121: I would say we can solve all three cases (functions, matches, constructed types) with something like
```scheme
(extension
  "[%" @append_indent_start
  "]" @prepend_indent_end
)
```
Does it work?

nbacquey: We could even add `@prepend_empty_softline` on the closing bracket for good measure.

Xophmeister: That does work :+1: ...and I think with the softline before the closing bracket, the output looks nicer (imo):

```ocaml
let x =
  [%expr function
    | false -> 0.
    | true -> 1.
  ]
```

There's some double indentation though:
```ocaml
let h =
  [%madcast: float ->
      bool
  ]
```

See 5ae369c37bfe

languages/ocaml.scm
```diff
@@ -1100,6 +1104,20 @@
   )
 )
 
+(attribute_payload
+  (function_type
+    "->" @append_indent_start
+    (_) @append_indent_end
+    .
+  )
+)
```
nbacquey, line 1113: Isn't this query causing the double indentation? I think it can be removed

languages/ocaml.scm
```diff
@@ -1086,12 +1086,16 @@
...
+; level of constructed types and PPX extensions, not for each arrow.
 ;
 ; (?used_slot:bool ref ->
 ;   Longident.t loc ->
 ;   Path.t * Env.t)
 ;
+; let h =
+;   [%madcast: float ->
+;     bool]
+;
```
nbacquey, line 1098: This can be removed as well

nbacquey (changes requested):

languages/ocaml.scm
```diff
@@ -1086,12 +1086,16 @@
 )
 
 ; Make an indented block where a function type arrow starts. Only for the root
-; level, not for each arrow.
+; level of constructed types and PPX extensions, not for each arrow.
```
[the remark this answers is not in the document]
Xophmeister: Done in 40e20efa6a1

languages/ocaml.scm
Xophmeister: Done in 40e20efa6a1

languages/ocaml.scm
Xophmeister: Done in 40e20efa6a1

nbacquey (approved):

PR #285, opened 2023-02-14
[OCaml] Fixup after #280

nbacquey:
Fixes double bracketing and update comment in the sample file.

### DIFF
--- a/tests/samples/expected/ocaml.ml
+++ b/tests/samples/expected/ocaml.ml
@@ -931,7 +931,7 @@ let x =
     | true -> 1.
   ]
 
-(* New line for structures in module definitions if they appear in the input *)
+(* New line for structures in module definitions *)
 module MFloat: SERIALISABLE with
   type t = float
 = struct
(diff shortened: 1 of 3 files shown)


Xophmeister (approved):
Sorry! I completely misunderstood what you meant by double-bracketing

tests/samples/expected/ocaml.ml
```diff
@@ -931,7 +931,7 @@ let x =
     | true -> 1.
   ]
 
-(* New line for structures in module definitions if they appear in the input *)
+(* New line for structures in module definitions *)
```
aspiwack, line 934: Does this diff imply that if I write a one-line module a newline will be inserted and it'll become multiline?

Xophmeister: It's still a softline, so a one-line module will remain on a single line.

nbacquey: The diff means that a newline will be inserted if the context is multi-line, even if the newline isn't _precisely there_ in the input (basically it's `scoped_softline` versus `input_softline`). Your one-line module would stay as such.

PR #298, opened 2023-02-15
Introduce `@single_line_no_indent` predicate

### LINKED ISSUE #287: Support line number directives
As of https://github.com/tweag/topiary/commit/1f90307eba5a316f7336903ecfcd817396534a2f, we offer no support for line number directives, which are statements of the form
```ocaml
# 1 "path/to/file.ml"
```
Those line number directives need to be alone in their line, and the `#` must be the first character of the line.
Thus, even if we add this query to the rules:
```scheme
(line_number_directive) @prepend_hardline @append_hardline
```
We would format
```ocaml
module Foo = struct
  val bar
# 1 "v1/pervasives.mli"
end
```
as
```ocaml
module Foo = struct
  val bar
  # 1 "v1/pervasives.mli"
end
```
which is invalid OCaml code: the `#` is indented.

I suggest we add a predicate to print a node on a single, unindented line.

nbacquey:
The matched node will be printed alone, on a single line, with no indentation.

#### Example

```scheme
(#language! ocaml)
; line number directives must be alone on their line, and can't be indented
(line_number_directive) @single_line_no_indent
```

With this PR, the following
```ocaml
module Foo = struct
  val bar

# 1 "v1/pervasives.mli"
end
```
will be formatted as
```ocaml
module Foo = struct
  val bar
# 1 "v1/pervasives.mli"
end
```
The loss of the line break is a bit of a pity, but given the low usage of the line number statement, I don't think it matters that much. We'll open a low-priority issue if we deem it necessary.

Closes #287

### DIFF
--- a/src/pretty.rs
+++ b/src/pretty.rs
@@ -29,7 +29,17 @@ fn atoms_to_doc<'a>(
                 Atom::Empty => RcDoc::text(""),
                 &Atom::Hardline => RcDoc::hardline()
                     .append(RcDoc::concat(repeat(RcDoc::space()).take(*indent_level))),
-                Atom::Leaf { content, .. } => RcDoc::text(content.trim_end()),
+                Atom::Leaf {
+                    content,
+                    single_line_no_indent,
+                    ..
+                } => {
+                    if *single_line_no_indent {
+                        RcDoc::hardline().append(RcDoc::text(content.trim_end()))
+                    } else {
+                        RcDoc::text(content.trim_end())
+                    }
+                }
                 Atom::Literal(s) => RcDoc::text(s),
                 Atom::MultilineOnlyLiteral { .. } => unreachable!(),
                 Atom::IndentEnd => unreachable!(),
--- a/src/pretty.rs
+++ b/src/pretty.rs
@@ -1,14 +1,20 @@
 use crate::{Atom, FormatterResult};
+use core::iter::repeat;
 use pretty::RcDoc;
 
-pub fn render(atoms: &[Atom], indent_level: isize) -> FormatterResult<String> {
-    let doc = atoms_to_doc(&mut 0, atoms, indent_level);
+pub fn render(atoms: &[Atom], indent_offset: usize) -> FormatterResult<String> {
+    let doc = atoms_to_doc(&mut 0, atoms, indent_offset, &mut 0);
     let mut rendered = String::new();
     doc.render_fmt(usize::max_value(), &mut rendered)?;
     Ok(rendered)
 }
 
-fn atoms_to_doc<'a>(i: &mut usize, atoms: &'a [Atom], indent_level: isize) -> RcDoc<'a, ()> {
+fn atoms_to_doc<'a>(
+    i: &mut usize,
+    atoms: &'a [Atom],
+    indent_offset: usize,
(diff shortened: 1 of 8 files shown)


src/pretty.rs
```diff
@@ -17,16 +23,32 @@ fn atoms_to_doc<'a>(i: &mut usize, atoms: &'a [Atom], indent_level: isize) -> Rc
...
+                }
                 Atom::Literal(s) => RcDoc::text(s),
                 Atom::MultilineOnlyLiteral { .. } => unreachable!(),
                 Atom::IndentEnd => unreachable!(),
                 Atom::IndentStart => {
                     *i += 1;
-                    atoms_to_doc(i, atoms, indent_level).nest(indent_level)
+                    *indent_level += indent_offset;
+                    let res = atoms_to_doc(i, atoms, indent_offset, indent_level);
+                    *indent_level -= indent_offset;
```
Xophmeister, line 50: `indent_level` is `usize`, so if the query rules aren't correct, this will panic when `indent_offset > indent_level`. Previously the indentation was signed, which I suppose was done to get around the panicking, despite resulting in the formatter returning clipped output.

TBH, I think panicking is better than weird output. However, maybe it's worth an issue to handle this error more gracefully.

Xophmeister: By virtue of L48, we know that `indent_level >= indent_offset`...so maybe this never happens. However, it is clearly possible to write a query with unbalanced indentation:

```scm
(#language! ocaml)
(comment) @append_hardline @append_indent_end
```

Formats:

```
(* Line 1 *)
(* Line 2 *)
```

...as:

```
(* Line 1 *)
```

So it doesn't panic :shrug: As such, ignore this comment, because it's not relevant to your changes! Nonetheless, I'll add an issue to capture this more generally.

Xophmeister (approved):
This looks good :+1: It seems like a very unusual edge case -- I don't immediately recall seeing similar syntactic requirements in other languages -- but it's great that it's handled.

I've added some comments, but this is more to aid my understanding than anything else!

src/pretty.rs
Xophmeister: (Added Issue #299)

src/pretty.rs
```diff
@@ -17,16 +23,32 @@ fn atoms_to_doc<'a>(i: &mut usize, atoms: &'a [Atom], indent_level: isize) -> Rc
...
-                Atom::Leaf { content, .. } => RcDoc::text(content.trim_end()),
+                &Atom::Hardline => RcDoc::hardline()
+                    .append(RcDoc::concat(repeat(RcDoc::space()).take(*indent_level))),
+                Atom::Leaf {
+                    content,
+                    single_line_no_indent,
+                    ..
+                } => {
+                    if *single_line_no_indent {
+                        RcDoc::hardline().append(RcDoc::text(content.trim_end()))
```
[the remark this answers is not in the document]
nbacquey: The following `\n` is added as an appended `HardLine` in `atom_collection.rs`. It allows it to be merged at post-processing with a possible following `Hardline`, like the one which comes from
```scheme
"end" @prepend_spaced_softline
```

torhovland:
Shouldn't `line_number_directive` be added to `allow_blank_line_before`?

PR #313, opened 2023-02-20
[perf] pre-compute `parent_leaf_node` values

### LINKED ISSUE #312: [Performance] improve performance of `parent_leaf_node()`
The attached flame graph corresponds to a Topiary run on a large (6k lines), real-world OCaml file. It shows that a good part of the time is spent in the `parent_leaf_node` function, which is very inefficient, as it does a lot of redundant work.

flamegraph

nbacquey:
This PR decreases the processing time of the larger OCaml files from ~12s to ~7s.
Closes #312


Xophmeister (approved):
:+1:

PR #314, opened 2023-02-21
[perf] Store query matches locally to avoid running query twice

nbacquey:
Decreases the processing time of larger OCaml files from ~7s to ~1.6s

### DIFF
--- a/src/atom_collection.rs
+++ b/src/atom_collection.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{HashMap, HashSet};
 
 use tree_sitter::Node;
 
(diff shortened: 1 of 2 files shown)


Xophmeister (approved):
Looks great :+1:

Not for this PR, but just thinking out aloud: Do any of the performance metrics show the time spent hashing? If it's a lot, maybe moving to a faster algorithm, rather than Rust's default with DoS-protection, would be worthwhile...

nbacquey:
Right now, nothing seems to indicate we spend much time hashing. The most recent flame graph indicates that most of the time is spent applying the query, in the C bindings themselves
flamegraph

aspiwack:
> from ~7s to ~1.6s

:exploding_head:

torhovland:
And the best part: by not running something twice 😀

aspiwack:
Dividing by 3 by doing 1/2 as much work. Pretty :sunglasses: .

PR #315, opened 2023-02-21
Only extract root-level predicates from query files to act as language configuration pragmata

### LINKED ISSUE #311: The query configuration parser is too lenient
**Describe the bug**
The query configuration parser, which extracts the language and (optionally) indentation offset, uses a regular expression for `(#PREDICATE! VALUE)` tokens in the query language. It matches them anchored to the beginning of a line.

The same syntax can also be used in query rules...

**To Reproduce**
...so write a query rule with a predicate that's at the beginning of a line. For example:

```scm
(some_node
(_) @append_delimiter
(#!delimiter ";")
)
```

It may not be pretty, but this is a valid formatting query, which the configuration parser will think contains a configuration directive and fail with:

```console
[2023-02-17T14:53:55Z ERROR topiary] Unknown configuration predicate 'delimiter'
```

**Expected behavior**
Only root level (i.e., not associated with any query) directives should be parsed as configuration directives.

**Environment**
 - OS name + version: NixOS 22.05
 - Version of the code: 04d7fd8b7368976

**Additional context**
n/a

Xophmeister:
This PR replaces the regular expression that extracted configuration pragmata from query files with a correct parser, which can discern between root-level predicates (which we want) and badly formatted rule-level predicates (which we don't).

### Benchmarking
I took the opportunity to benchmark before-and-after:

```bash
for lang in "ocaml.ml" "rust.rs" "bash.sh"; do
  for _ in $(seq 50); do
    command time --format "${lang%%.*}\t%e\t%M" \
      ./target/release/topiary -sq languages/${lang%%.*}.scm \
                               -f tests/samples/input/$lang \
    2>&1 \
    >/dev/null
  done
done \
| datamash --sort groupby 1 mean 2 sstdev 2 mean 3 sstdev 3
```

**tl;dr** They're about the same.

#### With `regex` (32a33fb5)

Release build: 9,904,032 bytes

Formatting metrics:

| Language | User time (seconds)         | Max. Residency (KB)        |
| :------- | :-------------------------- | :------------------------- |
| bash     | 0.0104 ± 0.0019794866372216 |  7753.84 ± 75.186668379712 |
| ocaml    | 0.1984 ± 0.0042185208743524 | 14429.6  ± 98.209272599331 |
| rust     | 0.0006 ± 0.002398979374821  |  7388.8  ± 76.950023537677 |

#### With `tree-sitter-query` (3ecec57)

Release build: 9,891,064 bytes

Formatting metrics:

| Language | User time (seconds)         | Max. Residency (KB)        |
| :------- | :-------------------------- | :------------------------- |
| bash     | 0.0106 ± 0.002398979374821  |  7361.6  ± 101.36054056522 |
| ocaml    | 0.1996 ± 0.006375975315549  | 14158.64 ± 76.202137202875 |
| rust     | 0.0004 ± 0.0019794866372216 |  7042.64 ± 85.024308648906 |

### Notes
* The first few commits of this branch represent an abandoned attempt to write my own parser, using Nom. It quickly became clear that using Tree-sitter itself was a better approach.
* I was not aware of [`tree_sitter::Query::general_predicates`](https://docs.rs/tree-sitter/0.20.9/tree_sitter/struct.Query.html#method.general_predicates) until I'd finished my implementation. Never mind.
* By depending on the Tree-sitter grammar for Tree-sitter queries, formatting rules can be written (in future PRs) to format our Tree-sitter queries.
* Resolves #311

image


nbacquey:
I think you benchmark could be improved, in the sense that right now, it doesn't *use* the configuration values found in the file.
To force topiary to use this value, you should not use the `-f` option, and call topiary with something like
```bash
cat tests/samples/input/$lang | topiary -sq languages/${lang%%.*}.scm
```

src/configuration.rs
```diff
@@ -1,80 +1,281 @@
...
+                            "The #indent-level! pragma must have a parameter".into(),
                             None,
                         ));
                     }
                 }
+
                 _ => {
                     return Err(FormatterError::Query(
-                        format!("Unknown configuration predicate '{predicate}'"),
+                        format!("Unknown pragma '#{predicate}!'"),
```
nbacquey, line 207: Maybe we could be a little more informative here? Like specifying that the error happened during configuration parsing, not during the parsing of the query for its application to the input? Also, maybe we could provide a list of known pragma?
That may be out of the scope of this PR, though.

nbacquey (commented):
Other than my single comment (which your are free to disregard), the PR looks good!
Although, I don't feel confident enough in Rust to comment on the mechanism you used to iterate over the pragmas, so I'll defer approval for now.

src/configuration.rs
Xophmeister: I've made a _slight_ change to the wording in d24a0e9. I didn't provide a list of known pragmata, largely because I think it ought to be done more generally -- rather than simply using a static string -- and that deserves its own PR. (I'll add an issue for this.)

Xophmeister: (Submitted Issue #316)

Xophmeister:
> I think your benchmark could be improved, in the sense that right now, it doesn't _use_ the configuration values found in the file. To force topiary to use this value, you should not use the `-f` option, and call topiary with something like
> 
> ```shell
> cat tests/samples/input/$lang | topiary -sq languages/${lang%%.*}.scm
> ```

This is true. The configuration is parsed regardless, but it's not used; that shouldn't make a difference to the benchmarking...but it's easy to rerun :slightly_smiling_face: 

This PR (d24a0e9[^1]):
[^1]: After rebasing on 10faa3e

| Language | User Time (seconds)         | Max. Residency (KB)        |
| :------- | :-------------------------- | :------------------------- |
| bash     | 0.0108 ± 0.0027404751561787 | 7349.76 ±  89.048154823955 |
| ocaml    | 0.21   ± 0.0088063057185271 | 16330.4  ± 115.78163824675 |
| rust     | 0.0016 ± 0.0037032803990902 | 7011.44 ±  90.740302231035 |

With `regex` (10faa3e):

| Language | User Time (seconds)         | Max. Residency (KB)        |
| :------- | :-------------------------- | :------------------------- |
| bash     | 0.01   ± 3.422529947954e-21 |  7655.76 ± 82.576509977112 |
| ocaml    | 0.1872 ± 0.0060743690358484 | 16740.72 ± 90.284593121429 |
| rust     | 0      ± 0                  |  7293.6  ± 72.64113637518  |

> Other than my single comment (which your are free to disregard), the PR looks good! Although, I don't feel confident enough in Rust to comment on the mechanism you used to iterate over the pragmas, so I'll defer approval for now.

Thanks, @nbacquey :pray:

Pinging @torhovland and/or @ErinvanderVeen, if you have time.

src/configuration.rs
```diff
@@ -1,80 +1,281 @@
+use std::str::FromStr;
```
ErinvanderVeen, line 1: As we are targeting a library use-case at some point, it would make sense to start writing documentation in the module headers. For now, this is good though!

ErinvanderVeen (approved):

src/configuration.rs
Xophmeister: Good point :ok_hand: I'll add this as an issue so it won't get forgotten about.

PR #317, opened 2023-02-22
Support tree-sitter query language

### LINKED ISSUE #101: Add a formatter for tree-sitter's query language
Our formatters' logic, are written in tree-sitter's query language, a [scheme-like language](https://github.com/tweag/topiary/pull/100#issuecomment-1305173729). These queries are currently not formatted by a code formatter.

In the spirit of dog-fooding (and because I'm not actually aware of an existing formatter), I suggest that we make a formatter for the query language, and format our queries with it.

nbacquey:
This PR adds support for the tree-sitter query language:
* Add a query file and the boilerplate to support `.scm` files as input
* Format our query files with Topiary
* Add a test to check that our query files are properly formatted

Closes #101
Based upon #315 for the tree-sitter query grammar

### DIFF
--- a/languages/bash.scm
+++ b/languages/bash.scm
@@ -276,35 +276,43 @@
 )
 
 (if_statement
-  . _ "then"
+  .
+  _
+  "then"
   [(command) (list) (pipeline) (compound_statement) (subshell) (redirected_statement) (variable_assignment)] @append_hardline
   .
   "\n"
 )
 
 (elif_clause
-  . _ "then"
+  .
+  _
+  "then"
   [(command) (list) (pipeline) (compound_statement) (subshell) (redirected_statement) (variable_assignment)] @append_hardline
   .
   "\n"
 )
 
 (else_clause
-  . "else"
+  .
+  "else"
   [(command) (list) (pipeline) (compound_statement) (subshell) (redirected_statement) (variable_assignment)] @append_hardline
   .
   "\n"
 )
 
 (case_item
-  . _ ")"
+  .
+  _
+  ")"
   [(command) (list) (pipeline) (compound_statement) (subshell) (redirected_statement) (variable_assignment)] @append_hardline
   .
   "\n"
 )
 
 (do_group
-  . "do"
+  .
+  "do"
   [(command) (list) (pipeline) (compound_statement) (subshell) (redirected_statement) (variable_assignment)] @append_hardline
   .
   "\n"
@@ -324,7 +332,7 @@
 )
 
 (pipeline
-  [(_) "|" "|&" ] @append_space
+  [(_) "|" "|&"] @append_space
   .
   _
 )
@@ -418,7 +426,8 @@
 [
   (if_statement)
   (elif_clause)
-] "then" @append_hardline @append_indent_start
+]
+"then" @append_hardline @append_indent_start
 
 ; New line after "else" and start indent block
 (else_clause
@@ -442,7 +451,6 @@
   ";"* @do_nothing
   .
   "then" @prepend_delimiter @prepend_space
-
   (#delimiter! ";")
 )
 
@@ -459,8 +467,8 @@
 ; in the context of a (test_command); it does work in other contexts
 ; See https://github.com/tweag/topiary/pull/155#issuecomment-1364143677
 (binary_expression
-   left: _ @append_space
-   right: _ @prepend_space
+  left: _ @append_space
+  right: _ @prepend_space
 )
 
 ;; Case Statements
@@ -468,7 +476,11 @@
 ; Indentation block between the "in" and the "esac"
 (case_statement
   .
-  "case" . _ .  "in" @append_hardline @append_indent_start
+  "case"
+  .
+  _
+  .
+  "in" @append_hardline @append_indent_start
   _
   "esac" @prepend_hardline @prepend_indent_end
   .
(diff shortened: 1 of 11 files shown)


nbacquey:
Will stay as Draft until #315 is merged

languages/tree-sitter-query.scm
```diff
@@ -0,0 +1,94 @@
...
+; Sometimes we want to indicate that certain parts of our source text should
+; not be formatted, but taken as is. We use the leaf capture name to inform the
+; tool of this.
+[
+  (capture)
+  (identifier)
+  (string)
+] @leaf
+
+(comment) @prepend_hardline @append_hardline
```
Xophmeister, line 13: What about comments like this:

```scm
(foo
  (_) ; last named node
  .
)
```

I think this will probably work:

```suggestion
(comment) @prepend_input_softline @append_hardline
```

src/language.rs
```diff
@@ -46,6 +47,7 @@ const EXTENSIONS: &[(Language, &[&str])] = &[
     (Language::OcamlInterface, &["mli"]),
     (Language::Rust, &["rs"]),
     (Language::Toml, &["toml"]),
+    (Language::TreeSitterQuery, &["scm"]),
```
Xophmeister, line 50: Maybe we should move away from the `.scm` extension. Tree-sitter query files are Scheme-like, but not Scheme proper, and by squatting on the Scheme extension we limit ourselves if/when we want to format actual Scheme.

My personal preference would be `.topiary` for our query files. While they are generic Tree-sitter queries -- so, perhaps `.tsq` would suffice -- they contain Topiary-specific predicates. Perhaps, therefore:

```suggestion
    (Language::TreeSitterQuery, &["topiary", "tsq"]),
```

...with the appropriate changes to the `README` and elsewhere, plus renaming of everything under the `languages` directory.

Xophmeister (commented):
Looks good :+1:

I've made a few suggestions, but no blockers. One of which we should probably discuss amongst the team...

Xophmeister:
If you rebase against `chris/config-fix` and change the base branch in the PR to `chris/config-fix`, that ought to clear up the diff output to only include your changes. When #315 is merged, the base branch should change automatically to `main` (although you'll probably have to do _another_ rebase to deal with the squash).

src/language.rs
nbacquey: I agree on your point, I made it myself while writing this.
`.topiary` and `.tsq` both look right to me

Xophmeister: I also raised the point on Slack, to see what the rest of the team think.

languages/tree-sitter-query.scm
nbacquey: I would rather avoid the `input_softline` predicates, since I think Topiary should be strongly opinionated. Although you're right, the comment you give as an example won't be processed correctly.
I guess until we rework the comment processing altogether, `@prepend_input_softline` will do.

nbacquey:
> If you rebase against `chris/config-fix` and change the base branch in the PR to `chris/config-fix`, that ought to clear up the diff output to only include your changes. When #315 is merged, the base branch should change automatically to `main` (although you'll probably have to do _another_ rebase to deal with the squash).

Indeed, it looks better now

src/language.rs
ErinvanderVeen: My single issue with switching away from `.scm` is that we would go against what is common. Even the "official" tree-sitter documentation explicitly uses `.scm`.

> and by squatting on the Scheme extension we limit ourselves if/when we want to format actual Scheme.

Yeah, but this is not a solution. Topiary should still be able to format the tree-sitter query files from other projects, e.g. Neovim, helix, emacs, and thus it will always encounter `.scm` files that are actually query files. Renaming our query files to `.tsq` will ensure that we know when we are formatting our own query files, but not from these other projects.

Xophmeister: I wasn't aware how entrenched `.scm` was in the Tree-sitter ecosystem. That does change things...

(Also, devil's advocate: Scheme is traditionally formatted with both indents and alignments on s-expressions...which Topiary doesn't [[currently](https://github.com/tweag/topiary/issues/170)] support.)

nbacquey: I'm currently leaning towards leaving it as it is right now. If we ever decide to format proper Scheme files, the refactoring will be minimal anyway

Xophmeister (approved):

Xophmeister (approved):

languages/bash.scm
```diff
@@ -459,16 +467,20 @@
...
 )
 
 ;; Case Statements
 
 ; Indentation block between the "in" and the "esac"
 (case_statement
   .
-  "case" . _ .  "in" @append_hardline @append_indent_start
+  "case"
+  .
```
aspiwack, line 480: Question: is it typical for a `.` to be on its own line? Or should we rather treat it like a delimiter?

tests/samples/expected/tree-sitter-query.scm
```diff
@@ -0,0 +1,1508 @@
+; Configuration
+(#language! ocaml)
```
aspiwack, line 2: In the future, having such a large example file may prove unwieldy. We'll see, but if it gets difficult to review style changes, we should probably consider breaking it up in smaller chunks.

languages/bash.scm
Xophmeister: I also personally prefer it to act as a delimiter. However, in Tree-sitter's example documentation, I think it's always written on its own line. Probably for the case of when you have a query like:

```scheme
(
  ((some) (really) (complicated) (thing))
  .
  [(another) (really) (complicated) (thing)]
)
```

Issue #295, opened 2023-02-14
Migrate to a Cargo workspace, to keep library and binary distinct


Xophmeister:
**Is your feature request related to a problem? Please describe.**
Many of the dependencies and functionality are relevant only to the Topiary library or binary, but rarely both (besides transitive dependency). It will make the code more maintainable, as additional functionality and adjacent tools are developed, to demarcate a clear separation.

**Describe the solution you'd like**
In #294, I have restructured the binary's source files, such that they have their own directory. This allows for binary-only modules to sit alongside it, rather than polluting the library namespace. This is a halfway-house to using Cargo workspaces, which will additionally allow the separation of dependencies (e.g., the library doesn't care about `tempfile`, etc.) and the isolation of any future tools (e.g., [debuggers](https://github.com/tweag/topiary/issues/276#issuecomment-1429540897), visualisers, [playground](https://github.com/tweag/topiary/issues/171), etc.).

**Describe alternatives you've considered**
n/a

**Additional context**
The Cargo workspace _should_, in principle, be straightforward to configure. It wasn't added as part of #294 because it requires additional work to be compatible with the Nix setup.

A workspace may also be prerequisites (or, at least, highly desirable) for #137 and #171 

---

**Tracking**
* [ ] #303
* [ ] #304
* [ ] #305

Issue #11, opened 2022-08-30
Command line option for outputting the tree-sitter syntax tree


torhovland:
Helpful when developing queries.

---

Implementation plan:
* [x] Refactor/structure source, so binary-related functions can be kept distinct from the library.
* [x] ~Implement/expose Tree-sitter syntax tree output in the library.~ Shore up `Language` machinery for Tree-sitter's sake.
* [x] Implement JSON serialisation of syntax output in the binary.
* [x] Expose feature through CLI flag (+ document)
* [ ] Stretch: Implement DOT serialisation of syntax output in the binary

Xophmeister:
I'm going to take this, now the v0.1 issues are done, to improve my Rust :ok_hand:

PR #318, opened 2023-02-23
Refactor Language interface

Xophmeister:
The original purpose of this PR was to expose extension points to Tree-sitter's parser, at the library level, such that the binary could emit a serialised form of the parsed tree (i.e., to satisfy #11). After much thought, however, it is my opinion that it is _not_ the library's job to provide this and is only offered by the binary as a "nice-to-have". Specifically:

* Walking the syntax tree in Tree-sitter is relatively straightforward. Serialising the output is a harder task, but has nothing to do with formatting.
* Exposing Tree-sitter is no different from using Tree-sitter directly.
* As such, if the exposure was to provide an interface -- presuming, for the sake of argument, it's the library's responsibility -- then we'd have the job of reimplementing or wrapping Tree-sitter's functionality. Due to orphan rules, these _may_ then need to be wrapped again in the binary.
* That's a lot of wasted effort, when we could just have the binary also depend on Tree-sitter. (Currently, this requires zero effort, as the library and binary are still conjoined; see #295.)
* What _is_ needed, however, is Topiary's library's machinery to deal with various languages (including #4, once implemented), such that it can be passed to Tree-sitter without additional fuss.

This PR is therefore just a refactoring of the `Language` type: Rather than explicit methods, we use conversion traits instead, plus there's more documentation. It's not strictly necessary, but ought to make the subsequent code for the binary slightly more idiomatic.

* [x] Based on #317

### DIFF
--- a/src/bin/topiary/main.rs
+++ b/src/bin/topiary/main.rs
@@ -75,7 +75,7 @@ fn run() -> CLIResult<()> {
     let language: Option<Language> = if let Some(language) = args.language {
         Some(language.into())
     } else if let Some(filename) = args.input_file.as_deref() {
-        Some(PathBuf::from(filename).try_into()?)
+        Some(Language::detect(filename)?)
     } else {
         // At this point, Clap ensures that args.query must be present.
         // We will read the language from the query file later.
@@ -86,7 +86,7 @@ fn run() -> CLIResult<()> {
         query
     } else if let Some(language) = language {
         // Deduce the query file from the language, if the argument is missing
-        language.try_into()?
+        language.query_file()?
     } else {
         // Clap ensures we won't get here
         unreachable!();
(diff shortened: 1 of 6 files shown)


Xophmeister:
(Rebasing was not very clean; now fixed.)

torhovland (approved):

Xophmeister:
@torhovland My -- probably naive -- reasoning for this is that the `TryFrom` implementations "feel more useful" (don't ask me to quantify or justify this!) against the concrete type, rather than its reference, for the sake of an API. However, I'm not very certain about that, given the consequences of ownership.

Do you think, for the API, implementations against the references would be more idiomatic?

torhovland:
Well, the `From` and `TryFrom` implementations are typically added to serve a need. So when you do `self.into()` and the compiler tells you that

```
the trait `From<&language::Language>` is not implemented for `Vec<tree_sitter::Language>`
```

I would take that as a hint that I need that implementation.

Xophmeister:
Good point. I suppose what I'm trying to do is anticipate the need of future downstream API users, but perhaps that is futile, or at the very least, secondary compared to current needs.

PR #319, opened 2023-02-23
Enforce opinion about optional `|` before first match case

### LINKED ISSUE #281: Enforce an opinion about optional `|` before first match case
I think Topiary should be opinionated on whether we should put the optional `|` before the first match case in a pattern matching expression.

I would argue that the following two snippets are correctly formatted:
```ocaml
let not = function true -> false | false -> true
```
```ocaml
let not = function
  | true -> false
  | false -> true
```
While the following two are not:
```ocaml
let not = function | true -> false | false -> true
```
```ocaml
let not = function
  true -> false
  | false -> true
```

In order to only allow the first two snippets, we need two capabilities:
* produce a symbol only in a multi-line context (which `@append_multiline_delimiter` does),
* delete a symbol only if the context is single-line (which we cannot currently do).

nbacquey:
This PR introduces a new formatting predicate, `singleline_delete`, which deletes a node only if the context is single-line. It also uses this new predicate, along with `prepend_multiline_delimiter`, to enforce the following rule in OCaml:
**The optional `|` before the first match case must be present if and only if the context is multi-line.**

Formats
```ocaml
let not = function
  true -> false
  | false -> true

let not = function | true -> false | false -> true
```
as
```ocaml
let not = function
  | true -> false
  | false -> true

let not = function true -> false | false -> true
```

TODO:
- [X] Edit OCaml queries to delete first `|` in match cases
- [X] Add code to test the feature in OCaml samples

Closes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,24 @@ Remove the matched node from the output.
 )
 ```
 
+### `@singleline_delete`
+
+Remove the matched node from the output, online if the context is single-line.
+
+#### Example
+
+```scheme
+; Delete the optional "|" before the first match case
+; if the context is single-line
+(
+  (match_case)? @do_nothing
+  .
+  "|" @singleline_delete
+  .
+  (match_case)
+)
+```
+
 ### `@do_nothing`
 
 If any of the captures in a query match are `@do_nothing`, then the
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -685,6 +685,25 @@
   (match_case) @prepend_spaced_softline
 )
 
+; Allow (and enforce) the optional "|" before the first match case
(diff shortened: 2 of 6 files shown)


nbacquey:
Regardless of the failing CI, which I think is not related to the modifications on the codebase, the PR is ready for reviews

Xophmeister:
(The failing CI is fixed in #324. Let me cherry pick that commit so it can be done sooner...)

Xophmeister (changes requested):
This looks super-useful :+1:

Could you please add some documentation for this in the `README`? Otherwise, it looks good to me.

Xophmeister (approved):

PR #323, opened 2023-02-28
Clippy errors didn't fail CI build.

torhovland:


tests/cli-tester.rs
```diff
@@ -61,9 +61,9 @@ fn test_no_clobber() {
         .arg("--language")
         .arg("json")
         .arg("--input-file")
-        .arg(&input_path)
+        .arg(input_path)
```
Xophmeister, line 64: Clippy isn't warning me about this, for some reason. The reference is because `input_path` used to be a `Path`, rather than `&Path`, FWIW... Is there some configuration to make Clippy more vocal about these kinds of things?

src/tree_sitter.rs
```diff
@@ -97,7 +97,7 @@ pub fn apply_query(
     Ok(atoms)
 }
 
-fn capture_name<'a, 'b>(query: &'a Query, capture: &'b QueryCapture) -> &'a str {
+fn capture_name<'a>(query: &'a Query, capture: &QueryCapture) -> &'a str {
```
Xophmeister, line 100: Is this equivalent to:

```rust
fn capture_name<'a>(query: &'a Query, capture: &'_ QueryCapture) -> &'a str
```

i.e., The lifetime of `capture` doesn't matter because the compiler can prove it's within `'a`?

src/language.rs
```diff
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use crate::{FormatterError, FormatterResult, IoError};
 
 /// The languages that we support with query files.
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```
Xophmeister, line 7: Is this just for the sake of readers: `Eq` denotes an equivalence relation, but it's not something that the compiler proves to be true?

Xophmeister (approved):
The Rust code changes look good, but I'm curious how you got Clippy to catch the instances that you've changed. Looking closely into your changes, it's clear why you've made them, but my Clippy isn't complaining and I wouldn't have spotted them otherwise.

I've approved this, but note that I don't follow the Nix/Flake code. To my untrained eye, they look innocuous enough :+1:

tests/cli-tester.rs
torhovland: My guess is you're on an older version:

```
cargo --version
cargo 1.67.1 (8ecd4f20a 2023-01-10)
```

src/tree_sitter.rs
torhovland: Not necessarily within `'a`, but yes, it doesn't matter. See https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html#lifetime-elision

src/language.rs
torhovland: See https://rust-lang.github.io/rust-clippy/master/#derive_partial_eq_without_eq

PR #324, opened 2023-03-03
Visualise syntax tree from CLI

### LINKED ISSUE #11: Command line option for outputting the tree-sitter syntax tree

Xophmeister:
This PR provides JSON visualisation of the Tree-sitter syntax tree from the Topiary CLI, via the `--visualise` command. It currently only supports JSON output, but it's designed to be extensible (I'd particularly like to have Graphviz output at some point).

The JSON output has the following fields for each node:

* `kind`: The node type, or raw token.
* `is_named`: Whether the node is anonymous or not.
* `is_extra`: Whether the node is "extra" (i.e., can appear anywhere, like comments).
* `is_error`: Whether the node represents an error.
* `is_missing`: Whether the node was inserted by Tree-sitter to fix an error.
* `start` and `end`: The node's row and column range (1-based, rather than Tree-sitter's 0-based).
* `children`: A list of child nodes.

Resolves #11 (and also fixes the CI by upgrading the failing `cachix/install-nix-action` action)


torhovland (commented):
This mostly looks good to me. But I think there is some code here that would fit better in the library. All direct dependencies to Tree-sitter could go there, I think. There may even be some duplication here that could be factored out.

Xophmeister:
@torhovland
> But I think there is some code here that would fit better in the library. All direct dependencies to Tree-sitter could go there, I think.

How can that be done without breaking the orphan rule? (Besides moving all the serialisation dependency into the library, but that seems redundant.)

torhovland:
Well, frankly I think all of the visualisation code belongs in the library. So you'd either get:

```
use topiary::{formatter, visualiser, Language};
```

or even just

```
use topiary::{formatter, Language};
```

where visualisation becomes an option to `formatter` the same way `skip_idempotence` is.

Xophmeister:
> Well, frankly I think all of the visualisation code belongs in the library.

OK, that makes it easy then :ok_hand:

src/bin/topiary/error.rs
```diff
@@ -48,11 +49,13 @@ impl From<FormatterError> for TopiaryError {
 impl From<io::Error> for TopiaryError {
     fn from(e: io::Error) -> Self {
         match e.kind() {
-            io::ErrorKind::NotFound => Self::Bin("File not found".into(), CLIError::IOError(e)),
+            io::ErrorKind::NotFound => {
+                Self::Bin("File not found".into(), Some(CLIError::IOError(e)))
+            }
 
             _ => Self::Bin(
                 "Cound not read or write to file".into(),
```
torhovland, line 57: Oh, would you mind fixing this typo as well?

src/lib.rs
```diff
@@ -137,23 +148,36 @@ pub fn formatter(
...
+            if !skip_idempotence {
+                idempotence_check(&trimmed, &query, language)?
+            }
+
+            write!(output, "{trimmed}")?;
+        }
 
-    write!(output, "{trimmed}")?;
+        Operation::Visualise { output_format } => {
+            let (tree, _) = tree_sitter::parse(&content, &configuration.language.grammars())?;
```
torhovland, line 173: `parse` should now just take the language (not the grammars) as input, so the API is similar to `apply_query` above.

torhovland (approved):

src/lib.rs
Xophmeister: `apply_query` calls `parse` internally, which takes the grammars: 

https://github.com/tweag/topiary/blob/5e83b4612629de19ab7304dee9d15a584812a300/src/tree_sitter.rs#L81-L82

This was done for the `.ml`/`.mli` split. I'll refactor `parse` so it just takes the language, cycling through the grammars internally...

PR #326, opened 2023-03-07
GraphViz visualisation support

Xophmeister:
This PR introduces basic GraphViz visualisation support, with the node shape depending on its type: elliptical for named nodes; rectangular for anonymous.

For example:

```
cargo run -- -vdot -ljson <<< '{"test":123}' \
| dot -Tsvg
```

Renders:

test

There's scope for improvement, but these are perhaps best addressed in separate PRs:
* Use a GraphViz-specific abstraction, rather than `fmt::Display`.
* Use, e.g., colour to indicate a node's status (e.g., red for error nodes, etc.).
* Embed other metadata (e.g., position) into the node.

### DIFF
--- a/src/bin/topiary/visualise.rs
+++ b/src/bin/topiary/visualise.rs
@@ -5,12 +5,14 @@ use clap::ValueEnum;
 
 #[derive(ValueEnum, Clone, Copy, Debug)]
 pub enum Visualisation {
+    Dot,
     Json,
 }
 
 impl From<Visualisation> for topiary::Visualisation {
     fn from(visualisation: Visualisation) -> Self {
         match visualisation {
+            Visualisation::Dot => Self::GraphViz,
             Visualisation::Json => Self::Json,
         }
     }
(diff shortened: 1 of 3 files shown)


src/bin/topiary/visualise.rs
```diff
@@ -5,12 +5,19 @@ use clap::ValueEnum;
...
+
+    // All other output formats should be listed
+    // in alphabetical order
+    Dot,
 }
 
 impl From<Visualisation> for topiary::Visualisation {
     fn from(visualisation: Visualisation) -> Self {
         match visualisation {
+            Visualisation::Dot => Self::GraphViz,
```
torhovland, line 20: Why two different names?

torhovland (approved):

torhovland:
@Xophmeister Better add some documentation too, I suppose.

src/bin/topiary/visualise.rs
Xophmeister: My rationale is that Clap converts the enum member into `lower-kebab-case` and, for the sake of the CLI argument, `dot` is preferable to `graph-viz` (or even `graphviz` or `gv`). However, internally, `Visualisation::GraphViz` is more meaningful than `Visualisation::Dot`.

PR #333, opened 2023-03-10
Use tree-sitter-facade to enable Wasm support

torhovland:
First step on the way to implementing #171.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,21 +26,22 @@
         {
             "type": "lldb",
             "request": "launch",
-            "name": "Debug executable 'tree-sitter-formatter'",
+            "name": "Debug executable 'topiary'",
             "cargo": {
                 "args": [
                     "build",
-                    "--bin=tree-sitter-formatter",
-                    "--package=tree-sitter-formatter"
+                    "--bin=topiary",
+                    "--package=topiary"
                 ],
                 "filter": {
-                    "name": "tree-sitter-formatter",
+                    "name": "topiary",
                     "kind": "bin"
                 }
             },
             "args": [
-                "<", "tests/samples/input/rust.rs",
-                "-l", "rust"
+                "--query", "languages/debug.scm",
+                "--input-file", "languages/debug.scm",
+                "-s"
             ],
             "cwd": "${workspaceFolder}"
         },
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,16 +17,16 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tempfile = "3.4.0"
 test-log = "0.2"
-tree-sitter = "0.20"
 tree-sitter-json = "0.19"
 tree-sitter-rust = "0.20.3"
 tree-sitter-toml = "0.20.0"
-
-# Needs a version > 0.19
-tree-sitter-ocaml = { git = "https://github.com/tree-sitter/tree-sitter-ocaml" }
+tree-sitter-facade = "0.9.0"
 
 tree-sitter-bash = { git = "https://github.com/tree-sitter/tree-sitter-bash" }
 tree-sitter-nickel = { git = "https://github.com/nickel-lang/tree-sitter-nickel" }
+
+# Needs a version > 0.19
+tree-sitter-ocaml = { git = "https://github.com/tree-sitter/tree-sitter-ocaml" }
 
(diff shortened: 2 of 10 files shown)


src/configuration.rs
```diff
@@ -1,281 +1,80 @@
-use std::str::FromStr;
```
Xophmeister, line 1: Reopens #311

src/error.rs
```diff
@@ -101,7 +101,7 @@ impl Error for FormatterError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match self {
             Self::Idempotence => None,
-            Self::Internal(_, source) => source.as_ref().map(|e| e as &dyn Error),
+            Self::Internal(_, source) => source.as_ref().map(|e| &**e),
```
Xophmeister, line 104: Isn't `&**e` equivalent to `*e`? Is this related to [this comment](https://tweag.slack.com/archives/CKMLS3094/p1678342865049809)?

src/language.rs
```diff
@@ -193,5 +194,36 @@ impl From<&Language> for Vec<tree_sitter::Language> {
...
+            Language::Nickel => vec!["nickel"],
+            Language::Ocaml => vec!["ocaml", "ocaml-interface"],
+            Language::OcamlImplementation => vec!["ocaml"],
+            Language::OcamlInterface => vec!["ocaml-interface"],
+            Language::Rust => vec!["rust"],
+            Language::Toml => vec!["toml"],
+            Language::TreeSitterQuery => vec!["query"],
+        }
+        .iter()
+        .map(|name| format!("tree-sitter-{}.wasm", name))
```
Xophmeister, line 218: These are presumably the grammar dependencies, compiled to WASM and distributed in the package (i.e., we can rely on this standard filename format)?

Also, what about `.mli`? IIRC, `Language::OCamlInterface` still uses `tree-sitter-ocaml`, but it exposes a separate `tree_sitter::Language`.

**tl;dr** If this works, then it may be worth documenting in comments why it works.

Xophmeister (approved):
I've left a few comments -- they're not really suggestions -- but LGTM :100:

.vscode/launch.json
```diff
@@ -26,21 +26,22 @@
...
-                    "name": "tree-sitter-formatter",
+                    "name": "topiary",
                     "kind": "bin"
                 }
             },
             "args": [
-                "<", "tests/samples/input/rust.rs",
-                "-l", "rust"
+                "--query", "languages/debug.scm",
+                "--input-file", "languages/debug.scm",
```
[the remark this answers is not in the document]
torhovland: In my opinion, in tool specific config files like this, there are things that are user-specific settings, and other things that are useful setup config for everyone (or at least every user of the tool). So I'd say some of this makes sense to have checked in.

Cargo.toml
```diff
@@ -11,25 +11,26 @@ clap = { version = "4.1", features = ["derive"] }
...
 log = "0.4"
-ouroboros = "0.15"
 pretty_assertions = "1.3"
+regex = "1.7.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tempfile = "3.4.0"
 test-log = "0.2"
-tree-sitter = "0.20"
+tree-sitter-facade = { git = "https://github.com/tweag/tree-sitter-facade" }
```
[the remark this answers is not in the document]
torhovland: I don't think so. They are potentially useful outside of Topiary, for anyone who needs to target Tree-sitter both natively and through Wasm. We are clearly not the only ones who have run into that need.

And since https://github.com/silvanshade/tree-sitter-facade is archived, we are the de-facto maintainers of that code now.

src/configuration.rs
torhovland: Good point. I've done it.

src/error.rs
torhovland: `e` is an `&Box<dyn Error>`. `*e` results in `expected reference, found struct Box`. We need to deref that `Box` with an extra `*`. Now we have a `dyn Error`. But since the function wants to return an `&dyn Error` (wrapped in an `Option`), we add an `&`.

.vscode/launch.json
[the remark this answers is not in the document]
Xophmeister: Fair enough. My opinion on extra-repository stuff is much more strict, but it is still an opinion :) Given that, and the easiest thing to do here is nothing, then :+1:

PR #336, opened 2023-03-13
Compile Wasm

torhovland:
Related to #171 .

`web_tree_sitter::Language::load_path()` is async, so now Topiary is as well. It's not entirely straightforward to make `formatter()` async, and it may be nice to keep the async stuff out of it anyway, so the API has changed so that grammars are loaded outside of it, then passed in to `formatter()`.

Also added Wasm building to Nix, and running that from CI.

### DIFF
--- a/src/bin/topiary/main.rs
+++ b/src/bin/topiary/main.rs
@@ -6,7 +6,7 @@ mod visualise;
 use std::{
     error::Error,
     fs::File,
-    io::{stdin, BufReader, BufWriter},
+    io::{stdin, BufReader, BufWriter, Read},
     path::PathBuf,
 };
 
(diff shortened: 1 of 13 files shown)


src/bin/topiary/main.rs
```diff
@@ -112,7 +113,21 @@ fn run() -> CLIResult<()> {
         unreachable!();
     };
 
-    let mut query = BufReader::new(File::open(query_path)?);
+    let mut query_reader = BufReader::new(File::open(query_path)?);
+
+    // It's not very nice that formatter wants an io::Read and
+    // Configuration::parse wants a &str. Should we just let formatter take
+    // &str as well?
```
Xophmeister, line 120: I quite liked the idea of the formatter being genericised over `io::{Read, Write}` and it makes sense when the configuration parsing is internal. However, yes, if the configuration parsing is factored out, then the API feels a bit clumsy, so unifying it to `&str` seems reasonable.

Xophmeister (approved):
Full disclosure: I'm not familiar with Rust async -- but it looks like it follows the usual model -- and I'm afraid the Nix changes are beyond me.

That said, I can more-or-less understand the PR and your changes make sense to me...and if it works, it works :+1: I think it was a good idea to keep the formatter synchronous and delegate language loading, to keep the API relatively tidy.

src/bin/topiary/main.rs
torhovland: OK, I'll leave that in there as a comment to follow up on separately.

PR #340, opened 2023-03-13
Improve the escaping for GraphViz output

Xophmeister:
The node kind was escaped in the original GraphViz output. However, for anonymous nodes that contained escapable whitespace characters (e.g., `\n`), these ended up getting reinterpreted by GraphViz as their corresponding whitespace. That wasn't very helpful in the output, so now these characters are double-escaped, so they can be scrutinised.

### DIFF
--- a/src/graphviz.rs
+++ b/src/graphviz.rs
@@ -4,7 +4,6 @@ use std::{borrow::Cow, fmt, io};
 
 use crate::{tree_sitter::SyntaxNode, FormatterResult};
 
-// TODO Add tests for this
 /// We double-escape whitespace (\n and \t) so it is
 /// rendered as the escaped value in the GraphViz output
 fn escape(input: &str) -> Cow<str> {
@@ -92,4 +91,22 @@ pub fn write(output: &mut dyn io::Write, root: &SyntaxNode) -> FormatterResult<(
     writeln!(output, "}}")?;
 
     Ok(())
+}
+
+#[cfg(test)]
+mod test {
+    use super::escape;
+
+    #[test]
+    fn double_escape() {
+        // PBT would be handy, here
+        assert_eq!(escape("foo"), "foo");
+        assert_eq!(escape("'"), r#"\'"#);
+        assert_eq!(escape("\n"), r#"\\n"#);
+        assert_eq!(escape("\t"), r#"\\t"#);
+        assert_eq!(
+            escape("Here's something\nlonger"),
+            r#"Here\'s something\\nlonger"#
+        );
+    }
 }
--- a/src/graphviz.rs
+++ b/src/graphviz.rs
@@ -4,62 +4,58 @@ use std::{borrow::Cow, fmt, io};
 
 use crate::{tree_sitter::SyntaxNode, FormatterResult};
 
-// TODO Factor out the repetition here
 // TODO Add tests for this
+/// We double-escape whitespace (\n and \t) so it is
+/// rendered as the escaped value in the GraphViz output
 fn escape(input: &str) -> Cow<str> {
-    // We double-escape whitespace (\n and \t) so it is
-    // rendered as the escaped value in the GraphViz output
-    let mut is_escaped = false;
-    let mut buffer: String = String::new();
+    let mut buffer: Option<String> = None;
+
     let mut start: usize = 0;
     let length = input.len();
 
-    for (upto, c) in input.chars().enumerate() {
-        match c {
-            '\n' => {
-                if !is_escaped {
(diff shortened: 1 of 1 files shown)


src/graphviz.rs
```diff
@@ -34,3 +92,21 @@ pub fn write(output: &mut dyn io::Write, root: &SyntaxNode) -> FormatterResult<(
...
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+    use super::escape;
+
+    #[test]
+    fn double_escape() {
+        // PBT would be handy, here
```
torhovland, line 102: Better expand that TLA.

src/graphviz.rs
```diff
@@ -1,9 +1,67 @@
...
+
+    let mut start: usize = 0;
+    let length = input.len();
+
+    let append = |buffer: &mut Option<String>, from: &mut usize, to: usize, suffix: &str| {
+        // Allocate buffer only when necessary
+        if buffer.is_none() {
+            // Best case:  length + 1  (i.e., single escaped character in input)
+            // Worst case: length * 3  (i.e., every character needs double-escaping)
+            *buffer = Some(String::with_capacity(length * 2));
```
torhovland, line 20: So why multiply with two here (and not by three)?

torhovland: For simplicity, I would just do:

```
        if buffer.is_none() {
            *buffer = Some(String::new());
        }
```

src/graphviz.rs
```diff
@@ -1,9 +1,67 @@
...
+        }
+
+        if let Some(ref mut buffer) = buffer {
```
torhovland, line 23:
```suggestion
        if let Some(buffer) = buffer {
```

torhovland (approved):

torhovland:
The buffer doesn't have to be an `Option`, and that simplifies things:

```rust
fn escape(input: &str) -> Cow<str> {
    let mut buffer = String::new();

    let mut start: usize = 0;
    let length = input.len();

    let append = |buffer: &mut String, from: &mut usize, to: usize, suffix: &str| {
        // Decant the unescaped chunk from the input,
        // followed by the escaped suffix provided
        *buffer += &input[*from..to];
        *buffer += suffix;

        // Fast-forward the tracking cursor to the next character
        *from = to + 1;
    };

    for (idx, current) in input.chars().enumerate() {
        match current {
            // Double-escape whitespace characters
            '\n' => append(&mut buffer, &mut start, idx, r#"\\n"#),
            '\t' => append(&mut buffer, &mut start, idx, r#"\\t"#),

            otherwise => {
                // If char::escape_default starts with a backslash, then we
                // have an escaped character and we're off the happy path
                let mut escaped = otherwise.escape_default().peekable();
                if escaped.peek() == Some(&'\\') {
                    append(
                        &mut buffer,
                        &mut start,
                        idx,
                        &otherwise.escape_default().to_string(),
                    );
                }
            }
        }
    }

    if buffer.is_empty() {
        input.into()
    } else {
        // Decant whatever's left of the input into the buffer
        buffer += &input[start..length];
        buffer.into()
    }
}
```

torhovland:
Also, it might be useful to test the `Cow` mechanism:

```rust
    #[test]
    fn unescaped_borrowed() {
        match escape("foo") {
            Cow::Borrowed("foo") => {}
            _ => {
                panic!("expected a borrowed, unmodified str")
            }
        }
    }

    #[test]
    fn escaped_owned() {
        match escape("'") {
            Cow::<str>::Owned(s) => {
                assert_eq!(s, r#"\'"#.to_string());
            }
            _ => {
                panic!("expected an owned, escaped string")
            }
        }
    }
```

src/graphviz.rs
Xophmeister: Here I'm trying to reduce the number of allocations required as the escaped string gets built up. I used `Option<String>` originally, because I didn't realise that `String::new` _didn't_ allocate. I've just confirmed that now from the manual. As such, I'll get rid of the `Option`, per your other comment, and switch this to `buffer.reserve(whatever)`.

The buffer size is set to `length * 2` as a "happy medium". Given that the strings it encounters will be very short, `length * 3` makes sense too :+1:

src/graphviz.rs
Xophmeister: Property based testing, à la [`quickcheck`](https://crates.io/crates/quickcheck) or [`proptest`](https://crates.io/crates/proptest).

torhovland: Yeah, I meant in the code comment 🙂

PR #342, opened 2023-03-14
Build and deploy Wasm app

torhovland:
- Build Wasm playground with Nix.
- The most rudimentary playground frontend imaginable (formats `{"foo":"bar"}` in the browser console).
- Build and deploy playground in CI.
- Scripts for local development of web site and playground.
- Optimizing Wasm for size.


update-web-playground.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 nix build .#web-playground
 rm -rf website/playground/wasm-app
-ln -s ../../result/post website/playground/wasm-app
+ln -s ../../result website/playground/wasm-app
```
Xophmeister: Is `website/playground/wasm-app` always a symlink? In which case, (FWIW) you could update it without deleting first with `ln -sf`.

CONTRIBUTING.md
```diff
@@ -55,11 +55,19 @@ the `generate-coverage.sh` script can be run from the root of this
 repository to avoid contaminating your environment, but note it will
 download a full toolchain on each run.
 
-## Website and web playground
+## Web site and web playground
 
 If you have [Deno](https://deno.land/) installed, you can start a local web
 server like this:
```
Xophmeister: I wonder if it's worth bundling something with the Nix stuff to include additional tools, like Deno and Graphviz? I know you can put them behind arguments in a normal `shell.nix`, but I don't know how it works with flakes.

Anyway, not for this PR regardless, but perhaps worth an issue...

Xophmeister (approved):

update-web-playground.sh
torhovland: I actually tried that first, but I got a permission issue (at least on my Mac) when I tried to recreate it.

PR #343, opened 2023-03-14
Fix premature draining of query buffer

Xophmeister:
Change the `formatter` API such that it takes the query as a `&str`, rather than `io::Read`. Previously, the `BufReader` was getting drained in `src/bin/topiary/main.rs` to parse the configuration, leaving the formatter with an empty query buffer.

### DIFF
--- a/src/bin/topiary/main.rs
+++ b/src/bin/topiary/main.rs
@@ -113,13 +113,14 @@ async fn run() -> CLIResult<()> {
         unreachable!();
     };
 
-    let mut query_reader = BufReader::new(File::open(query_path)?);
+    let query = {
+        let mut reader = BufReader::new(File::open(&query_path)?);
+        let mut contents = String::new();
+        reader.read_to_string(&mut contents)?;
 
-    // It's not very nice that formatter wants an io::Read and
-    // Configuration::parse wants a &str. Should we just let formatter take
-    // &str as well?
-    let mut query = String::new();
-    query_reader.read_to_string(&mut query)?;
+        contents
+    };
+
     let mut configuration = Configuration::parse(&query)?;
 
     // Replace the language deduced from the query file by the one from the CLI, if any
@@ -142,7 +143,7 @@ async fn run() -> CLIResult<()> {
     formatter(
         &mut input,
         &mut output,
-        &mut query_reader,
+        &query,
         &configuration,
         &grammars,
         operation,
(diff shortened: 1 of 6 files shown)


Xophmeister:
I've not written any tests, as I'm not sure how I would write one that would exhibit the old behaviour to catch regressions...

Xophmeister:
The WASM build is failing and I'm not sure where it needs to be fixed:

```
error: builder for '/nix/store/v1r1z09xc6sx06m42523wvmpd6nm02xk-topiary-0.1.0.drv' failed with exit code 101;
       last 10 log lines:
       >     = note:      expected reference `&str`
       >             found mutable reference `&mut &[u8]`
       > note: function defined here
       >    --> /build/source/src/lib.rs:155:8
       >     |
       > 155 | pub fn formatter(
       >     |        ^^^^^^^^^
       >
       > For more information about this error, try `rustc --explain E0308`.
       > error: could not compile `topiary-playground` due to previous error
       For full logs, run 'nix log /nix/store/v1r1z09xc6sx06m42523wvmpd6nm02xk-topiary-0.1.0.drv'.
```

torhovland:
> I've not written any tests, as I'm not sure how I would write any that would exhibit the old behaviour to catch regressions...

We just need a test that runs the CLI and verifies that it produces formatted output. But we can come back to that.

torhovland:
> NB: My first attempt at fixing this changed my `Cargo.lock` file, upgrading `windows_i686_msvc` to `0.42.2`. This broke the build for some reason -- version conflicts, I guess -- so I just reverted the `Cargo.lock`.

This is a mess, but I'm hoping we can sort it out when we move to a workspace.

src/bin/topiary/main.rs
```diff
@@ -113,13 +113,14 @@ async fn run() -> CLIResult<()> {
         unreachable!();
     };
 
-    let mut query_reader = BufReader::new(File::open(query_path)?);
+    let query = {
```
torhovland, line 116: Neat!

src/lib.rs
```diff
@@ -166,18 +166,12 @@ pub fn formatter(
             e,
         ))
     })?;
-    let query = read_input(query).map_err(|e| {
-        FormatterError::Io(IoError::Filesystem(
-            "Failed to read query contents".into(),
```
torhovland, line 171: We may want to move this error message somewhere else now. I suppose now you don't get a very nice message if you reference a query file that doesn't exist.

torhovland (approved):

PR #347, opened 2023-03-15
Test that formatting happens

### LINKED ISSUE #344: Adapt `cli-tester` so it verifies that code has been formatted
Our CLI broke (#343) without the tests noticing. It stopped applying the query, resulting in unformatted, raw Tree-sitter output.

We already have `cli-tester.rs`, but it doesn't assert that any input code has been modified by the formatter. It should.

Xophmeister:
Modify `cli-tester.rs` such that it confirms that _some_ formatting (hopefully the correct formatting) has occurred. This should catch the error fixed by PR #343.

Resolves #344

### DIFF
--- a/tests/cli-tester.rs
+++ b/tests/cli-tester.rs
@@ -6,8 +6,10 @@ use std::{
 };
 use tempfile::NamedTempFile;
 
-// Exemplar JSON state that won't be affected by the formatter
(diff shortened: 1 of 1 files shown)


torhovland (approved):

torhovland:
I see that it fails when I modify `JSON_EXPECTED`. Good1

PR #348, opened 2023-03-15
Basic playground frontend

torhovland:
- Basic playground.
- Format errors.
- Add grammar wasm files.
- Show status in case of long-running format operation.
- Docs on how to update grammars.


Xophmeister (approved):
LGTM :100:

Are the WASM files checked in just as a convenience, for now? Presumably the CI can build and deploy them, in the future...

torhovland:
It can, yes, but they take some time. Not sure if it makes sense to put that on the CI. Anyway, we can discuss that later.

PR #352, opened 2023-03-16
Playground frontend in React

torhovland:
Implements a basic version of #171.

A version without sample input and query is already deployed to https://topiary.tweag.io/playground.

- Move wasm-app to subdir.
- Working playground in React.
- favicon, etc.
- Layout.
- Button style.
- Switch to TypeScript.
- Move Tree-sitter files.
- Load Tree-sitter.
- Rename script.
- Remove prototype web playground.
- Build React app in CI.
- Sample input and query.


.github/workflows/ci.yml
Xophmeister, line 1: Not for this PR, but does the building of the WASM app need to happen on both Ubuntu and macOS? On the one hand, it proves it works on macOS, but isn't the main utility here to stage for deployment, which only needs to be done once?

web-playground/react-app/README.md
Xophmeister, line 1: I guess it doesn't hurt to keep this vendored `README`... (You see it all the time. It wouldn't surprise me if Facebook's Ts&Cs mandated it :sweat_smile:)

Xophmeister (approved):

web-playground/react-app/README.md
torhovland: Yeah, I tried to modify the template as little as possible. We can always delete more later.

.github/workflows/ci.yml
torhovland: You're right. It's important to check that the CLI app builds on multiple architectures, but it doesn't really matter for the playground.

Xophmeister:
Ah nuts: I resolved the conversations and it auto-merged... Let me restore the branch, at least.

PR #353, opened 2023-03-16
Nickel: Annotations

Xophmeister:
This PR introduces:

* [x] Formatting rules for annotated symbol definitions
* [x] Formatting rules for nested annotations
* [x] Updated Nickel integration tests
* [x] Exposed Nickel as a supported language in the CLI

This will flow `|`-annotations on to new lines, when the annotations as a whole cover multiple lines. It will not reflow type annotations, because they can be nested and this causes idempotency issues (there doesn't appear to be a way of establishing the depth of the nesting from the grammar alone). For example:

```nickel
{
  single-line : type | annot | annot = whatever,

  multi-line : type
    | annot
    | annot
    = whatever
}
```

For let bindings, the RHS will appear indented, on a new line, if it covers multiple lines:

```nickel
let single-line = fun x => x

let multi-line =
  fun x => x
```

Likewise for function definitions:

```nickel
let multi-line-function =
  fun x =>
    x
```

Miscellaneous stuff:
* Allow blank lines between record fields.
* Designate builtins as leaves so their surrounding `%` characters don't get parsed as mod operators.
* Began to tidy up the query file, with comments and explanations of formatting choices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "487f1e0fcbe47deb8b0574e646def1c903389d95241dd1bbcc6ce4a715dfc0c1"
+
+[[package]]
 name = "bstr"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -158,7 +164,7 @@ version = "3.2.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "71655c45cb9845d3270c9d6df84ebe72b4dad3c2ba3f7023ad47c144e4e473a5"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "clap_lex 0.2.4",
  "indexmap",
  "textwrap",
@@ -166,11 +172,11 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.1.10"
+version = "4.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ce38afc168d8665cfc75c7b1dd9672e50716a137f433f070991619744a67342a"
+checksum = "42dfd32784433290c51d92c438bb72ea5063797fc3cc9a21a8c4346bebbb2098"
 dependencies = [
- "bitflags",
+ "bitflags 2.0.2",
  "clap_derive",
  "clap_lex 0.3.3",
  "is-terminal",
@@ -525,9 +531,9 @@ dependencies = [
 
 [[package]]
 name = "io-lifetimes"
-version = "1.0.7"
+version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76e86b86ae312accbf05ade23ce76b625e0e47a255712b7414037385a1c05380"
+checksum = "09270fd4fa1111bc614ed2246c7ef56239a3063d5be0d1ec3b589c505d400aeb"
 dependencies = [
  "hermit-abi 0.3.1",
  "libc",
@@ -536,9 +542,9 @@ dependencies = [
 
 [[package]]
 name = "is-terminal"
-version = "0.4.4"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21b6b32576413a8e69b90e952e4a026476040d81017b80445deda5f2d3921857"
+checksum = "8687c819457e979cc940d09cb16e42a1bf70aa6b60a549de6d3a62a0ee90c69e"
 dependencies = [
  "hermit-abi 0.3.1",
  "io-lifetimes",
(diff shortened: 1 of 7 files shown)


Xophmeister:
(Ignore the CI for now...)

Xophmeister:
There's a double indent somewhere and there's an idempotency problem on, e.g.:

```nickel
{
  foo
    : forall a. Array { field : String, value : a } -> a
    | doc "foo"
    = 1
}
```

aspiwack:
I think this PR is at high risk of turning into a bit of a rabbit hole. I suggest you split it in smaller PRs each making progress. Everytime you have an improvement, you can merge it. If anything, you'll feel better that way.

Xophmeister:
> I think this PR is at high risk of turning into a bit of a rabbit hole. I suggest you split it in smaller PRs each making progress. Everytime you have an improvement, you can merge it. If anything, you'll feel better that way.

This PR can't (or, at least, "probably shouldn't") be merged until the upstream changes are stabilised. That said, I can still set this branch as a base and merge smaller PRs into it :+1:

aspiwack:
@yannham What is the status of the unstable branch of the tree-sitter  grammar? (in particular: why does it exist?)

torhovland:
> This PR can't (or, at least, "probably shouldn't") be merged until the upstream changes are stabilised. That said, I can still set this branch as a base and merge smaller PRs into it 👍

I don't think depending on a particular commit of the Nickel grammar should be keeping us from merging this. That is simply the best version of the grammar that we have. We can always come back and update our dependency later.

Xophmeister:
I agree with @torhovland: There's no harm in pinning a dependency to a commit, regardless of which branch it happens to be on... I'll split this PR up :slightly_smiling_face:

yannham:
@aspiwack the rationale is that the master branch of tree-sitter, of nickel-lang.org (including the documentation), etc. all points to the current stable version of Nickel (e.g. if you use nvim-tree-sitter then it automatically picks the version from the master branch, and current stable and current master are quite deeply incompatible). Be we all use the latest version, so we put the up-to-date grammar on the `unstable` branch for now.

yannham:
> This will flow |-annotations on to new lines, when the annotations as a whole cover multiple lines. It will not reflow type annotations, because they can be nested and this causes idempotency issues (there doesn't appear to be a way of establishing the depth of the nesting from the grammar alone).

Out of curiosity, what's special about type annotations? Is that you encountered concreted cases with nesting? Because, on paper, contract annotations and type annotations accept exactly the same syntax on the right of the separator `|`/`:`, which is mostly an arbitrary nickel expression (almost, if you put other type/contract annotations, you might have to add parentheses). I would expect them to have the same characteristics, and thus the same issues.

Xophmeister:
> Out of curiosity, what's special about type annotations? Is that you encountered concreted cases with nesting? Because, on paper, contract annotations and type annotations accept exactly the same syntax on the right of the separator `|`/`:`, which is mostly an arbitrary nickel expression (almost, if you put other type/contract annotations, you might have to add parentheses). I would expect them to have the same characteristics, and thus the same issues.

It's a good question that, I'm afraid, I don't have a satisfactory answer for :sweat_smile: I was getting idempotency errors when trying to format this from the stdlib:

```nickel
let partition : forall a. (a -> Bool) -> Array a -> { right : Array a, wrong : Array a }
```

It has a multi-line context, because of the `| doc`, so in the first parse it turns into:

```nickel
let partition
  : forall a. (a -> Bool) -> Array a -> { right
      : Array a, wrong
      : Array a }
```

Besides the double indent, this makes sense; even though it looks horrible. However, some other rule is then turning it into:

```nickel
let partition
  : forall a. (a -> Bool) -> Array a -> {
      right
      : Array a,
      wrong
      : Array a }
```

This is being caused by some other rule that I've yet to "internalise". Once I find it -- and tweak it -- I can possibly relax the condition on type annotations...

torhovland (approved):
Just putting an approval here. Merge when any discussions are resolved.

tests/samples/expected/nickel.ncl
```diff
@@ -4,14 +4,19 @@
 {
   arrays_arrays = let my_array_lib =
     {
-      map : forall a b. (a -> b) -> Array a -> Array b = fun f arr =>
+      map : forall a b. (a -> b) -> Array a -> Array b
+        | doc m%"
+          Here is some documentation
+          "%
```
vkleen, line 10: I think we'd usually like to indent the documentation once further, so
```suggestion
        | doc m%"
            Here is some documentation
          "%
```

aspiwack:
> the rationale is that the master branch of nickel-tree-sitter, of nickel-lang.org (including the documentation), etc. all point to the current stable version of Nickel (e.g. if you use nvim-tree-sitter then it automatically picks the version from the master branch, and current stable and current master are quite deeply incompatible). Be we all use the latest version, so we put the up-to-date grammar on the `unstable` branch for now.

Thanks. So most definitely no obstacle on depending on the `unstable` branch in Topiary.

tests/samples/expected/nickel.ncl
Xophmeister: No luck, I'm afraid. If I try to format a multiline string as if it were just another syntactic construction, then strange things happen:
1. If I force a newline after the multiline string start token, it breaks idempotency because it keeps trying to insert a new line on each run.
2. The indentation -- which was the hope that formatting would address -- appears to be unaffected.

As such, I think the safest option for now would treat Nickel strings no differently than strings in any other language; i.e., they shouldn't be formatted and it's up to the input to be aligned manually (which I've done in 6888c0f7d1f0).

PR #354, opened 2023-03-16
Introduce conditional predicates to reduce number of directives

### LINKED ISSUE #337: Use predicates to avoid verbose actions
With the current way we are increasing the number of Topiary actions, we might consider using predicates to reduce this number. Probably using predicates like
```scheme
#is_multiline?
#is_singleline?
```

nbacquey:
### Testing context with predicates

Sometimes, similarly to what happens with softlines, we want a query to match only if the context is single-line, or multi-line. This PR introduces several predicates that achieve this result.

### `#single_line_only!` / `#multi_line_only!`

These predicates allow the query to trigger only if the matched nodes are in a single-line (resp. multi-line) context.

#### Example

```scheme
; Allow (and enforce) the optional "|" before the first match case
; in OCaml if and only if the context is multi-line
(
  "with"
  .
  "|" @delete
  .
  (match_case)
  (#single_line_only!)
)

(
  "with"
  .
  "|"? @do_nothing
  .
  (match_case) @prepend_delimiter
  (#delimiter! "| ")
  (#multi_line_only!)
)
```

### `#single_line_scope_only!` / `#multi_line_scope_only!`

These predicates allow the query to trigger only if the associated custom scope containing the matched nodes are is single-line (resp. multi-line).

#### Example

```scheme
; Allow (and enforce) the optional "|" before the first match case
; in function expressions in OCaml if and only if the scope is multi-line
(function_expression
  (match_case)? @do_nothing
  .
  "|" @delete
  .
  (match_case)
  (#single_line_scope_only! "function_definition")
)
(function_expression
  "|"? @do_nothing
  .
  (match_case) @prepend_delimiter
  (#multi_line_scope_only! "function_definition")
  (#delimiter! "| ") ; sic
)
```

This PR uses those new predicates to reduce the number of verbose formatting directives in `ocaml.scm`, and removes deprecated atoms and formatting directives.

Closes #337

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -799,16 +799,16 @@
 (function_expression
   (match_case)? @do_nothing
   .
-  "|" @singleline_scoped_delete
+  "|" @delete
   .
   (match_case)
-  (#scope_id! "function_definition")
+  (#single_line_scope_only! "function_definition")
 )
 (function_expression
   "|"? @do_nothing
   .
-  (match_case) @prepend_scoped_multiline_delimiter
-  (#scope_id! "function_definition")
+  (match_case) @prepend_delimiter
+  (#multi_line_scope_only! "function_definition")
   (#delimiter! "| ") ; sic
 )
 (function_expression
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -690,9 +690,10 @@
 (
   "with"
   .
-  "|" @singleline_delete
+  "|" @delete
   .
   (match_case)
+  (#single_line_only! )
 )
 
 (
@@ -700,8 +701,9 @@
   .
   "|"? @do_nothing
   .
-  (match_case) @prepend_multiline_delimiter
+  (match_case) @prepend_delimiter
   (#delimiter! "| ") ; sic
+  (#multi_line_only! )
 )
 
 ; Multi-line definitions must have a linebreak after "=" and before "in":
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -5,7 +5,7 @@ use std::{
 
 use tree_sitter_facade::Node;
 
-use crate::{Atom, FormatterError, FormatterResult};
+use crate::{Atom, FormatterError, FormatterResult, ScopeCondition};
 
 #[derive(Debug)]
 pub struct AtomCollection {
@@ -60,6 +60,29 @@ impl AtomCollection {
         Ok(atoms)
     }
 
+    // wrap inside a conditional atom if #single/multi_line_scope_only! is set
+    fn wrap(&mut self, atom: Atom, predicates: &QueryPredicates) -> Atom {
+        if let Some(scope_id) = &predicates.single_line_scope_only {
+            let id = self.next_id();
+            Atom::ScopedConditional {
+                id,
+                scope_id: scope_id.to_string(),
+                condition: ScopeCondition::SingleLineOnly,
+                atom: Box::new(atom),
+            }
+        } else if let Some(scope_id) = &predicates.multi_line_scope_only {
+            let id = self.next_id();
+            Atom::ScopedConditional {
+                id,
+                scope_id: scope_id.to_string(),
+                condition: ScopeCondition::MultiLineOnly,
+                atom: Box::new(atom),
+            }
+        } else {
+            atom
+        }
+    }
+
     /// This gets called a lot during query processing, and needs to be efficient.
     pub fn resolve_capture(
         &mut self,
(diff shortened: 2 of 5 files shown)


aspiwack:
All these predicates make sense. But only having `#single_line_only!` and `#multi_line_only!` does invite pattern duplication.

I wonder how feasible (not for this PR) it is to go further than this. And have `#desugar_if_single_line!` and `#desugar_if_multi_line!` predicates with the following behaviour:

- expected syntax `#desugar_if_X_line! <capture> <catpure>*`
- semantics: in X line nodes, `#desugar_if_X_line! @name @capture_1 @capture_n` behaves as if we had written `@capture_1 @capture_n` where `@name` is in the query.

Example

```
(foo
   bar @sol
   (#desugar_if_single_line! @sol @append_space)
   (#desugar_if_multi_line! @sol @append_newline)
)
```

This makes `@sol` equivalent to `@append_spaced_softline` (I think).

We could consider the possibility to define `#desugar_*` statement at toplevel to apply to all patterns in the query file.

nbacquey:
I've been thinking about similar stuff, but I don't think pattern duplication would happen *that* often.
I think that the `spaced_softline` directives currently are the only ones that would suffer this duplication if we rewrite them with `#single/multi_line_only`. We don't *have* to do that: it's OK to keep the spaced softline directives, even if we could rewrite them with `#single/multi_line_only`.

aspiwack:
(by the way: the name “desugar” is not great, I suggest “expand” instead)

I'll reiterate that I don't want this PR to change, it is most likely to introduce a desirable feature. But I think that, contrary to your point, the choice of “what do I do when I'm single line vs multiline” is quite common. Central even to the whole engine design. There are all the spaced softline variants. There are things like “add a | if multiline, remove a | is single line”. Similar thoughts with terminal delimiters.

At any rate, from a language design perspective, saying that our primitives are not magic but can be defined in terms of a more fundamental construct is always a good thing. So I still believe that they are valuable. But, again, not for this PR.

nbacquey:
I have noted that you do not think this PR should change.

> But I think that, contrary to your point, the choice of “what do I do when I'm single line vs multiline” is quite common. Central even to the whole engine design.

I agree with that. My point is that the answer to "what do I do when I'm single line vs multiline ?” is often "nothing" in one case or the other. If only one context (single-line or multi-line) leads to an actual action, then I would argue that using `#X_line_only!` does not lead to pattern duplication. I would even rather have
```scheme
(foo
   bar @do_something
   (#single_line_only!)
)
```
than
```scheme
(foo
   bar @sol
   (#expand_if_single_line! @sol @do_something)
   (#expand_if_multi_line! @sol @do_nothing)
)
```

nbacquey:
I don't think the CI fails because of the PR. I'll relaunch it after I get a review.
Speaking of which: this PR is now open for reviews.

languages/ocaml.scm
```diff
@@ -690,18 +690,20 @@
 (
   "with"
   .
-  "|" @singleline_delete
+  "|" @delete
   .
   (match_case)
+  (#single_line_only! )
```
Xophmeister, line 696:
```suggestion
  (#single_line_only!)
```

languages/ocaml.scm
```diff
@@ -690,18 +690,20 @@
...
 
 (
   "with"
   .
   "|"? @do_nothing
   .
-  (match_case) @prepend_multiline_delimiter
+  (match_case) @prepend_delimiter
   (#delimiter! "| ") ; sic
+  (#multi_line_only! )
```
Xophmeister, line 706:
```suggestion
  (#multi_line_only!)
```

Xophmeister (approved):
Looks great :+1:

Just thinking out aloud -- no need for this to block this PR, etc. -- but is there any need for both `(#{single,multi}_line_only!)` and `(#{single,multi}_line_scope_only!)`. Can the second form not be inferred from the presence of a `(#scope_id!)` predicate?

topiary/src/atom_collection.rs
```diff
@@ -60,129 +60,127 @@ impl AtomCollection {
...
             })
         };
         let requires_scope_id = || {
-            scope_id.ok_or_else(|| {
+            predicates.scope_id.as_deref().ok_or_else(|| {
                 FormatterError::Query(format!("@{name} requires a #scope_id! predicate"), None)
             })
         };
 
+        let mut is_multi_line = false;
```
torhovland, line 106: Can be made non-mut:

```
let is_multi_line = if let ...
```

languages/ocaml.scm
nbacquey: Unfortunately this is not a typo: see #383 
This is currently how predicates without parameters are formatted. If the space is removed, the tests fail because query file is no longer "correctly" formatted.

topiary/src/atom_collection.rs
```diff
@@ -60,129 +60,127 @@ impl AtomCollection {
         Ok(atoms)
     }
 
+    // wrap inside a conditional atom if #single/multi_line_scope_only! is set
+    fn wrap(&mut self, atom: Atom, predicates: &QueryPredicates) -> Atom {
```
torhovland, line 64: This name is too generic. Wrap what? 🙂

nbacquey:
Maybe we can say that `(#{single,multi}_line_scope_only!)` is `(#{single,multi}_line_only!)` + `(#scope_id!)`, I hadn't thought about that. The problem is, `#scope_id!` is also used for `scoped_softline` directives, which means the predicates above and `scoped_softline` should always refer to the same `scope_id` when used together in a query. Maybe we want that, maybe we don't. I'm not sure at the moment.

topiary/src/tree_sitter.rs
```diff
@@ -225,34 +224,82 @@ fn collect_leaf_ids(matches: &Vec<LocalQueryMatch>, capture_names: &[String]) ->
...
+            ..predicates.clone()
+        })
+    } else if let "scope_id!" = operator {
         let arg =
             predicate.args().into_iter().next().ok_or_else(|| {
                 FormatterError::Query(format!("{operator} needs an argument"), None)
             })?;
+        Ok(QueryPredicates {
+            scope_id: Some(arg),
+            ..predicates.clone()
```
torhovland, line 248: All the clones here look suspect, but I've not studied the code enough to see if you really need them. Have you considered if there is a better way?

torhovland (approved):

PR #373, opened 2023-03-20
Update paths to test files

### LINKED ISSUE #371: Playground INPUT_SOURCE not optional
**Describe the bug**
Running `./playground.sh "json"` results in an error:
```
❯ ./playground.sh "json"
find: ‘tests/samples/input’: No such file or directory
Error: Couldn't find input source file ''

Usage: playground.sh (LANGUAGE | QUERY_FILE) [INPUT_SOURCE]

LANGUAGE can be one of the supported languages (e.g., "ocaml", "rust",
etc.); alternatively, give the path to the query file itself, as
QUERY_FILE.

The INPUT_SOURCE is optional. If not specified, it defaults to trying
to find the bundled integration test input file for the given language.
```

Input source should be optional, but clearly isn't.

**To Reproduce**
Run `./playground.sh "json"

**Expected behavior**
The playground would run with the json queries.

torhovland:
Fixes #371.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ If you do `cargo install flamegraph`, you can generate a performance flamegraph
 like this:
 
 ```bash
-CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph -- -l ocaml < tests/samples/input/ocaml.ml > formatted.ml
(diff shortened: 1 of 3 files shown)


Xophmeister (approved):
LGTM :100: 

I fixed the directory in `playground.sh` in #372, but didn't update any documentation. It's exactly the same change, so it won't cause any conflicts.

PR #376, opened 2023-03-21
Support arbitrary indentation (#105)

### LINKED ISSUE #105: Hard-tab support
At the risk of starting a Holy War, `(#indent-level! 4)` implies soft-tabs; i.e., a number of spaces. _Some_ people prefer hard-tabs (i.e., a single `\t` at each indentation level) and some codebases -- misguided though they may be -- expect them per their style guide.

I propose updating the format of this directive such that its parameter is either:
* A digit, representing the number of spaces in a soft-tab (as now);
* Or a `t`, say, representing hard-tab indentation.

ErinvanderVeen:
Resolves #105 

I was/am not sure about the regular expressions parsing of the arguments. They used to support multiple arguments, but I've reduced that to 1 because I could not reconcile that easily with "   " string arguments.

Escaping is done by an old crate, but more modern alternatives are all GPL licensed. This isn't an issue to us, but could be a problem to future users of our library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,17 +292,16 @@ this:
 ```scheme
 ; Configuration
 (#language! rust)
-(#indent-level! 4)
+(#indent! "    ")
 ```
 
 The `#language!` pragma must be included in any query file and dictates
 which language to parse. The queries themselves will refer to node types
 that are specific to this language.
 
-The `#indent-level!` pragma is optional and controls how many spaces to
-indent a block whenever `@append_indent_start` or
-`@prepend_indent_start` is processed. If it is omitted, it defaults to
-two spaces.
+The `#indent!` pragma is optional and controls what indentation to use to
+indent a block whenever `@append_indent_start` or `@prepend_indent_start` is
+processed. If it is omitted, it defaults to two spaces.
 
 ### `@allow_blank_line_before`
 
--- a/topiary/src/configuration.rs
+++ b/topiary/src/configuration.rs
@@ -1,19 +1,20 @@
 use crate::{language::Language, FormatterError, FormatterResult};
 use regex::Regex;
+use unescape::unescape;
 
 pub struct Configuration {
     pub language: Language,
-    pub indent_level: usize,
+    pub indent: String,
 }
 
 impl Configuration {
     pub fn parse(query: &str) -> FormatterResult<Self> {
         let mut language: Option<Language> = None;
-        let mut indent_level: usize = 2;
+        let mut indent: String = String::from("  ");
 
         // Match lines beginning with a predicate like this:
         // (#language! rust)
-        // (#indent-level! 4)
+        // (#indent! "    ")
         // (#foo! 1 2 bar)
         let regex =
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -292,17 +292,16 @@ this:
 ```scheme
 ; Configuration
 (#language! rust)
-(#indent-level! 4)
+(#indent! "    ")
```
aspiwack, line 295: Because of how common using multiple spaces is, and how unreadable a bunch of spaces between double-quotes is, I would avise keeping `(#indent-level! <num>)` as an alternative to `(#indent! <string>)`.

(maybe use `#indent! "    "` in one of the query file in order to make sure that the code is used)

ErinvanderVeen: Hej!

There are a few reasons that I hesitate to do that.

1. I want to keep this code as simple as possible because it's quite hacky as is.
2. We are moving to a toml based configuration file before we move to 0.2, which could possibly* allow both integers and strings in a single field.

* Not 100% about this, TBD

ErinvanderVeen: Maybe, even then, I think this approach is good enough for people to get started with other forms of indentation, without putting too much effort into it. Effort that is likely to be wasted when moving to a configuration file.

topiary/src/configuration.rs
```diff
@@ -41,20 +39,32 @@ impl Configuration {
...
+
+                        if arg.chars().all(char::is_whitespace) {
+                            indent = arg;
+                        } else {
+                            return Err(FormatterError::Query(
+                                format!(
+                                    "The #indent! parameter must only contain whitespace, but got '{arg}'"
+                                ),
+                                None,
+                            ));
```
Xophmeister, line 63: Aw, no emoji indentation :disappointed:

topiary/src/configuration.rs
```diff
@@ -23,16 +24,13 @@ impl Configuration {
                 .name("predicate")
                 .expect("predicate capture group")
                 .as_str();
-            let mut arguments = capture
-                .name("arguments")
-                .expect("arguments capture group")
-                .as_str()
-                .split(' ');
-            log::info!("Predicate: {predicate} -  Arguments: {arguments:?}");
+            let argument = capture.name("arguments").map(|arg| arg.as_str());
```
Xophmeister, line 27: Is this just for expediency's sake, as we don't yet have any multi-argument predicates?

topiary/src/configuration.rs
Xophmeister, line 63: Aw, no emoji indentation :disappointed:

Xophmeister (approved):
LGTM :100:

topiary/src/configuration.rs
ErinvanderVeen: Yeah! We used to split on spaces, which meant that `"   "` would be parsed as `\"` and `\"`, that's not so great in this case.

topiary/src/configuration.rs
ErinvanderVeen: Want to remove the check? I'd be happy to, it seems arbitrarily limiting anyway.

Xophmeister: No, I'm just joking :slightly_smiling_face: Limiting to whitespace is fine.

PR #377, opened 2023-03-21
Implemented antispaces

Xophmeister:
The PR introduces the concept of "antispaces", that can be inserted into the formatting stream with the `@append_antispace` and `@prepend_antispace` capture names. When an `Antispace` atom is encountered during post-processing, it and any runs of preceding and/or succeeding `Space` or `Antispace` atoms are removed from the stream. This ultimately allows spacing exception rules to be written more easily.

For example, if we define the JSON formatting rules such that _any_ named node is surrounded by spaces, this upsets the spacing around list and pair delimiters. However, we can now fix that easily with an antispace:

```scheme
(#language! json)

(string) @leaf
(_) @append_space @prepend_space

":" @prepend_antispace
"," @prepend_antispace @append_empty_softline

(object
  .
  "{" @append_spaced_softline @append_indent_start
  (_)
  "}" @prepend_spaced_softline @prepend_indent_end
  .
)

(array
  .
  "[" @append_spaced_softline @append_indent_start
  (_)
  "]" @prepend_spaced_softline @prepend_indent_end
  .
)
```

(:bulb: This is not objectively better than the current JSON formatting rules; it's about the same. It's just an example to illustrate the point.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,6 +497,26 @@ with space as delimiter.
 ] @append_space
 ```
 
+### `@append_antispace` / `@prepend_antispace`
+
+It is often the case that tokens need to be juxtaposed with spaces,
+except in a few isolated contexts. Rather than writing complicated rules
+that enumerate every exception, an "antispace" can be inserted with
+`@append_antispace` / `@prepend_antispace`; this will destroy any spaces
+(not newlines) from that node, including those added by other formatting
+rules.
+
+#### Example
+
+```scheme
+[
+  ","
+  ";"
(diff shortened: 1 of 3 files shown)


torhovland (approved):
Nicely done. I feel you should also find at least one place in the query files to use this, to sort of justify it.

Xophmeister:
> I feel you should also find at least one place in the query files to use this, to sort of justify it.

The Nickel rules await :rocket:

PR #378, opened 2023-03-21
Nickel spacing tweaks

Xophmeister:
* Don't add spaces before `,`, `;` and `.` delimiters[^1]
* Don't add spaces immediately inside parentheses
* Don't surround record opening/closing tokens with spaces
* Don't add spaces/newlines to empty records

[^1]: I've been liberal, here. I don't know if all these delimiters are attested.


torhovland (approved):
Nice!

PR #379, opened 2023-03-22
Update installation procedure

nbacquey:


README.md
```diff
@@ -111,7 +111,7 @@ The project can be built and installed with Cargo from the repository
 directory:
 
 ```bash
-cargo install --path .
+cargo install --path topiary-cli
 ```
 
 The `TOPIARY_LANGUAGE_DIR` variable can be set either at build time or
```
torhovland, line 117: Hmm, looks like the build time variable might be wrong now.

torhovland: We should say

```bash
TOPIARY_LANGUAGE_DIR=/path/to/languages cargo install --path topiary-cli
```

torhovland (approved):

README.md
nbacquey: Building with the variable set still works, I've just tested it on this branch.

torhovland: But that line in the doc should tell the user to do `TOPIARY_LANGUAGE_DIR=/path/to/languages cargo install --path topiary-cli` because otherwise `topiary` won't just work when you invoke it from a directory that doesn't have a `languages` dir.

nbacquey: Won't it work if the variable is set at runtime as well?
I'm not sure about what you're saying. Has there been a feature change invalidating this paragraph, except that now topiary can also check `../languages`?
```
The `TOPIARY_LANGUAGE_DIR` variable can be set either at build time or
runtime. It should point to the directory where Topiary's language query
files (`.scm`) are located. Otherwise, Topiary will fall back to the
`languages` directory in the current working directory.
```

PR #384, opened 2023-03-23
Simplify workspace building

torhovland:
No longer fails on `cargo build` and `cargo run`. And the latter can now be used to run the CLI, like before.

Simplified the Nix build a little because of this.

Going back to using target_arch rather than features for conditional compilation, in order to be compatible with tree-sitter-facade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ echo '{"foo":"bar"}' | topiary --language json
 if you have those installed:
 
 ```bash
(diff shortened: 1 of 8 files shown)


torhovland:
@Xophmeister What I changed in Nix is:

- It's not necessary to invoke Clippy on all 3 crates, just `cargo clippy` and `cargo clippy -p topiary-playground --target wasm32-unknown-unknown`.
- It's not necessary to specify which crate to run benchmarks on.
- It's not necessary to explicitly build the Topiary library crate.

PR #385, opened 2023-03-23
Odds-and-ends of Nickel formatting

Xophmeister:
* [x] Conditionals
* [x] Annotations (again)
  * [x] Not reflowed on to a new line if:
    * [x] There is no equal sign, or
    * [x] The equal sign is "blocked" by some other node (e.g., a comment)
  * [x] Idempotency of multi-line nested type annotations
* [x] Multi-line sequences of infix operators
* [x] Formatting for `let...in...` expressions
* [x] Consistent formatting of let expressions and record fields
* [x] No spacing after prefix operators (`-` and `!`)
* [x] Generalise `(match)` and `(destruct)` statements to use record formatting rules

**Note** There is an idempotency issue with `let` expressions. However, the case in which it occurs is quite specific; it is not, for example, manifested by the Nickel standard library. It is documented in the query file and I will add an issue, but it shouldn't be a blocker for this PR.

### DIFF
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -86,10 +86,13 @@
 "(" @append_antispace
 ")" @prepend_antispace
 
-; Don't insert spaces between the negation operator and its operand
+; Don't insert spaces between infix operators and their operand
 (infix_expr
   .
-  "-" @append_antispace
+  [
+    "-"
+    (infix_u_op_5 "!")
+  ] @append_antispace
   .
   (infix_expr)
   .
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -142,7 +142,9 @@
 ; The binding expression should appear on a new line, indented, if its
 ; RHS is multi-line (pushing the "in" to an unindented new line).
 ; Similarly, the result expression (i.e., after the "in") should appear
-; on an indented new line, if that is multi-line.
+; on an new line, if that is multi-line. We don't start an indentation
+; block for the result expression, to avoid long diagonals in a series
+; of let expressions (which is idiomatic).
 
 (let_in_block
   (#scope_id! "let_binding_rhs")
@@ -169,10 +171,7 @@
 
 (let_expr
   (#scope_id! "let_result")
-  (let_in_block
-    "in" @append_indent_start
-  )
-  (term) @prepend_spaced_scoped_softline @append_indent_end
+  (term) @prepend_spaced_scoped_softline
 )
 
 ;; Annotations
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -203,6 +203,22 @@
   (term) @prepend_spaced_scoped_softline @prepend_indent_start @append_indent_end
 )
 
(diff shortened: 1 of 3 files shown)


languages/nickel.scm
```diff
@@ -86,83 +87,153 @@
...
+  (#scope_id! "let_binding_rhs")
+  "=" @begin_scope
+  .
+  (term)
+  .
+  "in" @end_scope
+)
+
+(let_in_block
+  (#scope_id! "let_binding_rhs")
```
torhovland, line 157: It seems that a scope isn't necessary here.

languages/nickel.scm
```diff
@@ -86,83 +87,153 @@
...
+(let_expr
+  (#scope_id! "let_result")
+  (let_in_block
+    "in" @begin_scope
+  )
+  (term) @end_scope
+)
+
+(let_expr
+  (#scope_id! "let_result")
```
torhovland, line 171: It seems that a scope isn't necessary here either.

languages/nickel.scm
```diff
@@ -86,83 +87,153 @@
...
+
+; Flow a chain of infix expressions over new lines, in a multi-line
+; context. Note that we _don't_ want this to happen for comparison
+; operators, which fall under nodes (infix_b_op_7) and (infix_b_op_8).
+(uni_term
+  (#scope_id! "infix_chain")
+  (infix_expr) @begin_scope
+) @end_scope
+
+(infix_expr
```
torhovland, line 110: This pattern doesn't seem to make any difference in the sample file.

languages/nickel.scm
```diff
@@ -86,83 +87,153 @@
...
-;   {
-;     foo : {
-;       foo : String,
-;       bar : Number
-;     }
-;     = ...
-;   }
+; context.
 (
   (#scope_id! "signature")
```
torhovland, line 208: It seems that a scope isn't necessary here either.

languages/nickel.scm
```diff
@@ -187,11 +258,36 @@
   t1: (applicative) @append_space
 )
 
+;; Conditionals
+
+; Flow multi-line match cases into an indented block after the =>
+(match_case
```
torhovland, line 264: This pattern doesn't seem to make any difference in the sample file.

torhovland (commented):
Looks like we either have unnecessary patterns or we are missing some sample code.

Xophmeister:
I've updated the tests with the concatenated Nickel stdlib. This now hits every query in the Nickel formatting definitions.

I've also changed the `let...in...` indentation, mentioned by the Nickel team. I've yet to fix [this](https://github.com/tweag/nickel/pull/1198#discussion_r1147862341), because I'm still trying to find a minimal reproducing example. (When I do, I'll add it to the tests :innocent:)

aspiwack:
As always, we should merge this earlier rather than later as it's clear improvement. And when it's done, we can document what we want to improve further.

PR #388, opened 2023-03-24
Use Rc<Atom>

torhovland:
This PR changes `Vec<Atom>` to `Vec<Rc<Atom>>`. Why?

We were operating with atoms directly. Since they cannot be moved out of their atom collection, this meant they sometimes had to be cloned. We should be operating with references instead. But this introduces lifetime issues. Atoms reference themselves (through `ScopedConditional`), and needed a lifetime same as that of the atom collection, which is essentially static. The post-processing function tries to create a new vector of atoms, and then assign that to be the new atom collection. But that doesn't work, because the new vector elements are references to the old vector.

We could have introduced a new set of atoms, solely for the ownership, and then let all atom collections during processing reference that. But that would leak memory slightly (not a big problem in our scenario, admittedly) unless we added some manual memory management. To fix that we could have a set of `Rc<Atom>` instead. But if we do that, we don't really need the new set. We can just let our current atom collection hold `Rc<Atom>`.

So that's what this PR is doing.

The ergonomics of this solution are very nice. No lifetime issues to ponder. Almost felt like cheating!

What about performance?

It is essentially unchanged. Actually, just slightly improved.

### DIFF
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -1,7 +1,6 @@
 use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
-    ops::Deref,
     rc::Rc,
 };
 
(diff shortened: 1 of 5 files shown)


torhovland:
Not an improvement.

PR #391, opened 2023-03-27
More in-place processing and less cloning.

torhovland:
I'm happy to discuss the individual changes here, and whether or not they are improvements.

### DIFF
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -347,8 +347,8 @@ impl AtomCollection {
 
             self.prepend
                 .entry(target_node_id)
-                .and_modify(|atoms| atoms.push(atom.clone()))
-                .or_insert_with(|| vec![atom]);
+                .or_insert(vec![])
+                .push(atom);
         }
     }
 
@@ -366,8 +366,8 @@ impl AtomCollection {
 
             self.append
                 .entry(target_node_id)
-                .and_modify(|atoms| atoms.push(atom.clone()))
-                .or_insert_with(|| vec![atom]);
+                .or_insert(vec![])
+                .push(atom);
         }
     }
 
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -275,6 +275,8 @@ impl AtomCollection {
                     let prepends = self.prepend.entry(*id).or_default();
                     let appends = self.append.entry(*id).or_default();
 
+                    // Rather than cloning the atom from the old vector, we
+                    // swap it out with an empty one.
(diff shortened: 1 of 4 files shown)


topiary/src/atom_collection.rs
```diff
@@ -266,23 +269,25 @@ impl AtomCollection {
...
+                    expanded.append(prepends);
+                    expanded.push(swapped_atom);
+                    expanded.append(appends);
                 }
                 _ => {
-                    log::debug!("Not a leaf: {atom:?}");
-                    expanded.push(atom.clone());
+                    let mut swapped_atom = Atom::Empty;
+                    mem::swap(&mut swapped_atom, atom);
+                    expanded.push(swapped_atom);
```
Xophmeister, line 290: This swapping business is a bit harder to understand as a Rust newbie. AIUI, it's similar to what `Option::take` is doing to satisfy ownership. If it's an idiomatic thing to do -- which it probably is -- then I think it's fine; if not, _maybe_ it deserves a bit more comment.

(Incidentally, you removed the `log::debug!`. Was that intentional?

topiary/src/atom_collection.rs
```diff
@@ -519,7 +524,7 @@ impl AtomCollection {
                                     let multiline_only =
                                         *condition == ScopeCondition::MultiLineOnly;
                                     let new_atom = if multiline == multiline_only {
-                                        Some((**atom).clone())
+                                        Some(atom.deref().clone())
```
Xophmeister, line 527: Out of curiosity, is this just syntactic sugar? (Not that that's a bad thing :slightly_smiling_face:)

Xophmeister (commented):
The overall theme of how the in-place processing is accomplished, in light of Rust's ownership model, comes through quite clearly. (If I understand correctly!) The code is harder to understand, but not detrimentally -- the post-processing refactor was the hardest to follow -- and I think that could be easily improved by adding comments explaining what's going on.

I presume the in-place processing is faster, given how much less allocation and copying is happening. It would be interesting to see before-and-after benchmarks.

aspiwack:
Do we observe improvement in processing time or memory?

torhovland:
Our benchmark is showing a 2% improvement in performance. 

Valgrind is showing 83,914,675 bytes allocated in 255,611 allocations during formatting of the OCaml sample file in a release build before the change, and 77,448,106 bytes in 208,046 allocations after the change.

Not anything to write home about. This is more about idiomatic code than pure performance.

topiary/src/atom_collection.rs
torhovland: Added more explanation. And added back the logging.

PR #392, opened 2023-03-27
Test that sample files trigger every pattern in the query

nbacquey:
to do:
- [x] Add content to non-exhaustive sample files or remove useless patterns
  - [x] In Rust
  - [x] In Bash
  - [x] In OCaml 
- [x] Document the process to check exhaustivity
- [x] Point again to the `head` of `tree-sitter-facade` once https://github.com/tweag/tree-sitter-facade/pull/2 has been merged
- [x] Figure out why the tests pass on CI. They should fail, as there are patterns that are not matched by the rust sample file, at least (should be alright when #397 is merged).

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -609,20 +609,6 @@
   [
     (comment)
     (if_expression)
-    "%"
-  ]? @do_nothing
-)
-
-(
-  "else"
-  .
-  "%"
-  .
-  (attribute_id) @append_spaced_softline
-  .
-  [
-    (comment)
-    (if_expression)
   ]? @do_nothing
 )
 
@@ -1183,21 +1169,6 @@
   "->" @append_indent_start
   (_) @append_indent_end
   .
-)
-
-; Make an indented block where a function type arrow starts. Only for the root
-; level, not for each arrow.
-;
-; (?used_slot:bool ref ->
-;   Longident.t loc ->
-;   Path.t * Env.t)
-;
-(constructed_type
-  (function_type
-    "->" @append_indent_start
-    (_) @append_indent_end
-    .
-  )
 )
 
 ; Make an indented block where a function/match starts in PPX syntax.
--- a/topiary/src/tree_sitter.rs
+++ b/topiary/src/tree_sitter.rs
@@ -329,19 +329,19 @@ fn check_input_exhaustivity(
     root: &Node,
     source: &[u8],
 ) -> FormatterResult<()> {
-    let pattern_count = original_query.inner.pattern_count();
+    let pattern_count = original_query.pattern_count();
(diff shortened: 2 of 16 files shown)


topiary/src/tree_sitter.rs
```diff
@@ -84,6 +84,7 @@ pub fn apply_query(
     input_content: &str,
     query_content: &str,
     grammars: &[tree_sitter_facade::Language],
+    should_check_input_exhaustivity: bool,
```
Xophmeister, line 87: I wonder if it's possible to set this _only_ in the context of tests, so you don't have to pass it through as a parameter everywhere. Maybe there's a conditional compilation pragma that does this...

topiary/tests/sample-tester.rs
```diff
@@ -84,3 +85,47 @@ async fn formatted_query_tester() {
...
+        )
+        .unwrap_or_else(|e| {
+            if let FormatterError::PatternDoesNotMatch(_) = e {
+                panic!("Found untested query in file {query_file:?}:\n{e}");
+            } else {
+                panic!("{e}");
+            }
+        })
+    }
+}
```
Xophmeister, line 131: Missing EOF `\n`

Xophmeister (approved):
This looks good :+1:

As per one of my comments, I wonder if it's possible to trigger the exhaustiveness testing with a conditional compilation pragma. I don't know if that's possible and, if it is, whether it can be isolated to specific tests (e.g., you don't want the exhaustiveness testing to run on the input/expected integration test, because it'll take too long)...

languages/ocaml.scm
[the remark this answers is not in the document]
nbacquey: The pattern that matched on `else%` was removed, because `else%` is not a valid OCaml construction, so that pattern could never match.

The other pattern, about function arrows, is a bit different. I'm not sure what exactly it was supposed to achieve, and the problem is that the pattern does not match the example code snippet above. My guess is that the pattern was an early try at indenting type arrows that never really worked.
In any case, indentation of types is taken care of elsewhere, so I think it's safe to remove the pattern altogether.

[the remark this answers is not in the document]
nbacquey: For `rust.scm`, those were patterns that were not used in the sample file, but that would probably match in other files.
In any case, those patterns were only convoluted ways to add spaces in specific cases. I think we would have had to rewrite them anyway if we wanted to properly support Rust.

topiary/src/tree_sitter.rs
nbacquey: I think the last version is much more reasonable: I expose `apply_query` for testing, instead of passing a boolean flag everywhere.

Issue #390, opened 2023-03-27
Integration tests fail if errant files exist in the sample directories


Xophmeister:
**Describe the bug**
The integration tests crash if a file exists in `topiary/tests/samples/input` that it doesn't understand. This can happen if, for example, you are editing a test and your editor places a swap/undo file alongside it.

**To Reproduce**
With Vim:
* Open any file in `topiary/tests/samples/input`. Under normal configuration, Vim will create a swap file alongside it (e.g., `.whatever.swp`).
* Run the integration test suite.

This probably happens with other editors (e.g., I believe Emacs creates `whatever~` sibling files under normal configuration).

**Expected behavior**
Skip over editor artefacts, or unknown files. Although that then raises the question: What's good and what's not? Perhaps only matching files in `topiary/tests/samples/expected` would be a first start...although if you were editing both simultaneously -- which is likely -- the same problem may occur!

**Environment**
 - OS name + version: NixOS 22.05
 - Version of the code: a26a870

**Additional context**
In Vim, you can work around the problem by using the `-n` switch to not create swap files. (Or, Vim can be configured to keep swap files elsewhere.) Presumably other editors have similar facilities.

amanx123:
I would like to work on this, please assign me !

Xophmeister:
Thanks for your interest, @amanx123 

The Topiary team haven't prioritised this, so please feel free to address it. We look forward to your PR :slightly_smiling_face:

PR #394, opened 2023-03-28
Automated testing of web playground.

### LINKED ISSUE #365: Automated testing of Wasm playground
As grammars and Topiary code evolves separately, we are bound to break the playground at some point. We need automated tests to cover this.

Either unit tests against the Wasm code, or GUI tests (e.g. Cypress) against the web playground, or both.

torhovland:
Fixes #365.


.github/workflows/ci.yml
```diff
@@ -41,7 +47,15 @@ jobs:
...
         run: npm install --prefix web-playground
 
-      - name: Build web playground frontend
+      - name: Start web playground frontend
+        if: success() && matrix.os == 'ubuntu-latest'
+        run: npm start --prefix web-playground &
+
+      - name: Test web playground frontend
+        if: success() && matrix.os == 'ubuntu-latest'
+        run: npm run e2e --prefix web-playground
```
Xophmeister, line 56: Does the latter task depend on the first reaching a state where it's ready? Can it fail if the first is still pending?

web-playground/e2e/sample-tester.test.ts
```diff
@@ -0,0 +1,82 @@
...
+            console.log(`Testing ${inputPath} ${expectedPath} ${queryPath}`);
+
+            const encoding = "utf8";
+            const input = await fs.promises.readFile(inputPath, encoding);
+            const expected = await fs.promises.readFile(expectedPath, encoding);
+            const query = await fs.promises.readFile(queryPath, encoding);
+
+            await testInputFile(input, expected, query);
+        }
+    }, 20000);
```
Xophmeister, line 31: Is this `20000` a timeout, presumably in milliseconds? It might be worth adding it as a top-level constant, so it looks less like a magic number.

web-playground/e2e/sample-tester.test.ts
```diff
@@ -0,0 +1,82 @@
+import { FrameWaitForFunctionOptions, Page } from "puppeteer";
+import * as fs from 'fs';
+import * as path from 'path';
+
+describe('test all grammars with puppeteer', () => {
+    beforeEach(async () => {
+        await page.goto('http://localhost:3000/playground');
```
Xophmeister, line 7: Maybe this negates my first comment: The test will wait for the playground to be ready. What happens, though, if the playground is never ready (if it broke, say)? Does `beforeEach` have an implicit timeout?

web-playground/tsconfig.json
```diff
@@ -18,9 +18,17 @@
...
+      "puppeteer",
+      "jest-environment-puppeteer",
+      "expect-puppeteer"
+    ]
   },
   "include": [
     "src"
   ]
-}
+}
```
Xophmeister, line 34: No EOF `\n` (FWIW)

Xophmeister (approved):
I'm not an expert in this ecosystem, but the code is readable and I can certainly follow what it's trying to do. To which end, it looks great :+1:

I've left a few comments, but no blockers (I think). I'll approve, but with the caveat that it's probably worth having a second opinion!

.github/workflows/ci.yml
torhovland: No, I think it's up and running when the command ends.

Xophmeister: Cool :+1: Fair enough. My worry was that sending it to the background with `&` wouldn't guarantee that.

web-playground/e2e/sample-tester.test.ts
torhovland: Then the test quickly fails with `net::ERR_CONNECTION_REFUSED at http://localhost:3000/playground`.

web-playground/tsconfig.json
torhovland: Ha ha, I saw that and half expected a review comment 🙂

torhovland: Apparently I have a formatter that does that on JSON files!

torhovland: Fixed.

web-playground/e2e/sample-tester.test.ts
torhovland: Fixed.

.github/workflows/ci.yml
torhovland: Oh, right, I see now. Yes, you are right in principle. In practice, it seems to work. Do you want to wait and see if it's flaky, or do you know a good way to improve it now?

torhovland: Another way, which is either genius or super dirty, is to defer the testing until after the following build step.

Xophmeister: I think that's both super dirty and genius simultaneously :smile: A `curl` loop is probably safest and expresses the intention clearly... Let me add that :+1:

torhovland: Good!

Now that I think about it, it might be best to move the frontend build step ahead of all of this. So if there actually is something wrong with the app, we get an error about that and not just about the app refusing to start.

Xophmeister: I added the `curl` check in f884d2b

Xophmeister: Yep :disappointed: I don't know why. I've turned off the redirect to `/dev/null` to see if that reveals anything.

torhovland: My machines say this:

```
curl: option --connect-timeout=10: is unknown
```

Xophmeister: Yep, that's the problem. Apparently `curl` doesn't like long options with equal signs. #TIL

torhovland: Interesting. This one doesn't wait at all:

```bash
curl --connect-timeout 10 http://localhost:3000
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Connection refused
```

Xophmeister: I'm just trawling through the manual. I think it should be `--max-time` instead...

torhovland: That one also fails right away. Apparently these two are only max limits, not min.

Xophmeister: ...Sigh. Nope, that doesn't work either.

Xophmeister: I was hoping `curl` had a nicer way, but it looks like the loop is just going to be easier :+1:

torhovland: If anything, I should be apologising for dragging you into this 🙂

But this is definitely an improvement. Now we get to see the full output of the start command, so it will be easy to see if it has failed, and why.

torhovland:
> I've left a few comments, but no blockers (I think). I'll approve, but with the caveat that it's probably worth having a second opinion!

@ErinvanderVeen @nbacquey

web-playground/e2e/sample-tester.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    it('can format', async () => {
+        const rootDir = path.join(__dirname, "../../");
+        const inputDir = path.join(rootDir, "topiary/tests/samples/input/");
+        const expectedDir = path.join(rootDir, "topiary/tests/samples/expected/");
+        const queryDir = path.join(rootDir, "languages/");
+
+        for (let inputFileName of await fs.promises.readdir(inputDir)) {
```
nbacquey, line 18: We have a pending issue about tests breaking if there are unexpected files in the `input`/`expected` directories (#390).
Maybe we should mark this code location to be considered as well when this issue is addressed?

nbacquey (approved):

PR #396, opened 2023-03-28
Nickel: Open an indent block between multi-line parentheses

Xophmeister:
This PR makes a small tweak to the formatting of parentheses. Namely, if their contents are multi-line, then that starts an indentation block on a new-line.

### DIFF
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -86,11 +86,22 @@
   ";"
 ] @prepend_antispace
 
+; Create a scope between parentheses such that, if their contents are
+; multi-line, we start an indentation block on a new line
+(
+  (#scope_id! "parens")
+  "(" @begin_scope
+  ")" @end_scope
+)
+
 ; Don't insert spaces immediately inside parentheses
 ; WARNING Using parentheses in string interpolation will manifest the
 ; bug documented in Issue #395
-"(" @append_antispace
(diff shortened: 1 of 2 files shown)


languages/nickel.scm
```diff
@@ -86,11 +86,22 @@
   ";"
 ] @prepend_antispace
 
+; Create a scope between parentheses such that, if their contents are
+; multi-line, we start an indentation block on a new line
+(
+  (#scope_id! "parens")
+  "(" @begin_scope
+  ")" @end_scope
```
nbacquey, line 94: I'm not sure about the use of scopes here: If you have the following
```nickel
(foo)
(bar)
(baz)
```
Then either one of those can happen:
- Each parenthesized expression is in its own node, and there is no need for scopes (you can remove the scope and only use your second rule, and everything should be fine).
- All six parentheses are children of the same node, and your expression will match on all six pairs of opening/closing parentheses, which is not something you want.

nbacquey (changes requested):

languages/nickel.scm
Xophmeister: Good point :+1: I was getting a bit scope-happy, but in this case I can simplify...

Xophmeister: Simplified in bc22a65175ed

PR #397, opened 2023-03-28
Re-enable tests of topiary lib in CI

nbacquey:


Xophmeister (approved):
I'm not a Nix expert -- so don't trust my judgement :sweat_smile: -- but it looks good to me :+1:

PR #402, opened 2023-03-29
Nickel: else..if chains

### LINKED ISSUE #399: Nickel: else..if formatting
In a chain of `if..then..else` expressions, it would be better _not_ to break the line and increase the indent level at each expression. That is, rather than:

```nickel
if ... then
  ...
else
  if ... then
    ...
  else
    if ... then
      ...
    else
      ...
```

We'd have:

```nickel
if ... then
  ...
else if ... then
  ...
else if ... then
  ...
else
  ...
```

([Discussion](https://github.com/tweag/nickel/pull/1198#discussion_r1150806010))

Xophmeister:
Resolves #399

### DIFF
--- a/languages/nickel.scm
+++ b/languages/nickel.scm
@@ -277,18 +277,36 @@
 ; if...then...else expressions can either be single or multi-line. In a
 ; multi-line context, they will be formatted like so:
 ;
-;  if CONDITION then
-;    TRUE_TERM
-;  else
-;    FALSE_TERM
+;   if CONDITION then
+;     TRUE_TERM
+;   else
+;     FALSE_TERM
+;
+; NOTE If the FALSE_TERM is another if...then...else expression, in a
+; multi-line context, then the indentation block is "cancelled" to
+; create the illusion of an "else if" term:
+;
+;   if CONDITION1 then
+;     TRUE_TERM1
+;   else if CONDITION2 then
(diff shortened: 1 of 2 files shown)


languages/nickel.scm
```diff
@@ -277,18 +277,36 @@
...
 (ite_expr
   "then" @append_spaced_softline @append_indent_start
-  t1: (term) @append_indent_end
-  "else" @prepend_spaced_softline @append_spaced_softline @append_indent_start
-  t2: (term) @append_indent_end
+  t1: (term) @append_indent_end @append_spaced_softline
+)
+
+(ite_expr
+  "else" @append_spaced_softline @append_indent_start
```
nbacquey, line 304: It's not a blocker, but it's interesting that this works: the `append_spaced_softline` also gets cancelled by `do_nothing`, when the the second term is `ite_expr`, yet the spacing is correct on the samples.
It means that:
- there is another rule that inserts a space between `else` and `if`,
- the queries should work with `append_empty_softline`.

nbacquey (approved):

languages/nickel.scm
Xophmeister: `if` and `else` are both in the "surround spaces" alternation...although that was more luck than judgement on my part. Good catch!

PR #404, opened 2023-03-30
Don't process queries that match below a leaf node

nbacquey:
This PR can be tested by adding the following to `bash.scm`:
```scheme
(
  (variable_name) @append_delimiter
  (#delimiter! ".")
)
```
Then running the following:
```bash
$ echo 'echo $toto' | RUST_LOG=debug topiary -sq languages/bash.scm |& grep WARN
[2023-03-30T15:42:58Z WARN  topiary::atom_collection] Aborting because the match occurred below a leaf node: {Node variable_name (0, 6) - (0, 10)}
```


topiary/src/atom_collection.rs
```diff
@@ -120,6 +120,12 @@ impl AtomCollection {
             log::debug!("Aborting because context is single-line and #multi_line_only! is set");
             return Ok(());
         }
+        if let Some(parent_id) = self.parent_leaf_nodes.get(&node.id()) {
+            if *parent_id != node.id() {
+                log::warn!("Aborting because the match occurred below a leaf node: {node:?}");
```
Xophmeister, line 125: This also applies to the other warnings above this one, but "abort" is a bit strong. If I saw that in the logs, I might think that the whole program had halted, but that's not the case. Maybe "skip" would be less alarmist.

Xophmeister (approved):
Looks great :100:

topiary/src/atom_collection.rs
nbacquey: Replaced with `Skipping` in the latest version

Issue #405, opened 2023-03-31
Broken pre-commit-hooks integration in the flake


Niols:
**Describe the bug**

The pre-commit-hooks integration as provided by the flake is broken. This is possibly due to a recent change in the Nix integration for Cargo.

**To Reproduce**

```console
$ nix eval github:tweag/topiary/1af03d7#lib.x86_64-linux.pre-commit-hook
error: attribute 'app' missing

       at /nix/store/3pdg8mbmckf05kakcqk4yvk4pzwnll3f-source/flake.nix:58:26:

           57|                   for FILE; do
           58|                     if ${code.app}/bin/topiary --in-place --input-file "$FILE"; then
             |                          ^
           59|                       continue
(use '--show-trace' to show detailed location information)
```

**Expected behavior**

The evaluation should yield a working pre-commit-hooks configuration. For instance, when it was introduced in fd80715:

```console
$ nix eval github:tweag/topiary/fd80715#lib.x86_64-linux.pre-commit-hook
{ description = "A general code formatter based on tree-sitter."; enable = true; entry = "/nix/store/vdjw8qch0r3pw4qs3mhy85ak0bv621ry-topiary-inplace/bin/topiary-inplace"; files = "(\\.json$)|(\\.toml$)|(\\.mli?$)"; name = "topiary"; }
```

torhovland:
OK, we need to add a test for this.

PR #409, opened 2023-04-03
Fix pre-commit-hook settings and add a flake check for them

Niols:
closes https://github.com/tweag/topiary/issues/405

This PR:

- fixes the bug described in #405 
- adds a flake check to avoid future similar breakages.

As an evidence that this works, I will push twice, once without the fix and once with it. The CI should complain about the first commit, and be happy with the second.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,7 +44,10 @@
         checks = {
           inherit (code) clippy clippy-wasm fmt topiary-lib topiary-cli topiary-playground audit benchmark;
 
-          ## Check that the `lib.pre-commit-hook` output builds correctly.
+          ## Check that the `lib.pre-commit-hook` output builds/evaluates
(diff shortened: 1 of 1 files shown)


nbacquey (approved):
Although my nix-fu is too weak, I reckon that the the procedure you describe proves that this PR indeed fixes something :sweat_smile:

Niols:
My nix-fu was also too weak for this. The `deepSeq` + `hello` trick comes from @thufschmitt. Maybe the line
```nix
pre-commit-hook = builtins.deepSeq self.lib.${system}.pre-commit-hook pkgs.hello;
```
deserves some more comments? I'd happily add that.

Niols:
WDYT of 210eb17?

nbacquey:
Looks good to me ; you can merge at your convenience

Xophmeister (approved):

PR #410, opened 2023-04-03
Skip over zero-byte nodes

### LINKED ISSUE #383: [Tree-sitter query] Syntax tree for predicates always have a `parameters` node
**Describe the bug**
Tree-sitter query predicates follow the syntax `(#foo! bar)`, where `foo` is the predicate name, and `bar` is an optional parameter. The problem is that the syntax tree of a predicate _always_ contains a parameter, even if the original predicate does not.
This causes spacing issues:

This is the syntax tree of `(#foo! bar)`:
```
{Node program (0, 0) - (1, 0)}
  {Node predicate (0, 0) - (0, 11)}
    {Node ( (0, 0) - (0, 1)}
    {Node # (0, 1) - (0, 2)}
    {Node identifier (0, 2) - (0, 5)}
    {Node predicate_type (0, 5) - (0, 6)}
    {Node parameters (0, 7) - (0, 10)}
      {Node identifier (0, 7) - (0, 10)}
    {Node ) (0, 10) - (0, 11)}
```

This is the syntax tree of `(#foo!)`, notice the empty `parameters` node:
```
{Node program (0, 0) - (1, 0)}
  {Node predicate (0, 0) - (0, 8)}
    {Node ( (0, 0) - (0, 1)}
    {Node # (0, 1) - (0, 2)}
    {Node identifier (0, 2) - (0, 5)}
    {Node predicate_type (0, 5) - (0, 6)}
    {Node parameters (0, 6) - (0, 6)}
      {Node identifier (0, 6) - (0, 6)}
    {Node ) (0, 7) - (0, 8)}
```
Because we put a space between `predicate_type` and `parameters`, this causes `(#foo!)` to be formatted as `(#foo! )`. 

**To Reproduce**
```bash
$ echo '(#foo!)' | topiary -q languages/tree-sitter-query.scm
(#foo! )
```

**Expected behavior**
`(#foo!)` should be formatted as is.

Xophmeister:
This PR checks the byte-width of a node, when constructing the atom collection, to skip zero-width nodes. Along with a slight tweak to the Tree-Sitter Query formatting rules, this allows us to format parameter-less predicates without an undesired space.

(n.b., It would have been nicer to filter out zero-width nodes during the parse, but Tree-Sitter doesn't expose an API to clip nodes from the tree directly.)

Resolves #383

### DIFF
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -327,7 +327,9 @@ impl AtomCollection {
             node.is_named()
         );
 
-        if node.child_count() == 0 || self.specified_leaf_nodes.contains(&node.id()) {
+        if node.end_byte() == node.start_byte() {
+            log::warn!("Skipping zero-byte node: {node:?}");
+        } else if node.child_count() == 0 || self.specified_leaf_nodes.contains(&node.id()) {
             self.atoms.push(Atom::Leaf {
                 content: String::from(node.utf8_text(source)?),
                 id,
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -679,7 +679,7 @@
   "|" @delete
   .
   (match_case)
-  (#single_line_only! )
+  (#single_line_only!)
 )
 
 (
@@ -689,7 +689,7 @@
   .
   (match_case) @prepend_delimiter
   (#delimiter! "| ") ; sic
-  (#multi_line_only! )
+  (#multi_line_only!)
 )
 
 ; Multi-line definitions must have a linebreak after "=" and before "in":
--- a/languages/tree-sitter-query.scm
+++ b/languages/tree-sitter-query.scm
@@ -26,7 +26,6 @@
     (grouping)
     (named_node)
     (parameters)
-    (predicate_type)
     (quantifier)
     ":"
     "."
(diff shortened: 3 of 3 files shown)


topiary/src/atom_collection.rs
```diff
@@ -327,7 +327,9 @@ impl AtomCollection {
             node.is_named()
         );
 
-        if node.child_count() == 0 || self.specified_leaf_nodes.contains(&node.id()) {
+        if node.end_byte() == node.start_byte() {
+            log::warn!("Skipping zero-byte node: {node:?}");
```
nbacquey, line 331: I don't know if I would put `warn` here: in my opinion, `warn` should inform the user that something wrong (but not fatal) is happening with their input, or the query file, and should/could be fixed by them.
Here the error is entirely on topiary's side, the user cannot do anything about it. I would replace it by `info` or `debug`

nbacquey (approved):

Xophmeister:
Note: This fixes the problem of a query matching a zero-byte node, then applying formatting that it shouldn't. It _doesn't_ fix the problem of a query failing to match because a zero-byte node exists in the tree which the query doesn't account for (see #420).

PR #411, opened 2023-04-03
Use info logging, instead of warn, for zero-width nodes

Xophmeister:
Per https://github.com/tweag/topiary/pull/410#pullrequestreview-1369233669


nbacquey (approved):

aspiwack:
Can I be a pain? This sounds like a very verbose piece of output, and I would relegate it to the debug level.

PR #413, opened 2023-04-04
Reduce verbosity of failed tests

### LINKED ISSUE #401: Only show the differences when an integration test fails
When an integration test fails -- in the case of the output not matching the expected output -- then the full, colourised diff is printed. Given that the integration test input files are _very_ large, but the difference can be isolated to single lines, it can be difficult to page through the output to find the formatting errors.

Only output the changes, not the code that is invariant under formatting.

n/a

**Additional context**
n/a

nbacquey:
This new version uses `prettydif::text::diff_lines`, which is less generic than `pretty_assertions::assert_eq`, but was more easily modified, and suits our purpose well.

Currently, this is the output of a failed test with this PR:
image

I will edit https://github.com/romankoblov/prettydiff/pull/14 to add a skipping symbol (_e.g._ `[...]`) and line numbers.
Closes #401

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,10 +18,53 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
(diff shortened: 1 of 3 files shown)


nbacquey:
This is what the new version looks like:
image

torhovland:
So cool! And great how this got merged and released in `prettydiff`!

torhovland (approved):

PR #414, opened 2023-04-04
Rewrite detect_line_breaks_inner

Xophmeister:
Resolves #346

Note: The flattening of the syntax tree can be factored out, to avoid having to do it twice (for single line breaks and blank lines). I will do this refactor in a separate PR.

PR #417, opened 2023-04-06
Further refactoring of topiary::atom_collection

Xophmeister:
* [x] Factored out DFS flattening function, to avoid walking the tree multiple times.
* [x] Applied @torhovland's suggestions from #414.
  * [x] Renamed nodes in closure, for clarity.
  * [x] Use an output struct, for clarity.
  * [ ] ~Convert DFS flattening function into a recursive function, rather than a loop.~

### DIFF
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -9,6 +9,11 @@ use tree_sitter_facade::Node;
 
 use crate::{Atom, FormatterError, FormatterResult, ScopeCondition};
 
+struct NodesWithLinebreaks {
+    before: HashSet<usize>,
+    after: HashSet<usize>,
+}
+
 #[derive(Debug)]
 pub struct AtomCollection {
     atoms: Vec<Atom>,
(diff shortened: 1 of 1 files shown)


Xophmeister:
In a not completely sound scientific analysis,[^method] the recursive DFS runs at over twice the runtime of the loop version.[^results]

Loop version:
```rust
fn dfs_flatten<'tree>(node: &'tree Node<'tree>) -> Vec<Node> {
    let mut walker = node.walk();
    let mut dfs_nodes = Vec::new();

    'walk: loop {
        dfs_nodes.push(walker.node());

        if !walker.goto_first_child() {
            while !walker.goto_next_sibling() {
                if !walker.goto_parent() {
                    break 'walk;
                }
            }
        }
    }

    dfs_nodes
}
```

(Naive) recursive version:
```rust
fn dfs_flatten(node: Node) -> Vec<Node> {
    let mut dfs_nodes = vec![node];

    for child in node.children(&mut node.walk()) {
        dfs_nodes.append(&mut dfs_flatten(child))
    }

    dfs_nodes
}
```

I suspect this is for several reasons:
* We're doing a lot of allocation in the recursive version. The loop version will reallocate every time its vector reaches capacity, but with the $2^n$ scaling, that will almost certainly be significantly less frequently than every time the recursive version descends into a child (and then some).
* To satisfy ownership/borrowing, the recursive version uses concrete `Node` types, rather than simple references. This increases the amount of data needed to copy around at each point. (Plus a new stack frame on every recursion, but maybe Rust/LLVM optimises that out.)
* It's supposedly more efficient to use a single walker, but in our recursive version, we recreate a walker for each child.

It's probably be possible to overcome any/all of these limitations. It's not immediately obvious how, however, and doing so will certainly complicate the code; obviating the point of the recursive refactor. IMO, I'm not convinced that either is clearer to understand than the other, in their above forms. Therefore, I'm not including that change in this PR; beyond a comment in the code back to this discussion.

[^method]: Average execution time, in the debug build -- the release build optimises out the calls -- over 10,000 trials, taking the Topiary OCaml integration test source as its input.

[^results]: Loop: 1.86ms, Recursive: 4.21ms (226.3%)

nbacquey (approved):

PR #423, opened 2023-04-12
Nickel: Consistent formatting for bound expressions

### LINKED ISSUE #389: Final niggles in the Nickel formatting rules
As of a58d922d9309, there are a few outstanding niggles with the Nickel formatting rules, around `let` expressions that involve annotations. (Modulo feedback from the Nickel team on the style rendered by these rules against the stdlib.) In particular:

* Idempotency fails for multi-line `let` expressions with two (or more) annotations on the same line as the equals sign. For example
  ```nickel
  let x
    : TYPE | ANNOT = 1 in x
  ```
  I would expect -- although, see the final point -- this to be formatted as:
  ```nickel
  let x
    : TYPE
    | ANNOT
    = 1 in x
  ```
  Interestingly, this is what the second pass results in.

* The rules for annotated, multi-line `let` expressions doesn't push the equals sign on to a new line, like how record field formatting works. For example:
  ```nickel
  let x
    : TYPE = 1 in x
  ```
  ...should (again, see final point) be formatted as:
  ```nickel
  let x
    : TYPE
    = 1 in x
  ```

* The infamous final point regards the placing of the `in <EXPR>` term when annotations are present, in a multi-line `let` expression. Without annotations, the `in <EXPR>` is de-indented on a new line, with the `<EXPR>` going on to a new, indented line _if_ it's multi-line. With annotations, you can see from above that's not the case (errors notwithstanding). That is, what I ultimately want -- to be consistent with annotated record fields -- in those two cases would be:
  ```nickel
  let x
    : TYPE
    | ANNOT
    = 1
  in x

  let x
    : TYPE
    = 1
  in x
  ```

**Note** The idempotency error does _not_ manifest itself in the Nickel standard library. The other points also appear to be quite rare occurrences. Nonetheless, it would be nice to have these working as expected.

### LINKED ISSUE #422: Consistent formatting of bound expressions

There are mainly two ways to bind expression to names in Nickel: let-binding and record definitions. Of course, the two are different and used in different contexts, but they are morally part of the same schema, and share (at least partially) common syntax, in particular around annotations.

However, they are formatted differently:

```nickel
let foo = {
    bar = 1,
    baz = 2,
} in
{
  foo = {
    bar = 1,
    baz = 2,
  }
}
```

is reformatted to

```nickel
let foo =
  {
    bar = 1,
    baz = 2,
  }
in
{
  foo = {
    bar = 1,
    baz = 2,
  }
}
```

We can see that a multiline let-expression _always introduce a new line and identation level_ for the bound expression (the record inside `foo`), but a record field doesn't. There was a lengthy conversation with @Xophmeister, and one reason for the current behavior is that this avoid adding to many indentation levels when defining record fields: when coupled with metadata (which already introduces a level of indentation for the `=` sign), we can indeed quickly go far to the right when defining nested records.

The question of "should we add a new line or not" after a binding is an interesting one (and isn't clear-cut), but whatever we choose, I think the formatting of let-bound expressions and fields should be consistent and follow the same rules, as there doesn't seem to be a good reason why an argument for one style or the other would only apply to let-bindings and not record fields, or vice-versa.

After a more in-depth discussion with the Nickel team, we identified that the default should be to add a new line and a new indentation level (that is, following the current style of let-bindings). **However**, with the following exceptions, which are very common idioms and could inflate the indentation of the whole file unduly :

- if the next element is a record literal `{ .. }`, then don't add a new line (so, in the end, the previous example would in fact be formatted in the style of the the record field, and not the let-binding)
- if the child element is a function declaration (it's customary in ML to even have the argument to the left of the equal sign: while this syntax isn't valid in Nickel, it's still customary to write `let foo = fun x y z => ...`)
- if the child element is an array literal
- if the child element is a match statement
- if the child element is a expression inside parentheses

Xophmeister:
Resolves #422 

(Also resolves #389 "accidentally")

### DIFF
--- a/topiary/tests/samples/expected/nickel.ncl
+++ b/topiary/tests/samples/expected/nickel.ncl
@@ -211,13 +211,12 @@
         if length == 0 then
           acc
         else
-          let rec go =
-            fun acc n =>
-              if n == length then
-                acc
-              else
-                let next_acc = %elem_at% l n |> f acc in
-                %seq% next_acc (go next_acc (n + 1))
+          let rec go = fun acc n =>
+            if n == length then
+              acc
+            else
+              let next_acc = %elem_at% l n |> f acc in
+              %seq% next_acc (go next_acc (n + 1))
           in
           go acc 0,
 
@@ -247,13 +246,12 @@
         "%
       = fun f fst l =>
         let length = %length% l in
-        let rec go =
-          fun n =>
-            if n == length then
-              fst
-            else
-              go (n + 1)
-              |> f (%elem_at% l n)
+        let rec go = fun n =>
+          if n == length then
+            fst
+          else
+            go (n + 1)
+            |> f (%elem_at% l n)
         in go 0,
 
     prepend
@@ -377,12 +375,11 @@
         ```
         "%
       = fun pred l =>
-        let aux =
-          fun acc x =>
-            if (pred x) then
-              { right = acc.right @ [x], wrong = acc.wrong }
-            else
-              { right = acc.right, wrong = acc.wrong @ [x] }
+        let aux = fun acc x =>
+          if (pred x) then
+            { right = acc.right @ [x], wrong = acc.wrong }
+          else
+            { right = acc.right, wrong = acc.wrong @ [x] }
         in
         fold_left aux { right = [], wrong = [] } l,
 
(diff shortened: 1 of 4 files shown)


Xophmeister:
@yannham I _think_ this is pretty close to what you are looking for :crossed_fingers: You can see the diff on the stdlib (which we use for integration testing) [here](https://github.com/tweag/topiary/pull/423/files#diff-6986669f19c1caad9af03e15df1e512864cc90f72071dee180866ea9dc1210e0).

topiary/tests/samples/expected/nickel.ncl
```diff
@@ -1372,37 +1369,35 @@
...
-                  let contract = field_contracts."%{field}" in
-                  let label = %go_field% field label in
-                  let val = value."%{field}" in
-                  %record_insert% field acc (%assume% contract label val)
-                else
-                  acc
-            )
-            {}
-            (%fields% value)
+        let fields_with_contracts = array.fold_left
```
yannham, line 1388: I'm wondering why a newline isn't inserted here? isn't the child an application?

yannham (commented):
Mostly LGTM, despite the comment. However, I wonder why no record fields have been modified? Is it because they all happen to already fit the new formatting rules, by chance (they are all function definitions, record literals, arrays, etc. ? Which wouldn't be out of the question)

topiary/tests/samples/expected/nickel.ncl
Xophmeister: I found this quite curious, too... I've just dug into it and I _think_ I know the reason. The query to determine the exceptions is not strict enough and will match any sibling rather than just the first-sibling. In this case, there are three exceptions -- two parentheticals and one record -- that the query matches and thus, incorrectly, deduces the RHS to be an exception.

Let me fix that :sweat_smile:

Xophmeister: ...Yep, that was definitely the problem. Now fixed: https://github.com/tweag/topiary/pull/423/commits/f894cf2a3ae88963dea75018812f00b857ed438a#diff-6986669f19c1caad9af03e15df1e512864cc90f72071dee180866ea9dc1210e0

yannham (approved):
Ah yes, now that application is fixed there's a change indeed on a record field (the `contract.from_predicate` application). Looks great to me :+1: !

PR #427, opened 2023-04-14
Use Topiary logos

Xophmeister:
This PR replaces the stock photo and the header text with logos, on the website.

I've used the text logo, rather than the banner logo, in the header, as the shadow under the tree doesn't really work on a dark background. It looks like this in my browser:

image


website/index.html
```diff
@@ -91,9 +91,9 @@ <h1>Topiary</h1>
                 >
                     <div class="col-10 col-sm-8 col-lg-6">
                         <img
-                            src="images/jr-harris-MROejEmXGGE-unsplash.jpg"
+                            src="images/topiary_logo.png"
                             class="d-block mx-lg-auto img-fluid"
-                            alt="Cover photo by JR Harris on Unsplash"
+                            alt="Topiary logo by TODO"
```
Xophmeister, line 96: @ErinvanderVeen Would Maatje like credit here; she definitely deserves it :+1: If so, what would be best written here, to replace the current `TODO`?

torhovland:
You are of course welcome to switch to a header with light background. But it works like this too.

Xophmeister:
> You are of course welcome to switch to a header with light background. But it works like this too.

Dark looked better than the stock Bootstrap light background, imo.

website/index.html
ErinvanderVeen: She would not like to be credited on the logo.

ErinvanderVeen (approved):
Approved for when the TODO is removed ;)

PR #432, opened 2023-04-17
Add more samples/queries in playground

### LINKED ISSUE #356: More sample queries and inputs on the Playground
It should be easy to select sample queries for the various languages we support, as well as sample input.

Note that it is not necessary for the user to select grammars. This happens indirectly through the `#language!` predicate in the query. It would be nice to support _any_ third-party grammar as well, but this is not viable, as each grammar has to be built as Wasm up-front.

Ideally we would use the actual queries and sample input in the source repo.

For example a CI step that copies these into the playground frontend source.

nbacquey:
This PR adds a ts export, which is generated when building the library, and then copied by the CI in the playground.
It contains all languages and inputs in the library, which makes the playground able to load them.

Closes #356


torhovland:
Looks like you've moved past that problem now.

nbacquey:
I don't think I have: CI still fails because
```
cp: cannot stat 'topiary-playground/languages_export.ts': No such file or directory
```

torhovland:
I think the problem here is that you are generating that file as a side effect in Nix. Which means that when Nix thinks it doesn't have to build the derivation, no file gets generated. You can demonstrate that like this:

```
nix build .#topiary-playground  
ls topiary-playground/languages_export.ts
rm topiary-playground/languages_export.ts
nix build .#topiary-playground  
ls topiary-playground/languages_export.ts
```

nbacquey:
When I try your code snippet on my machine, the first `nix build .#topiary-playground` does not build `topiary-playground/languages_export.ts`, even if I forcibly remove the existing derivation with `nix store delete --ignore-liveness`. Besides, shouldn't the CI always try to build the derivation?

torhovland:
If I force a Nix build by simply changing some source code, it rebuilds that file.

The CI will not build the derivation if it finds the hash it needs in Cachix.

nbacquey:
In the end, I chose the coward's way, and committed the file. It's "only" 8k lines, and it should have a diff only when then samples/inputs get modified. I'll open an issue to add it as a proper nix build target.

topiary/build.rs
```diff
@@ -0,0 +1,74 @@
...
+    write!(
+        js_export,
+        "\
+        // This file is automatically generated by Cargo.\n\
+        // It is not intended for manual editing.\n\
+        // To generate, please run `cargo build -p topiary`\n\
+        const languages: {{[index: string]: any}} = {{\n\
+          {buffer}\
+        }};\n\n\
+        export default languages;\n"
```
Xophmeister, line 71: Same here RE `r#"..."#` literal.

topiary/build.rs
```diff
@@ -0,0 +1,74 @@
...
+        // Escape "${" to prevent this.
+        .replace("${", "\\${")
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=../languages/");
+    println!("cargo:rerun-if-changed=./tests/samples/input/");
+
+    // Export test samples and queries as JS files
+    let language_dir = current_dir().unwrap().join("../languages/");
```
Xophmeister, line 24: I wonder if we should limit this to the "officially supported" languages? Or at least indicate in the playground whether a language is experimental or not.

web-playground/src/App.tsx
```diff
@@ -28,9 +35,20 @@ function App() {
...
+            setInput(languages[l].input);
+            setQuery(languages[l].query);
+        }
+    }
+
     return (
         <div className="App">
             <div className="header">
+                <select onChange={e => changeLanguage(e.target.value)}>
+                    <option value="">Choose a language</option>
```
Xophmeister, line 49:
```suggestion
                    <option value="">Choose an example language</option>
```

I've added "example" here to make it feel more playground-like. Otherwise, it comes across -- despite it not being true -- as a bit "immutable".

(I'm not convinced that "example" is the best word...)

Xophmeister (approved):
This looks good :+1:

I've added a few comments/suggestions, but they're not blockers and probably more like future enhancements. The only thing I don't fully understand is why `languages_export.ts` needs to be checked in, if the CI will build it? (Maybe that's the point you mentioned earlier: A build won't necessarily be triggered on each run.)

Xophmeister:
> In the end, I chose the coward's way, and committed the file. It's "only" 8k lines, and it should have a diff only when then samples/inputs get modified. I'll open an issue to add it as a proper nix build target.

Got it :ok_hand: What's 8k lines of TypeScript, after all :sweat_smile:

topiary/languages_export.ts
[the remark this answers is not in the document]
Xophmeister: [Explained here](https://github.com/tweag/topiary/pull/432#issuecomment-1513135482)

topiary/build.rs
```diff
@@ -0,0 +1,74 @@
...
+        let content = to_js_string(path);
+        input_map.insert(prefix, content);
+    }
+
+    let mut buffer = String::new();
+
+    for (key, query) in language_map.into_iter().sorted() {
+        if let Some(input) = input_map.get(&key) {
+            buffer.push_str(&format!(
+                "  \"{key}\": {{\n    query: `{query}`,\n    input: `{input}`,\n  }},\n"
```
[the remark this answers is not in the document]
nbacquey: Applied in new version

web-playground/src/App.tsx
nbacquey: I've used "reference language" in the new version, to suggest that modifications can be added

web-playground/src/App.tsx
```diff
@@ -28,9 +35,20 @@ function App() {
         }
     }
 
+    function changeLanguage(l: string) {
+        if (languages[l]) {
+            setInput(languages[l].input);
```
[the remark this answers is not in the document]
Xophmeister: Yeah, I agree: it's not something for this PR.

topiary/build.rs
nbacquey: Maybe we could warn the user, but I don't think it makes sense to limit to the officially supported languages: the playground is just that, a playground. People shouldn't rely on its output for anything

Xophmeister: I agree; limiting is too strict. Perhaps all that needs to be done is add "experimental" against the appropriate languages in the selector. (Again, I don't think this is needed for this PR.)

PR #437, opened 2023-04-19
Ignore errant files in `topiary/test/samples/input` directory

### LINKED ISSUE #390: Integration tests fail if errant files exist in the sample directories

nbacquey:
This PR checks whether files in the `input` directory end with one of the recognized file extensions defined in `languages.rs`, and discards them otherwise.

TODO:
* [x] Fix import in sample-tester.test.ts

Closes #390


nbacquey:
CI currently fails, because of this line in `sample-tester.test.ts`:
```typescript
import isExtensionKnown from "../src/wasm-app/topiary_playground.js"
```
Indeed, `topiary_playground.js` contains `export` instructions, which `jest` can't natively process:
```
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.
```
I'm not sure what the proper `jest`/`babel` config options are at the moment.

nbacquey:
I duplicated a bit of logic to solve the aforementioned problem: the whitelist of extensions has been copied in `sample-tester.test.ts`. It should be removed when we migrate to a separate config file.

Xophmeister (approved):
LGTM :+1:

I'll merge, in your absence.

Issue #441, opened 2023-04-24
Failed idempotent check (ocaml-implementation)


nojb:
The idempotent check fails with
```ocaml
let _ =
  let module Lift = Mlfi_types.TypEq.Lift (struct type 'a c = 'a S.t end) in ()
```
(tried with the current `main`).

nojb:
Another one:
```ocaml
class memoized = object(this)
  method __memo__ : __memo__ = __memo__
  initializer memoize_obj this
end
```

nojb:
Another:
```
let split_rows_warnings rows_warnings =
  List.concat_map fst rows_warnings |> List.filter_map (function | Error s -> failwith s | Ok x -> x)
```

Xophmeister:
Thanks for these. We'll look into them :+1:

PR #449, opened 2023-05-04
Ocaml: modules can be on one line

aspiwack:
This solves the first example from #441 .

Generally speaking, all the hard lines are suspicious (in Ocaml in particular, but probably in every language), but I haven't t investigated further (I got hit by this one yesterday, so got incentivised to solve it; the playground is amazing for this sort of debugging by the way).

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -80,7 +80,7 @@
     (module_definition)
     (module_type_definition)
     (type_definition)
-  ] @append_hardline
+  ] @append_spaced_softline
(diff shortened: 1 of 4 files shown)


torhovland (approved):

aspiwack:
Hah! It doesn't suffice for #441 because it's a slightly different example than the one I tried, let me add some more softlines…

aspiwack:
Some testing. More softlines. We should probably investigate all softlines though.

PR #452, opened 2023-05-09
Fix idempotency issues in OCaml

### LINKED ISSUE #441: Failed idempotent check (ocaml-implementation)

nbacquey:
* Add support for `initializer` keyword
* Fix post-processing of `DeleteBegin`/`DeleteEnd` sections

Closes #441

### DIFF
--- a/languages/ocaml.scm
+++ b/languages/ocaml.scm
@@ -40,6 +40,7 @@
 ; Allow blank line before
 [
   (class_definition)
+  (class_initializer)
   (class_type_definition)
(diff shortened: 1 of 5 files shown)


Niols (commented):
I can only speak for the OCaml sample but it looks good to me.

PR #460, opened 2023-05-10
Update tree-sitter-nickel to latest unstable

vkleen:
This adds support for the new `{_ | Contract}` dictionary contract syntax.


Xophmeister (approved):
LGTM :+1: Let me update the Nickel WASM...then it's good to go :slightly_smiling_face:

Issue #418, opened 2023-04-07
Incorrect formatting for recursively defined TS node definition


bbannier:
It seems that `topiary::format` incorrectly formats recursively defined tree-sitter nodes, e.g., given a grammar

```js
module.exports = grammar({
	name: 'foo',

	rules: {
		foo: $ => seq(/FOO/, optional(seq('::', $.foo)))
	}
})
```

and an input `FOO::FOO` formatting without idempotency check produces `::FOO` instead of `FOO::FOO`. Since this construct is not allowed by the grammar this subsequently fails the idempotency check.

I believe this is a bug or limitation in topiary, but would be happy to know if I use it incorrectly.

* * *
Reproducer:

```toml
# Cargo.toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
tokio = { version = "1.27.0", features = ["macros", "rt"] }
topiary = { git = "https://github.com/tweag/topiary.git", rev = "8f790fc6b2a989850da3a7af5c30ab4a7e30c471" }
tree-sitter-facade = { git = "https://github.com/tweag/tree-sitter-facade" }
tree-sitter = "0.20.10"

[build-dependencies]
cc = "1.0.79"
```
```js
// grammar.js
module.exports = grammar({
	name: 'foo',

	rules: {
		foo: $ => seq(/FOO/, optional(seq('::', $.foo)))
	}
})
```
```rust
// build.rs
use {
    cc::Build,
    std::{
        env::{self, current_dir},
        path::PathBuf,
        process::Command,
    },
};

fn main() {
    let grammar = current_dir().unwrap().join("grammar.js");
    println!("cargo:rerun-if-changed=grammar.js");

    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());

    // Run `tree-sitter generate`.
    assert!(Command::new("tree-sitter")
        .arg("generate")
        .arg(&grammar)
        .current_dir(&out_dir)
        .status()
        .expect(
            "failed to generate tree-sitter bindings, is tree-sitter CLI installed and in PATH?"
        )
        .success());

    // Compile tree-sitter C output.
    let src_dir = out_dir.join("src");

    Build::new()
        .file(src_dir.join("parser.c"))
        .include(out_dir.join("src"))
        .warnings(false)
        .compile("tree-sitter-spicy");
}
```
```rust
// src/main.rs
extern "C" {
    fn tree_sitter_foo() -> tree_sitter::Language;
}

[...]

Issue #276, opened 2023-02-09
Grievances with the tree-sitter query language


nbacquey:
This issue will only serve to discuss weaknesses of the tree-sitter query language. We will be able to address them separately if needed.

### We can't use the (apparent?) full set of tree-sitter queries features
The [documentation](https://tree-sitter.github.io/tree-sitter/using-parsers#predicates) states that we can use predicates of the form
```scheme
(
  (identifier) @constant
  (#match? @constant "^[A-Z][A-Z_]+")
)
```
However, we can't currently compile them in the Rust library.

TODO: develop each point and give examples

Xophmeister:
Great list :+1: Just to expand a little, from my experience:

* Negation is amongst the Tree-Sitter issues (https://github.com/tree-sitter/tree-sitter/issues/705), but has been dormant for 2+ years.
* Anecdotally, the Kleene star appears to ignore anonymous nodes. (That mightn't be the exact constraint; more data needed!)
* At a glance, the `#match?` and `#eq?` predicates feel like they could be useful. However, I wonder how Topiary would use them, as we're using capture names as formatting directives, rather than arbitrary labels.

Xophmeister:
Another one:
* I have found myself needing, a few times, to write queries of the form:

  ```scm
  (X
    Y
    ; etc.
  )
  ```

  Where `X` ranges over a few nodes (say, `a`, `b` and `c`) and `Y` is constant. It would be nice if it were possible to alternate over head nodes (e.g., `([a b c] Y)`), rather than rewrite the rule several times and/or using scopes.

nbacquey:
Updated the initial message, and added your use-case @Xophmeister

nbacquey:
Added another one. I'm thinking about committing this file to the repository itself, what do you think?

aspiwack:
I don't see why not. But on the other hand, what purpose would serve a file that gathers this information?

nbacquey:
The same as this issue, I think:
* Warn contributors about the known shortcomings of the query language.
* Give pointers on which contributions to `tree-sitter` would be appreciated.

Except that it would stay visible, and its modifications would be `git`-tracked, which makes following its evolution easier than tracking a particular issue.

nbacquey:
I suppose another way to achieve that would be to add a `query language` issue tag, and create one issue per point

Xophmeister:
This is more of a meta-issue, but I figured it's worth mentioning here as it's somewhat relevant:

[...]

PR #462, opened 2023-05-11
[OCaml] Generic formatting rules for infix operators

### LINKED ISSUE #461: Suspicious formatting with `@@` and `let+`
**Describe the bug**
The following snippet 
```ocaml
List.of_seq @@
  let+ (a, v) = Map.to_seq model in
  match v with
  | true -> f a
  | false -> g a
```

Formats to
```ocaml
List.of_seq
@@ let+ (a, v) = Map.to_seq model in
match v with
| true -> f a
| false -> g a
```

(checked in the current playground)

**Expected behavior**

The lack of indentation is very suspicious.

At the very least we should have something like

```ocaml
List.of_seq
@@ let+ (a, v) = Map.to_seq model in
  match v with
  | true -> f a
  | false -> g a
```

Though, honestly, I'd prefer if the original snippet was already formatted. Because the ability to introduce visually cheap blocks is what makes `@@` valuable.

Another family of example would be with `fun`:

```ocaml
myBlockIntroducingFunction @@ fun x ->
  something horrible onto x
```

This is also something that I'd hope would be the form computed by Topiary (but isn't).

nbacquey:
Add a soft-line and indentation after infix operators that do not receive a specific treatment (_i.e._ everything except `||` and `&&`).
Note that we cannot target the infix operator `@@` specifically, because it doesn't appear in the CST. Indeed, it is generated by a regular expression in the grammar (see #418 for details).

Closes #461


topiary/tests/samples/input/ocaml.ml
```diff
@@ -982,4 +982,8 @@ let _ =
 
 (* Slight variant of #441 *)
 let _ =
-  let module Foo = struct let f = 0 let g = 1 in ()
+  let module Foo = struct let f = 0 let g = 1 end in ()
+
+(* Indentation/newlines after non-specific infix operators *)
+let _ = foo bar @@ let () = qux in
+  baz
```
aspiwack, line 989: Can we have an example of the form

```ocaml
myBlockIntroducingFunction @@ fun x ->
  something horrible onto x
```

as well? So that we can at least know for ourselves how these are formatted?

aspiwack (commented):
I don't really understand what's going on on the technical side.

But from a pure intellectual point of view, this all feels rather backward. I'd say that `@@` and `>>=` are basically the only two operators that we want to display at the end of the line in multiline formatting (even `|>` would go in front).

The following are my intuitive expectations.

```ocaml
foo @@
  bar @@
    baz
```

```ocaml
(* We probably don't need this in a world of `let*` if it's not obvious *)
foo >>= fun x ->
bar >>= fun y ->
baz
```

```ocaml
foo
+ bar
+ baz
```

```ocaml
foo
|| bar
|| baz
```

```ocaml
foo
@ bar
@ baz
```

etc…

topiary/tests/samples/input/ocaml.ml
nbacquey: Sure, I will add that. It should be formatted as
```ocaml
myBlockIntroducingFunction @@
  fun x ->
    something horrible onto x
```

nbacquey:
An additional word on what happens on the technical side:
In the [OCaml grammar](https://github.com/tree-sitter/tree-sitter-ocaml/blob/master/ocaml/grammar.js#L1900), some infix operators are hardcoded, and others are defined with regular expressions or `token` functions. The former appear in the CST, while the latter do not.

For instance, this is the definition of the additive operators in the grammar:
```javascript
    _add_operator: $ => choice(
      '+', '-', '+.', '-.',
      token(choice(
        seq('+', repeat1(OP_CHAR)),
        seq('-', choice(repeat1(/[!$%&*+\-./:<=?@^|~]/), repeat2(OP_CHAR)))
      ))
    ),
```
`+`, `+.`, `-`, and `-.` are hardcoded, while another additive operator like `-%` isn't.

This is the CST of `1 + 2`:
```
{Node infix_expression (0, 0) - (0, 5)} - Named: true
  {Node number (0, 0) - (0, 1)} - Named: true
  {Node infix_operator (0, 2) - (0, 3)} - Named: true
    {Node + (0, 2) - (0, 3)} - Named: false
  {Node number (0, 4) - (0, 5)} - Named: true
```

This is the CST of `1 -% 2`:
```
{Node infix_expression (0, 0) - (0, 6)} - Named: true
  {Node number (0, 0) - (0, 1)} - Named: true
  {Node infix_operator (0, 2) - (0, 4)} - Named: true
  {Node number (0, 5) - (0, 6)} - Named: true
```
Note that there is no `{Node -% (0, 2) - (0, 4)} - Named: false`.
The same thing happens with `||` and `@@`: `||` is hardcoded in the grammar, while `@@` is not.
Therefore, the following tree-sitter query may have matches:
```scheme
(infix_operator
  "||" @do_something
)
```
While the following will never match anything, and isn't even a correct query:
```scheme
(infix_operator
  "@@" @do_something
)
```

Hence the workaround I used in this PR, and `@@` being considered the default case, while `||` and `&&` are the specific case.

aspiwack:
Very clear, thanks.

Out of curiosity: is there a technical limitation which prevents _all_ infix operators from having the unnamed node thing at the leaf?

nbacquey:
The technical limitation is the following: in tree-siter grammars, tokens that are captured through regular expressions never appear in the CST.
I discovered this behaviour while investigating #418, and I think it is a known feature, but couldn't find any trace of it in the documentation.

Given that there is an infinite number of infix operators in OCaml, they are better described with regexes, and some of them cannot appear in the CST.

aspiwack:
Fascinating. Is this behaviour documented in #276 ?

aspiwack (approved):
Good as far as I'm concerned, but I feel that in the current situation, there's no helping getting torn between two different default behaviours for infix operators.

nbacquey:
> Fascinating. Is this behaviour documented in #276 ?

It is not. I'm adding it right now

PR #463, opened 2023-05-16
Separate configuration in a `toml` file

nbacquey:


ErinvanderVeen (changes requested):
With these changes I believe we can remove the configuration parameters from the query files.

nbacquey:
> With these changes I believe we can remove the configuration parameters from the query files.

Indeed. It is done.

languages/bash.scm
```diff
@@ -1,6 +1,3 @@
-; Configuration
-(#language! bash)
-
```
aspiwack, line 3: I guess that I'm a little confused regarding how you connect a language and the corresponding query file in this new setup?

nbacquey: This links the language to the query file:
```rust
    fn try_from(language: &Language) -> FormatterResult<Self> {
        let basename = Self::from(match language.name.as_str() {
            "bash" => "bash",
            "json" => "json",
            "nickel" => "nickel",
            "ocaml" => "ocaml",
            "ocaml_interface" => "ocaml",
            "ocaml_implementation" => "ocaml",
            "rust" => "rust",
            "toml" => "toml",
            "tree_sitter_query" => "tree-sitter-query",
            name => return Err(FormatterError::UnsupportedLanguage(name.to_string())),
        })
        .with_extension("scm");
```
and this links the language to the grammar:
```rust
    pub async fn grammars(&self) -> FormatterResult<Vec<tree_sitter_facade::Language>> {
        Ok(match self.name.as_str() {
            "bash" => vec![tree_sitter_bash::language()],
            "json" => vec![tree_sitter_json::language()],
            "nickel" => vec![tree_sitter_nickel::language()],
            "ocaml" => vec![
                tree_sitter_ocaml::language_ocaml(),
                tree_sitter_ocaml::language_ocaml_interface(),
            ],
            "ocaml_implementation" => vec![tree_sitter_ocaml::language_ocaml()],
            "ocaml_interface" => vec![tree_sitter_ocaml::language_ocaml_interface()],
            "rust" => vec![tree_sitter_rust::language()],
            "toml" => vec![tree_sitter_toml::language()],
            "tree_sitter_query" => vec![tree_sitter_query::language()],
            name => return Err(FormatterError::UnsupportedLanguage(name.to_string())),
        }
        .into_iter()
        .map(Into::into)
        .collect())
    }
```
Both functions are in `languages.rs`.

Xophmeister: This has always been my concern, too. The benefits of the config file are that you can associate a language with explicit extensions, rather than them being hardcoded, and you don't have to parse the query file _just_ to get the language name. However, the query and grammar are now almost completely divorced from each other; just related by virtue of filename.

Perhaps the config file also needs to have a `query_file` key for each language, to make this more explicit. (You could also imagine having a `grammar` key, for dynamic linking, but the case for this being embedded into the query file is even stronger.)

ErinvanderVeen: The old had the same problem, the connection language -> query file was based completely on the name of the language.

We haven't changed that, we've only removed the query -> language link to avoid having to parse the .scm file twice.

I have some plans to extend the configuration file with the grammar and maybe the query file in the future.

ErinvanderVeen (approved):

PR #468, opened 2023-05-22
Fix e2e tests 

nbacquey:
This PR fixes end-to-end tests for the playground, and fixes a leftover bug preventing tree-sitter queries for being correctly formatted


Xophmeister (approved):
:+1:

PR #472, opened 2023-05-23
Clippy pedantic

torhovland:
Minor improvements as suggested by `cargo clippy --all-targets --all-features -- -W clippy::pedantic`.

### DIFF
--- a/topiary/src/error.rs
+++ b/topiary/src/error.rs
@@ -103,8 +103,7 @@ impl fmt::Display for FormatterError {
 
             Self::Internal(message, _)
             | Self::Query(message, _)
-            | Self::Io(IoError::Filesystem(message, _))
-            | Self::Io(IoError::Generic(message, _)) => {
+            | Self::Io(IoError::Filesystem(message, _) | IoError::Generic(message, _)) => {
                 write!(f, "{message}")
             }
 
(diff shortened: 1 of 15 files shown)


topiary/src/error.rs
```diff
@@ -103,8 +103,7 @@ impl fmt::Display for FormatterError {
 
             Self::Internal(message, _)
             | Self::Query(message, _)
-            | Self::Io(IoError::Filesystem(message, _))
-            | Self::Io(IoError::Generic(message, _)) => {
+            | Self::Io(IoError::Filesystem(message, _) | IoError::Generic(message, _)) => {
```
Xophmeister, line 106: TIL you could do that :+1:

Xophmeister (approved):
I'm not 100% convinced that the alternation-style in `match` branches, so that the RHS mapping is unique, is always better -- e.g., the error mapping to source, as it puts the enum members out of alphabetical order -- but you can't argue with consistency, I suppose!

Otherwise, I think these are pretty good improvements. Would it be worth adding those Clippy flags to the `default.nix`?

torhovland:
> I'm not 100% convinced that the alternation-style in `match` branches, so that the RHS mapping is unique, is always better -- e.g., the error mapping to source, as it puts the enum members out of alphabetical order

Yes, I do see your point.

> Would it be worth adding those Clippy flags to the `default.nix`?

No, they are *too* pedantic. But I see now that we get even more suggestions by running:

```
cargo clippy --all-targets --all-features -- -W clippy::pedantic -W clippy::nursery -W clippy::cargo
```

Xophmeister (approved):
Looks like good changes to me :+1:

PR #477, opened 2023-05-24
languages_export.ts as build output

### LINKED ISSUE #436: Add `languages_export.ts` as a build target, instead of committing it to the repo
The playground uses `languages_export.ts` to get sample queries and inputs. This file is generated when building `topiary` (the library) from the files in `./languages/` and `./topiary/tests/samples/input/`. Because it is generated as a side-effect, the CI does not build it every time, only when the code in `./topiary/` changes. This would cause CI to fail when the file is not built. Therefore, it has been decided to add it to the repository (see discussion in #432).
Still, this is cumbersome for several reasons:
* It forces the maintainers to manually build the file and add the changes when the queries/sample inputs are modified.
* Because it may be re-built on the fly by the CI, the file that gets used by the playground may not be exactly the one that is committed to the repository.
* It adds a bit of noise, redundancy, and clunkiness to the diffs, because the file is 8k lines long.

I'd like the file to be generated as a proper nix derivation (or another mechanism), so that the CI would rebuild it every time it needs to, and it would no longer be part of the repository itself.

torhovland:
Fixes #436.


nbacquey (commented):
I'm not sure I understand the mechanism by which we ensure the file is always present, even if there was no need to rebuild `topiary-playground`. Could you please explain it?

torhovland:
The file is now an output of the Nix derivation. Nix will make sure to build the derivation if needed, or to just provide us a `result` symlink to a cached derivation when it can. In either case we have the file we need.

nbacquey (approved):

PR #480, opened 2023-05-24
Indicate experimental languages in web playground

### LINKED ISSUE #434: Warn the user when they are using "experimental" languages
I wonder if we should limit this to the "officially supported" languages? Or at least indicate in the playground whether a language is experimental or not.

_Originally posted by @Xophmeister in https://github.com/tweag/topiary/pull/432#discussion_r1170053417_

torhovland:
Fixes #434.


web-playground/src/App.css
```diff
@@ -67,4 +68,4 @@ button:not(:disabled) {
     background-color: #cd9ffa;
     border-color: #c792fa;
     color: #212529;
-}
+}
```
Xophmeister, line 71: Missing EOF \n

Xophmeister (approved):

web-playground/src/App.css
torhovland: VS Code kept doing that on CSS files. But I found the magic setting now. Fixed.

PR #482, opened 2023-05-25
Metadata fixes

torhovland:
Related to #478.

- Fixes Cargo metadata warnings.
- Fixes Crane config so we no longer to specify version in all Cargo.toml files.


nbacquey (approved):
Now that there's only one file containing the version number, we might want to change the release instructions in `MAINTAINERS.md` back to `bump the version in the root Cargo.toml`.

Otherwise, LGTM.
Thanks for addressing this issue!

PR #484, opened 2023-05-25
On-the-fly formatting

### LINKED ISSUE #357: Automated reformatting on change in the playground
The playground currently uses a button to trigger reformatting. This could instead be triggered on change of input or query.

Make sure this behaves well when the user types quickly and the formatting is slow.

It's probably a good idea to use [throttling](https://remysharp.com/2010/07/21/throttling-function-calls/) here.

Interactive reformatting as I type in the input box.

The current solution with a button is a simpler alternative. This could be combined with a keyboard shortcut, such as Ctrl+Enter, the way https://nickel-lang.org/playground does.

**Additional context**
See https://ormolu-live.tweag.io, https://tree-sitter.github.io/tree-sitter/playground, and https://regex101.com for inspiration.

torhovland:
Fixes #357.

### DIFF
--- a/web-playground/e2e/sample-tester.test.ts
+++ b/web-playground/e2e/sample-tester.test.ts
@@ -47,7 +47,7 @@ describe('test all grammars with puppeteer', () => {
         // less explicit and predictable.
         const onTheFlyCheckbox = await page.waitForSelector("#onTheFlyFormatting") ?? fail('Did not find checkbox');
         let isOnTheFlyEnabled = await (await onTheFlyCheckbox.getProperty("checked")).jsonValue();
-        
+
         if (isOnTheFlyEnabled)
             await onTheFlyCheckbox.click();
     });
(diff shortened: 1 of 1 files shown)


web-playground/e2e/sample-tester.test.ts
```diff
@@ -38,7 +38,15 @@ page.on("dialog", (dialog) => {
...
     beforeEach(async () => {
+        // Forward the console log from the browser
+        page.on('console', msg => console.log('PAGE LOG:', msg.text()));
+
         await page.goto('http://localhost:3000/playground');
+
+        // Test without on-the-fly formatting, because the debounce makes things
+        // less explicit and predictable.
+        const checkboxEl = await page.waitForSelector("#onTheFlyFormatting") ?? fail('Did not find checkbox');
+        await checkboxEl.click();
```
nbacquey, line 49: If I understand correctly, this disables on-the-fly formatting because the checkbox is ticked by default. This seems a bit error-prone: if we ever choose to disable on-the-fly formatting by default, this will enable it instead. Is there a `checkbox.disable()` instead?

nbacquey (changes requested):

PR #498, opened 2023-06-06
Fancy playground editor

### LINKED ISSUE #358: Nicer code editor in the playground
The playground currently uses simple textareas. This could be improved with something that looks more like a code editor.

A nice code editor looking playground.

If a parse error refers to line and column numbers, these should be easy to find in the input code.

The current solution is a simpler alternative that works.

**Additional context**
See https://nickel-lang.org/playground and https://jsfiddle.net for inspiration.

torhovland:
Fixes #358.

- Factor out Editor component.
- Use Ace editor.
- Working tests using Ace editor.
- Styling.
- Simplified code.


web-playground/e2e/sample-tester.test.ts
```diff
@@ -120,38 +118,37 @@ async function testInputFile(input: string, expected: string, query: string, lan
...
+    let textInput = await page.$(selector) ?? fail('Did not find text input control');
+    let textAreaSelector = `${selector} textarea`;
 
     // Clear the text area first, otherwise the following doesn't work.
-    await textInput.click({ clickCount: 3 });
+    await textInput.click();
+    await page.keyboard.down('ControlLeft')
+    await page.keyboard.press('KeyA')
+    await page.keyboard.up('ControlLeft')
     await textInput.press('Backspace');
```
Xophmeister, line 129: Out of curiosity: Is there not a better way of clearing the text area, rather than simulating Ctrl+A Backspace? Or does it have to be done this way to trigger the appropriate event?

web-playground/public/static/js/worker-html.js
Xophmeister, line 1: Yikes! This is vendored library code, right; I assume you didn't just produce 11k lines of JavaScript :sweat_smile:?... Anyway, does this need to be checked in, rather than somehow acquired at build time?

Xophmeister (approved):
As far as I can tell, this looks good to me...but you might want to get a second opinion!

The only thing that struck me is the very large `worker-html.js` file.

web-playground/e2e/sample-tester.test.ts
torhovland: That text editor has an API that can be used from within the app code, but as soon as this runs live in a browser it's just a big mess of HTML and magic. I have not found a better way to clear the text.

web-playground/public/static/js/worker-html.js
torhovland: Good point, I need to take a look at that.

PR #500, opened 2023-06-07
Process comments (and extra nodes) separately from the rest of the code

nbacquey:
This PR is an attempt at processing the comments separately from the rest of the code, so that formatting queries can be both resilient to the presence of random comments, and simple to write.

### Identifying comments
This is a challenge in itself:
- We can't use the same mechanism we use for leaf identification, _i.e._ a formatting directive in the query file that would look like `(block_comment) @comment`. Indeed, it would mean either running the query twice, or accounting for the presence of comments in the rest of the query file, both situations we want to avoid.
- The current implementation selects comment nodes from the CST with the following heuristics:
```node.is_extra() && node.kind().to_string().contains("comment")```. I think it should capture all types of comments for all supported languages, but I haven't checked that yet.
- Another option would be to have a separate query file for comments, one for each language, which would contain selecting and formatting directives for comments. It could also contain instructions on how to re-insert comments back in the code, once it's formatted.

### Anchoring comments
Each comment should be anchored to a non-comment node of the code. Ideally, it should be the node it's supposed to comment, but such semantics can't be deduced from the CST alone. This PR uses the following heuristics instead:
- If the comment node is alone on its line, anchor it to its next sibling in the CST. If it's the last of its siblings, anchor it to its previous sibling instead.
- If the comment node isn't alone on its line, anchor it to its previous sibling in the CST. If it's the first of its siblings, anchor it to its next sibling instead.
- I haven't thought at what it would mean for a comment node to be an only child in the CST, nor if that case can actually happen. In that case, it would probably be a safe bet to anchor it to its parent node.

### Extracting comments
`tree-sitter` grammars and queries don't offer lots of tools to edit an existing CST. The only reasonable way I've found is to use `input.replace_range()` to remove the comment's bytes from the input, then `tree.edit()` to mark a node as edited in the CST, then finally `reparse(old_tree, new_input, grammar)` to get the new CST, without comments. The query file would then be applied to this new CST.

### Re-inserting comments
This is a part I haven't had time to experiment with, but I think re-insertion should be done after processing all append/prepend directives, but before post-processing. I imagine something like this should work:
- If the anchor was before the comment, re-insert with `(anchor).(space).(comment)` (line breaks should already have been taken care of).
- If the anchor was after the comment, re-insert with `(comment).(line break).(anchor)`.

### State of the PR
- [x] Identifying comments
- [x] Anchoring comments
- [x] Extracting comments
- [x] Re-inserting comments
- [ ] Add more documentation where needed
- [ ] Update README

### DIFF
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -573,7 +573,7 @@ impl AtomCollection {
                 // REVIEW: the double borrow is sort of weird, is this idiomatic?
                 if node_section.contains(&(&comment).into()) {
                     match comment.commented {
-                        Commented::CommentedAfter(_) => self
+                        Commented::CommentedAfter { .. } => self
                             .comments_before
                             .entry(id)
                             .or_default()
@@ -933,7 +933,12 @@ impl AtomCollection {
 
     /// Create atoms out of comments, and put them at the correct place in the atom stream.
     fn post_process_comments(&mut self) {
-        let mut comments_queue: VecDeque<Atom> = VecDeque::new();
+        struct CommentWithContext {
+            comment: Atom,
+            blank_line_after: bool,
+            blank_line_before: bool,
+        }
+        let mut comments_queue_with_context: VecDeque<CommentWithContext> = VecDeque::new();
 
         // First pass: get atoms in reverse order, put comments that come before atoms at the correct position
         let mut atoms_with_comments_before: VecDeque<Atom> = VecDeque::new();
@@ -943,6 +948,8 @@ impl AtomCollection {
                 while let Some(Comment {
                     content,
                     original_column,
+                    blank_line_after,
+                    blank_line_before,
                 }) = self.comments_before.entry(id).or_default().pop_back()
                 {
                     let comment_atom = Atom::Leaf {
@@ -952,31 +959,60 @@ impl AtomCollection {
                         single_line_no_indent: false,
                         multi_line_indent_all: true,
                     };
-                    comments_queue.push_front(comment_atom);
+                    comments_queue_with_context.push_front(CommentWithContext{
+                        comment: comment_atom,
+                        blank_line_after,
+                        blank_line_before
+                    });
                 }
-            }
-            if Atom::Hardline == atom || Atom::Blankline == atom {
+                atoms_with_comments_before.push_front(atom);
+            } else if Atom::Hardline == atom || Atom::Blankline == atom {
+                let mut blank_line_before_first_comment = false;
                 // Prepend the comments, each one followed by a newline
-                while let Some(comment) = comments_queue.pop_back() {
-                    atoms_with_comments_before.push_front(Atom::Hardline);
+                while let Some(CommentWithContext{comment, blank_line_after, blank_line_before}) = comments_queue_with_context.pop_back() {
+                    if blank_line_after {
+                        atoms_with_comments_before.push_front(Atom::Blankline)
+                    } else {
+                        atoms_with_comments_before.push_front(Atom::Hardline);
+                    }
                     atoms_with_comments_before.push_front(comment);
+                    blank_line_before_first_comment = blank_line_before;
                 }
+                if blank_line_before_first_comment && atom == Atom::Hardline {
+                    atoms_with_comments_before.push_front(Atom::Blankline);
+                } else {
+                    atoms_with_comments_before.push_front(atom);
+                }
+            } else {
+                atoms_with_comments_before.push_front(atom);
             }
-            atoms_with_comments_before.push_front(atom);
         }
+        let mut blank_line_before_first_comment = false;
         // If we still have comments left, add them at the beginning of the file
-        while let Some(comment) = comments_queue.pop_back() {
-            atoms_with_comments_before.push_front(Atom::Hardline);
+        while let Some(CommentWithContext{comment, blank_line_after, blank_line_before}) = comments_queue_with_context.pop_back() {
+            if blank_line_after {
+                atoms_with_comments_before.push_front(Atom::Blankline)
+            } else {
+                atoms_with_comments_before.push_front(Atom::Hardline);
+            }
             atoms_with_comments_before.push_front(comment);
+            blank_line_before_first_comment = blank_line_before;
+        }
+        if blank_line_before_first_comment {
+            atoms_with_comments_before.push_front(Atom::Blankline);
         }
 
         // Second pass: get atoms in reverse order, put comments that come after atoms at the correct position
         let mut atoms_with_all_comments: VecDeque<Atom> = VecDeque::new();
+        let mut comments_queue: VecDeque<Atom> = VecDeque::new();
         while let Some(atom) = atoms_with_comments_before.pop_front() {
             if let Atom::Leaf { id, .. } = atom {
                 while let Some(Comment {
                     content,
                     original_column,
+                    // We don't care about blank lines here: they only matter for comments that
+                    // come before atoms.
+                    ..
                 }) = self.comments_after.entry(id).or_default().pop_front()
                 {
                     let comment_atom = Atom::Leaf {
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -1062,7 +1062,7 @@ impl AtomCollection {
         // antispaces may have produced more empty atoms.
         self.post_process_inner();
 
-        log::debug!("List of atoms after post-processing: {:?}", self.atoms);
+        log::debug!("List of atoms after post-processing: {:#?}", self.atoms);
     }
 
     /// This function post-processes the atoms in the collection.
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -1112,12 +1112,13 @@ impl AtomCollection {
                 ) => {
                     if head.dominates(moved_prev) {
                         *moved_prev = Atom::Empty;
+                        prev = head;
+                        remaining = tail;
                     } else {
                         *head = Atom::Empty;
+                        prev = moved_prev;
+                        remaining = tail;
                     }
-
-                    prev = moved_prev;
-                    remaining = tail;
                 }
                 // If a whitespace atom is followed by an indent atom, swap their positions.
                 (
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -931,14 +931,92 @@ impl AtomCollection {
         }
     }
 
+    /// Create atoms out of comments, and put them at the correct place in the atom stream.
+    fn post_process_comments(&mut self) {
+        let mut comments_queue: VecDeque<Atom> = VecDeque::new();
+
+        // First pass: get atoms in reverse order, put comments that come before atoms at the correct position
+        let mut atoms_with_comments_before: VecDeque<Atom> = VecDeque::new();
+
+        while let Some(atom) = self.atoms.pop() {
+            if let Atom::Leaf { id, .. } = atom {
+                while let Some(Comment {
+                    content,
+                    original_column,
+                }) = self.comments_before.entry(id).or_default().pop_back()
+                {
+                    let comment_atom = Atom::Leaf {
+                        content,
+                        id: self.next_id(),
+                        original_column,
+                        single_line_no_indent: false,
+                        multi_line_indent_all: true,
+                    };
+                    comments_queue.push_front(comment_atom);
+                }
+            }
+            if Atom::Hardline == atom || Atom::Blankline == atom {
+                // Prepend the comments, each one followed by a newline
+                while let Some(comment) = comments_queue.pop_back() {
+                    atoms_with_comments_before.push_front(Atom::Hardline);
+                    atoms_with_comments_before.push_front(comment);
+                }
+            }
+            atoms_with_comments_before.push_front(atom);
+        }
+        // If we still have comments left, add them at the beginning of the file
+        while let Some(comment) = comments_queue.pop_back() {
+            atoms_with_comments_before.push_front(Atom::Hardline);
+            atoms_with_comments_before.push_front(comment);
+        }
+
+        // Second pass: get atoms in reverse order, put comments that come after atoms at the correct position
+        let mut atoms_with_all_comments: VecDeque<Atom> = VecDeque::new();
+        while let Some(atom) = atoms_with_comments_before.pop_front() {
+            if let Atom::Leaf { id, .. } = atom {
+                while let Some(Comment {
+                    content,
+                    original_column,
+                }) = self.comments_after.entry(id).or_default().pop_front()
+                {
+                    let comment_atom = Atom::Leaf {
+                        content,
+                        id: self.next_id(),
+                        original_column,
+                        single_line_no_indent: false,
+                        multi_line_indent_all: true,
+                    };
+                    comments_queue.push_front(comment_atom);
+                }
+            }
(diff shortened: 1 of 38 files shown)

### CODE REVIEW

**Comment by 314eter** [comment]
In order to fix #430, this should be done for OCaml attributes too. They are not included with the current heuristic.

Attributes can appear almost everywhere in OCaml. To avoid overly complicating the grammar, tree-sitter-ocaml really allows them everywhere, even in some invalid places. If you replace the comments in the example of #489 with attributes, the idempotency rule will still be violated (the input will be parseable by tree-sitter, but the result will not be). But since the code was invalid OCaml to begin with, that's not a huge problem.

**Comment by ErinvanderVeen** [comment]
I think this heuristic can be avoided by using the languages.toml file.

File: topiary-core/src/comments.rs [hunk 1]
```diff
@@ -193,48 +207,56 @@ fn find_anchor<'tree>(
...
-        match anchor {
-            Commented::CommentedBefore(anchor) => {
-                log::warn!("Commented section precedes comment {:?}:\n{anchor:?}", &node)
-            }
-            Commented::CommentedAfter(anchor) => {
-                log::warn!("Commented section follows comment {:?}:\n{anchor:?}", &node)
-            }
-        }
-        new_input.replace_range((node.start_byte() as usize)..(node.end_byte() as usize), "");
+    // for each (comment, anchor) pair in reverse order, we:
```
Xophmeister, line 233: It might be worth mentioning that this is done in reverse order so that the deletions don't affect one another.

Xophmeister (commented):
I mostly follow this and it feels like it's on the right track. I did a commit-wise review, so some of my suggestions against earlier commits may no longer apply, or apply elsewhere.

I appreciate the modularisation of comments and types and the comments you added to the source to describe what's going on. This is quite a tricky operation, so those comments are super-encouraged to save our future selves' sanity :sweat_smile:

topiary-core/src/atom_collection.rs
```diff
@@ -37,6 +40,10 @@ pub struct AtomCollection {
     /// something to a node, a new Atom is added to this HashMap.
     /// The key of the hashmap is the identifier of the node.
     append: HashMap<usize, Vec<Atom>>,
+    /// Maps node IDs to comments before that node. Comments are stored in reading order.
+    comments_before: HashMap<usize, VecDeque<Comment>>,
+    /// Maps node IDs to comments after that node. Comments are stored in reading order.
+    comments_after: HashMap<usize, VecDeque<Comment>>,
```
Xophmeister, line 46: All comments before/after, or just those immediately before/after?

topiary-core/src/atom_collection.rs
```diff
@@ -535,15 +560,38 @@ impl AtomCollection {
...
                 multi_line_indent_all: false,
             });
             // Mark all sub-nodes as having this node as a "leaf parent"
             self.mark_leaf_parent(node, node.id());
+            // Test the node for comments
+            // `comments` is sorted in reverse order, so `pop()` gets the first one.
+            while let Some(comment) = comments.pop() {
+                let node_section: InputSection = node.into();
+                // REVIEW: the double borrow is sort of weird, is this idiomatic?
+                if node_section.contains(&(&comment).into()) {
```
Xophmeister, line 574: I can't speak to whether it's idiomatic (@yannham?); however, it makes sense here:

* `comment` is an `AnchoredComment` and you only implement `From<&AnchoredComment> for InputSection`; hence the first reference. You need to use `comment` later, so taking a reference makes sense otherwise you'll get a borrow violation.
* `InputSection::contains`'s argument expects a reference; hence the second. Here, the converted `InputSection` is only used once, so _maybe_ one could implement `From<&AnchoredComment> for &InputSection`, but that would probably require careful lifetime annotations.

My opinion is that the double-`&` is probably a better option than an additional, possibly complex `impl` that's only called once.

topiary-core/src/atom_collection.rs
```diff
@@ -535,15 +560,38 @@ impl AtomCollection {
...
+                        Commented::CommentedBefore(_) => self
+                            .comments_after
+                            .entry(id)
+                            .or_default()
+                            .push_back((&comment).into()),
+                    }
+                } else {
+                    comments.push(comment);
+                    break;
+                }
```
Xophmeister, line 590: > All comments before/after, or just those immediately before/after?

I believe this code answers my question: It's the just comments immediately before/after...I think.

topiary-core/src/atom_collection.rs
```diff
@@ -883,14 +931,138 @@ impl AtomCollection {
...
+            atoms_with_all_comments.push_back(atom);
+        }
+        // If we still have comments left, add them at the end of the file
+        while let Some(comment) = comments_queue.pop_back() {
+            atoms_with_all_comments.push_back(Atom::Space);
+            atoms_with_all_comments.push_back(comment);
+        }
+
+        self.atoms = atoms_with_all_comments.into()
+    }
```
Xophmeister, line 1054: I kind of see what's going on here, but I find it very hard to follow. Your comments help, but I think this big function would benefit from more documentation on the algorithm and the data structures that are being used.

topiary-core/src/atom_collection.rs
```diff
@@ -883,14 +931,138 @@ impl AtomCollection {
...
+
     /// This function merges the spaces, new lines and blank lines.
     /// If there are several tokens of different kind one after the other,
     /// the blank line is kept over the new line which itself is kept over the space.
     /// Furthermore, this function put the indentation delimiters before any space/line atom.
     pub fn post_process(&mut self) {
         self.post_process_scopes();
-        self.post_process_deletes();
         self.post_process_inner();
+        // Comments must be processed before deletes, because comment anchors might be deleted.
```
Xophmeister, line 1063: If that happens, then does that result in an orphaned comment error?

Xophmeister (commented):
This is as awesome as it is epic! Amazing work :bow:

* I've gone through this pretty thoroughly, but I must confess to not fully understanding everything. That said, you've named things very clearly, so I've used those names as a crux wherever I don't follow.  I've added comments if/when this "method" breaks down a bit.

* The PR description is very clear and I think -- if updated to reflect some of the changes since it was written -- it would make for good documentation in the `README`. (I've added some notes about the PR description below.)

* With regard to line-embedded comments /* like this */ what's the reason why they can't be anchored to the node (or the sibling node) immediately before the comment?

  (Similarly, for "orphaned" comments: Can they just assume the formatting of the current context, appended with a hardline, or is that not a safe fallback?)

* Meta: It's a real shame that Tree-sitter doesn't have a way, AFAIK, to introspect grammars. It would be really nice if one could just get a list of all "extra" nodes in any grammar, then you wouldn't need to bother with the `language.comment.scm` files... Perhaps worth an issue/feature request in the Tree-sitter repo :shrug:

* Meta: @Niols, would you be able to experiment with this branch against your OCaml codebase(s), paying attention to how it affects comment formatting? It would be really interesting to see how it plays with real world code.

---

Notes on the PR description, so it can be converted into documentation easily:

### Identifying comments

* It would be good to expand on why this isn't ideal, for the sake of the documentation:

  > - We can't use the same mechanism we use for leaf identification, _i.e._ a formatting directive in the query file that would look like `(block_comment) @comment`. Indeed, it would mean either running the query twice, or accounting for the presence of comments in the rest of the query file, both situations we want to avoid.

* Update this to match that you've gone with the second option; that is, comment query files:

  > - The current implementation selects comment nodes from the CST with the following heuristics:
  >   ```node.is_extra() && node.kind().to_string().contains("comment")```.  I think it should capture all types of comments for all supported languages, but I haven't checked that yet.
  > - Another option would be to have a separate query file for comments, one for each language, which would contain selecting and formatting directives for comments. It could also contain instructions on how to re-insert comments back in the code, once it's formatted.

### Anchoring comments

* Your heuristics make perfect sense to me and seem eminently sensible.  Some ASCII-art diagrams for the first two cases would make this section ideal for the documentation.

  It would also be worth documenting the:
  * interleaved comment edge-case (modulo my question, above);
  * and the case that the OCaml sample exposes, when comments are put after their node.

* Presumably, the second sentence in this heuristic isn't implemented:

  > - If the comment node is alone on its line, anchor it to its next sibling in the CST. If it's the last of its siblings, anchor it to its previous sibling instead.

  That's why we see final child comments losing their indent level, for example.

* It looks like you bail out in the only child case:

  > - I haven't thought at what it would mean for a comment node to be an only child in the CST, nor if that case can actually happen. In that case, it would probably be a safe bet to anchor it to its parent node.

  ```console
  $ cargo run -- fmt -lbash <<'SH'
  > (
  >   # foo
  > )
  > SH
  [2025-01-13T16:10:48Z ERROR topiary] Found an anchored comment, but could not attach it back to its anchor
      # foo
      The anchor was CommentedAfter { section: InputSection { start: Position { row: 2, column: 3 }, end: Position { row: 2, column: 3 } }, blank_line_after: false, blank_line_before: false }
  ```

  I agree with you that instances of only child comments are likely to be rare. Perhaps an option could be to just assume the formatting imposed by the parent as a fallback. Does that even make sense?

### Re-inserting comments

* I think this reads well; it just needs to be tweaked to change from the subjunctive to the indicative.

Niols:
At this point:

```ocaml
let dispatch = function
  | None -> (* FIXME: 404 page *) assert false

let () = ()
```

transforms into:

```ocaml
let dispatch = function
  | None -> assert false (* FIXME: 404 page *)

let () = ()
```

and then:

```ocaml
let dispatch = function
  | None -> assert false (* FIXME: 404 page *)
let () = ()
```

leading to an idempotency error.

Niols:
```ocaml
type t =
  (
    string,
    string,
    (string * person Slug.t list),
    (* bool, *)
    string,
    string,
    unit, (* FIXME *)
    string
  ) gen
[@@deriving yojson]
```
gets formatted by current `main` as:

```ocaml
type t =
  (string, string, (string * person Slug.t list), string, string,
  (* bool, *)
  unit, (* FIXME *)
  string) gen
[@@deriving yojson]
```

Both the former and the latter get formatted by this PR as:

```ocaml
type t =
  (* bool, *)
  (string, string, (string * person Slug.t list), string, string, unit, string) gen (* FIXME *)
[@@deriving yojson]
```




which, by this PR, gets formatted as:

Niols:
Last comment of a file gets put on the same line as the last non-comment, eg.:

```ocaml
let foo = ()
(** youpi foo *)

let bar = ()
(** youpi bar *)
```

becomes:

```ocaml
let foo = ()
(** youpi foo *)

let bar = () (** youpi bar *)
```

This happens even with an empty line before `(** youpi bar *)`

Niols:
As a single-line comment gets attached to the node after, it might lose wanted indentation, eg.:

```ocaml
[
  foo;
  (* bar; *)
]
```

becomes:

```ocaml
[
  foo;
(* bar; *)
]
```

and

```ocaml
module F = struct
  type t = unit
  (** comment for t *)
end
```

becomes:

```ocaml
module F = struct
  type t = unit
(** comment for t *)
end
```

mkatychev:
@nbacquey anything I can do to help split up this PR into tickets (and do the work ofc)? This is a solid approach as it opens up things like comment alignment.

mkatychev:
@nbacquey Just getting back, this sounds like a tricky issue, ASTs make sibling
relationships particularly tricky to deal with. I feel like having an `@anchor`
tag may be a particularly manual process especially if we have to enumarte
across multiple types in multiple languages.

Thinking through @Xophmeister's comment in
https://github.com/tweag/topiary/pull/500/files#r1914683695, I think an overfit
is initially ok (we can always say something like `node.is_extra() &&
node.has_tag("comment")`).

What are your thoughts on an approach that uses relative position where we we store an implicit (rather than explicit) anchor (I call it a slug in the example below)?

This would mean that things like column and row position would be built up after the fact and we would be constructing a linked list:
https://github.com/tweag/topiary/blob/7a31f23f3fd937585c61f9abb286624cac301e87/topiary-core/src/comments.rs#L209

```rs
enum NodeId {
  Significant(usize), // !node.is_extra()
  Extra(usize), // node.is_extra()
}

enum Margin {
  Newline, // \n\s{0}
  Whitespace, // \s*
  SpacedNewline, // (\n\s+|\s+\n)
}

type Sibling = (Margin, Option<NodeId>);
// struct ExtraNode/ExtraSlug?
struct CommentSlug {
  id: usize, // always a NodeId::Extra
  // Vec may moke more sense for precedence to differentiate between : [Sibling::Newline, Sibling::Whitespace]
  prev: Sibling, // for documentation's sake, these are tuples to express relative positioning
  next: Sibling,
}


```

1. pop all _extra_ nodes into a separate list collecting into a
   `Vec<CommentSlug>` (removing and serializing prev and next `Margin`s in the process)
1. format all significant nodes
1. join all extra nodes where prev/next node id is `Some(NodeId::Extra(other_id))`
1. Sort all `CommentSlug`s in this order into a `Vec<CommentSlug>` (or whatever):
  ```rs
  impl PartialOrd for CommentSlug {
      fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
          // tail slugs should always format first
          match (self.next.1, other.next.1) {
              (Some(_), None) => {
                  return Some(Ordering::Greater);
              }
              (None, Some(_)) => {
                  return Some(Ordering::Less);
              }
          }
          // head slugs get 2nd highest priority
          match (self.prev.1, other.prev.1) {
              unimplemented!(); // same logic as prev match
          }
          // median slugs get lowest priority

          // otherwise resolve by id, bigger (later) NodeId has priority
          self.id.partial_cmp(&other.id)
      }
  }
  ```
1. _handwaving `impl Serialize for Margin {}` here_
1. drain `Vec<CommentSlug>` inserting a serialized `(Margin, &'tree str, Margin)` in the relative byte offset

For comment alignment `CommentSlug` could always add a `parent: Option<ScopeId>`
field if needed (where scope is either a named scope or node id).

mkatychev:
Handwavy example for https://github.com/tweag/topiary/pull/500#issuecomment-3006574132:
```ocaml
type t =
  (
    string,
    string,
    (string * person Slug.t list),
    (* bool, *)
    string,
    string,
    unit, (* FIXME *)
    string
  ) gen
[@@deriving yojson]
```

* the `(* bool, *)` becomes: 
  ```rs
  CommentSlug {
    id: 3,
    prev: (Margin::SpacedNewline, Some(NodeId::Significant(2))), 
    next: (Margin::Newline, Some(NodeId::Significant(4))), 
  }
  ```
* the `(* FIXME *)` becomes: 
  ```rs
  CommentSlug {
    id: 6, 
    prev: (Margin::Whitespace, Some(NodeId::Significant(5))), 
    next: (Margin::Newline, Some(NodeId::Significant(7))), 
  }
  ```

Xophmeister:
@mkatychev I've only skimmed this, as @nbacquey is the expert, but I notice you're using node IDs. IIRC, these IDs aren't stable between Tree-sitter parses. Does that matter?

mkatychev:
Thanks for the feedback, and apologies for nagging, I think my proposal is full of holes but my assumption was that the section in [`tree_sitter::Node`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.id) here:

> -   If a new tree is created based on an older tree, and a node from the old tree is reused in the process, then that node will have the same id in both trees.

...suggests that [`tree_sitter::Parser::parse_with_options`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Parser.html#method.parse_with_options)'s `old_tree: Option<&Tree>` input parameter should be able to preserve the aforementioned node ID, did I skip over some logic?

mkatychev:
Moving to draft for now until we can give this (awesome) feature the attention it deserves

PR #506, opened 2023-06-13
Read user defined configuration file from $ROOT/.topiary/languages.toml

### LINKED ISSUE #505: User configuration file
### Proposed as a solution to https://github.com/tweag/topiary/discussions/504

Right now, the only way to configure the indentation is through the compile-time languages.toml file (which can be considered a regression from #448). We have always intended to also load a configuration file from `XDG_CONFIG/topiary/languages.toml`, but no issue existed. This issue serves as that.

ErinvanderVeen:
Resolves #505

This would allow users to define custom options.

This PR splits parsing of the configuration file into 2 parts. The first is parsing to toml::Value, this makes it easier to merge the values of multiple configuration files. From there it is `try_into` to our `Configuration`.

The `default_configuration` (2 functions) have been made part of the topiary crate itself. Other functions, like the one that reads from `.topiary/languages.toml` is part of the topiary-cli crate, which gets a new `configuration` module.

Remaining tasks:

- [ ] Find a way to find $ROOT, probably by looking up in the directories until we find .topiary


topiary/src/error.rs
```diff
@@ -207,3 +207,12 @@ impl From<tree_sitter_facade::ParserError> for FormatterError {
         Self::Internal("Error while parsing".into(), Some(Box::new(e)))
     }
 }
+
+impl From<toml::de::Error> for FormatterError {
+    fn from(e: toml::de::Error) -> Self {
+        Self::Internal(
+            "Error while parsing the internal configuration file".to_owned(),
```
nbacquey, line 214: I think `internal` is misleading here. Unless I'm mistaken, the error can also happen when reading a user-defined config file.

nbacquey (approved):

topiary/src/error.rs
ErinvanderVeen: Hej! The library is only responsible for parsing the default configuration, and such that error can only occur then. Parsing the user configuration is done in topiary-cli, and that error thus does have a different message: https://github.com/tweag/topiary/pull/506/files#diff-c77596c57513055d5393df388690627ffe805f456ccbed4f98e48d83304ea219R123

ErinvanderVeen: Having said that, at your discretion, it might be better to just .expect() after parsing the internal configuration file, this error should not occur anyway.

nbacquey: I just misunderstood, it's okay to leave it as it is.

torhovland:
Shouldn't these new file locations be mentioned in the `README.md`?

ErinvanderVeen:
Yes, of course! I'll make a quick follow-up PR.

PR #513, opened 2023-06-16
Loose mode for tolerating parsing errors

### LINKED ISSUE #363: Consider: a loose mode in which formatting works on ill-formed files
Our current design refuses to format a file that has parse errors. This is a sane default, as a file with parse errors may be misinterpreted and formatted very wildly compared to what is expected. But it is possible that, in some cases we want to try and do our best anyway.

I don't know for sure that this is something worth investing time on, so this issue is largely here to gauge users' apetite for something like this. If you are interested, please comment below, and describe the use-cases that you would want such best-effort formatting for.

As it's probably not a good default, the proposition is that we add a mode (a flag, presumably) whereby we don't error out on error nodes. Instead we replace the error node by its original content, unchanged, and apply the rest of the capture instructions normally.

N/A

**Additional context**
N/A

torhovland:
Fixes #363.

- Option for tolerating parse errors.
- Tolerate parse errors in playground.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,8 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        # macos-11 due to https://github.com/cachix/install-nix-action/issues/183
+        os: [ubuntu-latest, macos-11]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,8 @@ jobs:
   build:
     strategy:
       matrix:
-        # macos-11 due to https://github.com/cachix/install-nix-action/issues/183
-        os: [ubuntu-latest, macos-11]
+        # macos-13 due to https://github.com/cachix/install-nix-action/issues/183
+        os: [ubuntu-latest, macos-13]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,8 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        # macos-13 due to https://github.com/cachix/install-nix-action/issues/183
+        os: [ubuntu-latest, macos-13]
```
ErinvanderVeen, line 8: Please feel free to push changes like this directly to main.

topiary/src/atom_collection.rs
```diff
@@ -334,7 +334,10 @@ impl AtomCollection {
 
         if node.end_byte() == node.start_byte() {
             log::debug!("Skipping zero-byte node: {node:?}");
-        } else if node.child_count() == 0 || self.specified_leaf_nodes.contains(&node.id()) {
+        } else if node.child_count() == 0
+            || self.specified_leaf_nodes.contains(&node.id())
+            || node.kind() == "ERROR"
+        {
```
ErinvanderVeen, line 340:
```suggestion
        } else if node.child_count() == 0
            || self.specified_leaf_nodes.contains(&node.id())
            // We treat error nodes as leafs when `tolerate_parsing_errors` is set to true.
            // This ensures Topiary does not try to further apply transformations on them.
            // If `tolerate_parsing_errors` is set to false, this part of the code is only reached if the tree contains no ERROR nodes,
            // and as such the check below would be redundant.
            || node.kind() == "ERROR"
        {
```

ErinvanderVeen (changes requested):

ErinvanderVeen (approved):

PR #514, opened 2023-06-16
Linguist configuration to make Topiary a Rust project

ErinvanderVeen:
Topiary is considered by GitHub to be an OCaml project because our OCaml tests are rather substantial. I think it is best if we make it so GitHub identifies it as a Rust project for possible contributors.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Hide our tests from language detection so GitHub correctly detects Topiary as a Rust project
+topiary/tests/** linguist-vendored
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Hide our tests from language detection so GitHub correctly detects Topiary as a Rust project
+topiary/tests/** linguist-vendored
```
torhovland, line 2: Good idea. But strictly speaking, I think it should be:

```suggestion
topiary/tests/samples/** linguist-vendored
```

ErinvanderVeen:
Tested here: https://github.com/ErinvanderVeen/topiary

Issue #495, opened 2023-06-01
Document or automate the update process for WASM binaries


nbacquey:
When writing #494, I once again found myself unable to update the WASM grammar for OCaml, despite the instructions in `CONTRIBUTING.md`. I struggled with Docker configurations, manual ad hoc copy/pastes, and achieved nothing in the end.

I think we need to either automate the process, or make it fool-proof.

Xophmeister:
> I struggled with Docker configurations

I think the process launches a container, rather than you having to run everything inside a container.

That said, I agree: it took a while to work out, the first time I did it. Having it part of the CI build would be ideal, if it can be triggered _only_ when the dependency changes.

ErinvanderVeen:
Resolved with #610

PR #515, opened 2023-06-19
More details about how to update Wasm grammars.

torhovland:
Improves #495.


ErinvanderVeen:
Generally speaking, this entire process is not great and should be simplified, but this is a great improvement over what we had before.

ErinvanderVeen (approved):

PR #156, opened 2022-12-30
Dynamic loading of language support

ErinvanderVeen:
This PR is very early stage. Below are two lists; one for things that are required for a merge, one for optional improvements.

Required:
- [ ] Integrate the entire thing with FormatterResult/Error
- [ ] Resolve all of the TODO's introduced.
- [ ] Documentation
- [ ] Consolidate Configuration and LanguageDetection errors
- [ ] Have a builtin default configuration file, this is also what Helix does
- [ ] Pass CI ;)

Optional:
- [ ] Don't use `Repository::open` to test for availability.
- [ ] Allow some way for parsers to be present without their source-code. This would pave the way to providing the parsers via Nix.

PR #518, opened 2023-06-19
Simplify language matching by assuming the language exists

ErinvanderVeen:
We had two instances in language.rs where we matched on the language name (which was mostly the identity function), only to error if the language didn't exist.

I don't think we need to do this, especially as we move towards dynamic loading, where the engine would ideally not know about languages at all.

Also, the error will be found right away anyway, when trying the access the file.

There are some places where we cannot remove the matching, at least until #4 is implemented.


ErinvanderVeen:
Deprecated in favour of #156

PR #522, opened 2023-06-20
Resolve #519 catastrophic performances on ocaml files

### LINKED ISSUE #519: Catastrophic performances on OCaml files
**Describe the bug**

On trivial OCaml files, Topiary in its latest version takes forever to run.

**To Reproduce**

On my machine, this takes about 20s:
```console
$ echo 'let () = ()' > test.ml
$ time topiary -if test.ml
```
It was confirmed on @aspiwack's machine as well.

**Expected behavior**

I would expect Topiary to run under one second.

**Environment**
 - OS name + version: NixOS
 - Version of the code: ~v0.2.1,~ v0.2.2 and master all seem affected.
     _edit: v0.2.1 is slow, but not catastrophically slow; cf https://github.com/tweag/topiary/issues/525_

**Additional context**

Bisect and flamegraph coming soon.

ErinvanderVeen:
This PR temporarily reverts the OCaml bump so we can investigate what is going on with it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1505,9 +1505,8 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-ocaml"
-version = "0.20.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Niols (approved):
LGTM. (And I also tried it.)

PR #523, opened 2023-06-20
Create the query only once

ErinvanderVeen:
The query is then used between the formatting and the idempotence check. The TopiaryQuery struct helps keep the API clean.
#### main:
```
format_ocaml            time:   [655.12 ms 658.89 ms 662.75 ms]
```

#### after this PR:
```
format_ocaml            time:   [430.05 ms 432.33 ms 434.68 ms]
                        change: [-34.911% -34.385% -33.879%] (p = 0.00 < 0.05)
```

#### main:
flamegraph_main

#### after this PR:
flamegraph_new

### DIFF
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -20,7 +20,7 @@ use crate::{
     output::OutputFile,
     visualise::Visualisation,
 };
-use topiary::{formatter, Language, Operation, SupportedLanguage};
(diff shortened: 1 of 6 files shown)


Niols (approved):
LGTM. I'm no expert of the Rust code. I have benchmarked it quickly on my machine and it does bring a 40-50% speedup compared to the current version of `main`. The speedup becomes negligible when comparing with `-s`, which is to be expected. Nice one!

nbacquey (approved):
LGTM

PR #526, opened 2023-06-21
Indented leaves

### LINKED ISSUE #111: Multiline comments in Ocaml are badly indented
When an Ocaml file contains multiline comments, those are badly displayed because the first line is the only indented one.

So 
```ocaml
  (* The identity function does not have the declared type at all,
      it is just for formatting verification purpose.
  *)
  fun x -> x
```

can be turned into

```ocaml
    (* The identity function does not have the declared type at all,
      it is just for formatting verification purpose.
  *)
    fun x -> x
```

The reason for it, is that topiary evaluated that the function body should be indented by 2 levels whereas I indented it by only one, hence it indented the beginning of the comment with the function body, but since the whole comment is a leaf, the other lines kept the number of spaces I originally put in it.

Ideally, I would love the text in the comment to be indented consistently.

If not, possible, turning multiline comments into single line ones, for which this problem does not occur, could be a workaround.

If both are not possible (and modifying a leaf is not the purpose of tree-sitter queries, so both are probably not possible), then keeping the indentation the user gave to comments would probably be less ugly than the current result.

torhovland:
Fixes #111.

- Multi-line comment indenting.
- Documented new predicate, @multi_line_indent_all

Happily, the PR does not affect the benchmark.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ This name should be decided amongst the team before the release.
 
 ## [Unreleased]
 [unreleased]: https://github.com/tweag/topiary/compare/v0.2.3...HEAD
+* [#526](https://github.com/tweag/topiary/pull/526) Multi-line comments can be indented properly using the new predicate @multi_line_indent_all.
 
 ## [0.2.3] - 2023-06-20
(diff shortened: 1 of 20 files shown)


ErinvanderVeen (approved):
Really nice!

Issue #509, opened 2023-06-15
Exporting the query files


ErinvanderVeen:
Other projects that use the topiary library might want to make direct use of the queries shipped with Topiary, as proven by the effort by @vkleen to integrate Topiary into Nickel. These files should be exported and available for users of the library.

torhovland:
Not sure what you mean by exported, but there is `https://doc.rust-lang.org/std/macro.include_str.html`, which can be used to include file contents at compile-time.

Found it via https://users.rust-lang.org/t/packaging-assets-in-a-library-crate/4133.

Will create a sample and see how well it works.

torhovland:
@ErinvanderVeen @vkleen Would the above PR work for you?

Issue #511, opened 2023-06-15
Topiary as a git dependency breaks Nix builds with crane


vkleen:
This might be a rather specific failure case, but trying to use Topiary as a git repository dependency breaks Nix builds using [crane](https://crane.dev/). The problem is that in `topiary/src/configuration.rs` there is a `include_bytes!` macro importing `languages.toml` from the workspace root.

When `crane` vendors dependencies, the crates in the workspace end up in the vendor directory separately breaking the workspace structure. In particular, `languages.toml` is lost in the process. This ends up causing mysterious build failures.

It's debatable whether this really is an issue with Topiary. On the face of it, it feels like an issue with `crane`, but they might also be forced into doing this by the structure of a `cargo` vendor directory. I'm also wondering if something like this might lead to problems with publishing the crate to crates.io.

One possible solution would be to move `languages.toml` into the `topiary` subdirectory. I've verified that that fixes the immediate problem.

torhovland:
I have no strong feelings about this. With your recent PR a user should be able to override `languages.toml` with a file under `~/.topiary/`, and as long as that works, it's all good to me.

PR #528, opened 2023-06-22
Sample client and generators for built-in queries/languages.

### LINKED ISSUE #137: Make sure Topiary works when added as a crate in a third-party app
We need to make sure that the query files are distributed properly and that the client app will find them.

Being able to run Topiary from basically a Hello World app.

None.

**Additional context**
Related to #107 and #121.

torhovland:
Fixes #137 and #509.

- Sample client app.
- TopiaryQuery generators for each built-in language/query.


aspiwack:
I suppose this paves the way or removing the compile-time configuration of the directory which contains the (installed) query files, doesn't it?

vkleen:
I think this would work nicely for the `nickel format` use case, except that the location of the `languages` folder may produce the same issue as #511.

torhovland:
Hmm, I wonder if my sample works only because I use a path dependency, and I happen to have the query files where the compiler expects to find them. I need to investigate.

vkleen:
#511 would only be an issue with external Nix builds, I think.

torhovland:
`cargo package --list` is our friend. Interestingly, it shows that files like `languages.toml` and all our sample test files are included, but the query files are not.

torhovland:
OK, I see what's going on. `cargo package --list` is checking each of our workspace members. But the `languages` dir is outside of these, so it doesn't get included. I will try adding an explicit `include` now. Which makes me think we can do the same with that `languages.toml` from #511, and drop the symlink.

torhovland:
Although it looks like we cannot include files from parent directories, so the symlink route is the way to go anyway:
https://users.rust-lang.org/t/including-files-from-parent-directory-in-package/88969

torhovland:
So the PR should be fine now.

torhovland:
> I suppose this paves the way or removing the compile-time configuration of the directory which contains the (installed) query files, doesn't it?

Hmm, not directly. Because Topiary doesn't itself use these convenience generator functions for now. It could, but we already have some more complex logic for looking up query files. We need to keep this, because it allows users to override the built-in queries.

```
/// Convert a Language into the canonical basename of its query file, under the most appropriate
/// search path. We test 3 different locations for query files, in the following priority order,
/// returning the first that exists:
///
/// 1. Under the `TOPIARY_LANGUAGE_DIR` environment variable at runtime;
/// 2. Under the `TOPIARY_LANGUAGE_DIR` environment variable at build time;
/// 3. Under the `./languages` subdirectory.
```

torhovland:
Actually, making sure the `languages` directory is included in the crate, the same way we handle `languages.toml`, is the real fix of #509. The convenience functions are just for, well, convenience, and will likely get dropped when we support dynamic loading of grammars (#4).

Maybe it's best to drop them right now. I won't object to that.

@vkleen Can you check if this branch helps you, even without using the new convenience functions?

vkleen:
Building with crane and Nix fails with the branch as it is. Crane ends up copying dangling symlinks into the cargo vendor directory. Maybe that is an issue that should be addressed in crane upstream. Once topiary makes its way to crates.io, I imagine it will work fine. As you say, `cargo package` replaces the symlinks with copies of the actual files for publishing.

Just putting the query files into the package sources for cargo wouldn't help with accessing those files in `nickel format`, I think. There would still need to be either the helper functions like `topiary::nickel()` or just a global static string with `include_str!()` to get the right queries.

torhovland:
> Building with crane and Nix fails with the branch as it is. Crane ends up copying dangling symlinks into the cargo vendor directory. Maybe that is an issue that should be addressed in crane upstream. Once topiary makes its way to crates.io, I imagine it will work fine. As you say, `cargo package` replaces the symlinks with copies of the actual files for publishing.

I was trying to be clever, and switched the direction of the symlinks, so the real files could stay in root. But it sounds like that's what breaking Crane. Please try again now.

> Just putting the query files into the package sources for cargo wouldn't help with accessing those files in `nickel format`, I think. There would still need to be either the helper functions like `topiary::nickel()` or just a global static string with `include_str!()` to get the right queries.

I see. Then we'll keep them. But I wonder what the scalable, future solution should be when at some point we consider languages something that is loaded dynamically. @ErinvanderVeen

vkleen:
The nix builds succeed with the switch symlinks 👍

ErinvanderVeen:
It will sure be interesting to see how we solve this in the future!

samples/client-app/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "topiary-client-app"
+version = "0.0.1"
+edition = "2021"
+
+[dependencies]
+tokio = { version = "^1.28.1", features = ["rt-multi-thread", "macros"] }
+topiary = { path = "../../topiary" }
```
ErinvanderVeen, line 8:
```suggestion
topiary = { path = "../../topiary" }

```

ErinvanderVeen (changes requested):
Otherwise looks great!

ErinvanderVeen (approved):

PR #534, opened 2023-06-26
Parallel processing of multiple files

### LINKED ISSUE #521: Accept Variadic Arguments of INPUT_FILE
It's really annoying and not very intuitive to have to type `topiary --input-file file1 --input-file file2 -- input-file file3` or `topiary -f file1 -f file2 -f file3`. It also makes it difficult to integrate topiary with treefmt.

Write the cli such that one can type `topiary file1 file2 file3`.

**Additional context**
treefmt stipulates this in their [specification](https://numtide.github.io/treefmt/formatters-spec/#rules).

ErinvanderVeen:
This PR implements passing multiple files to topiary through the -f flag. Passing multiple files implements inplace formatting.

It also parallelizes the the formatting of these files (without sharing the Query, which is left to a future PR).

Resolves #521


topiary-cli/src/main.rs
```diff
@@ -158,16 +197,35 @@ async fn run() -> CLIResult<()> {
...
+                )
+                .unwrap();
+
+                output.into_inner().unwrap().persist().unwrap();
+            })
+        })
+        .collect();
+
+    for task in tasks {
+        task.await.unwrap();
```
torhovland, line 227: Isn't this just running them sequentially? Perhaps use `par_iter()` from Rayon?

topiary-cli/Cargo.toml
```diff
@@ -28,6 +28,8 @@ path = "src/main.rs"
 # Eventually we will want to dynamically load them, like Helix does.
 clap = { workspace = true, features = ["derive"] }
 env_logger = { workspace = true }
+futures = { workspace = true }
+tree-sitter-facade = { workspace = true }
```
torhovland, line 32: Can we avoid this dependency if we let the library convert from `topiary::Language`?

topiary-cli/src/main.rs
```diff
@@ -158,16 +197,35 @@ async fn run() -> CLIResult<()> {
...
+                };
+                let mut output: BufWriter<OutputFile> =
+                    BufWriter::new(OutputFile::new(&output).unwrap());
+
+                formatter(
+                    &mut input,
+                    &mut output,
+                    &query,
+                    &language,
+                    &grammar,
```
torhovland, line 216: Do we even need `grammar` as an argument? It's just derived from `language`.

topiary-cli/src/main.rs
ErinvanderVeen: I don't think so, this is from the documentation:
> The provided future will start running in the background immediately when spawn is called, even if you don’t await the returned JoinHandle.

https://docs.rs/tokio/latest/tokio/task/fn.spawn.html

topiary-cli/Cargo.toml
ErinvanderVeen: It was really only for the type, which we can infer with `_`. However, since the app is the foremost client-app for the library, we might still want to do this.

topiary-cli/src/main.rs
ErinvanderVeen: No we don't but I would like to discuss this real quick with you when we meet next.

ErinvanderVeen: For documentation; we decided to leave this until we decided to refactor the API.

flake.nix
```diff
@@ -62,7 +62,7 @@
                 name = "topiary-inplace";
                 text = ''
                   for FILE; do
-                    if ${code.topiary-cli}/bin/topiary --in-place --input-file "$FILE"; then
+                    if ${code.topiary-cli}/bin/topiary --in-place --input-files "$FILE"; then
```
Xophmeister, line 65: You don't need this loop any more. I think you can replace the call to Topiary with:

```bash
topiary --in-place --input-files "$@"
```

topiary-cli/src/error.rs
```diff
@@ -134,3 +147,12 @@ impl From<serde_toml_merge::Error> for TopiaryError {
         )
     }
 }
+
+impl From<tokio::task::JoinError> for TopiaryError {
+    fn from(e: tokio::task::JoinError) -> Self {
+        TopiaryError::Bin(
+            format!("Could not join parallel formatting tasks: {}", e),
```
torhovland, line 154: You shouldn't have to display the inner message from `e`. Because you supply `e` as an inner error (source), it is the error reporter's responsibility to show the inner message.

topiary-cli/src/main.rs
```diff
@@ -194,32 +194,33 @@ async fn run() -> CLIResult<()> {
...
-                    &mut output,
-                    &query,
-                    &language,
-                    &grammar,
-                    operation,
-                )
-                .unwrap();
-
-                output.into_inner().unwrap().persist().unwrap();
+                    let input = &input;
```
torhovland, line 199: Could you instead just do `match &input...` below?

topiary-cli/src/main.rs
```diff
@@ -194,32 +194,33 @@ async fn run() -> CLIResult<()> {
...
+                    output.into_inner()?.persist()?;
+
+                    Ok(())
             })
         })
         .collect();
 
     for task in tasks {
-        task.await.unwrap();
+        task.await??;
```
torhovland, line 223: Interesting. This deserves a comment.

torhovland (approved):

PR #535, opened 2023-06-27
530 idempotency parsing error message

### LINKED ISSUE #530: Better reporting on parsing error caused by idempotency problems

Consider the following piece of code:
```ocaml
```
It is valid OCaml code. However, running latest Topiary reports a parsing error:
```
$ nix run github:tweag/topiary/cba1993 -- -l ocaml <<EOF
EOF
[2023-06-22T18:18:45Z ERROR topiary] Parsing error between line 1, column 46 and line 1, column 47
```
The error comes in fact from the fact that the first pass of Topiary produces an invalid file (cf https://github.com/tweag/topiary/issues/529) and the second pass then fails parsing it. In this case, the location is somewhat correct but in other situations the line numbers might be completely wrong (if previous lines were affected by the first pass of formatting), leaving a user clueless.

I want an error message that explains that it is an idempotency error and that it is not the user's fault and that it should be reported.

**Additional context**

If I am not mistaken, v0.1.0 used to do this?

torhovland:
Fixes #530.

- Improved error message when idempotency fails due to invalid output in the first pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ This name should be decided amongst the team before the release.
 * [#526](https://github.com/tweag/topiary/pull/526) Multi-line comments can be indented properly using the new predicate @multi_line_indent_all.
 * [#528](https://github.com/tweag/topiary/pull/528) Added a sample app and convenience functions for using the built-in queries.
 * [#533](https://github.com/tweag/topiary/pull/533) Update tree-sitter-ocaml to 0.20.3
+* [#535](https://github.com/tweag/topiary/pull/535) Improved error message when idempotency fails due to invalid output in the first pass.
 
 ## [0.2.3] - 2023-06-20
 [0.2.2]: https://github.com/tweag/topiary/compare/v0.2.2...v0.2.3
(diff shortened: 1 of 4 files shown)


Niols:
Then:
```
$ nix run github:tweag/topiary/cba1993 -- -l ocaml <<EOF
EOF
[2023-06-27T08:37:48Z ERROR topiary] Parsing error between line 1, column 46 and line 1, column 47
```

Now:
```
$ nix run github:tweag/topiary/a3a04b3 -- -l ocaml <<EOF
EOF
[2023-06-27T08:37:56Z ERROR topiary] The formatter produced invalid output and
    failed when trying to format twice (idempotence check).
    
    If this happened with the built-in query files, it is a bug. It would be
    helpful if you logged this error at
    https://github.com/tweag/topiary/issues/new?assignees=&labels=type%3A+bug&template=bug_report.md
    
    The following is the error received when running the second time, but note
    that any line and column numbers refer to the formatted code, not the
    original input. Run Topiary with the --skip-idempotence flag to see this
    invalid formatted code.
[2023-06-27T08:37:56Z ERROR topiary] Cause: Parsing error between line 1, column 46 and line 1, column 47
```

Looks great to me!

nbacquey (approved):

PR #536, opened 2023-06-27
OCaml spacing error

### LINKED ISSUE #529: Idempotency error (and other peculiar behaviours) with `function` inside an `if-then-else` construct
**Describe the bug**

On:
```ocaml
```
Topiary produces an idempotency error, the first pass producing:
```ocaml
let id = if true then Fun.id else functionx -> x
```
which is not syntactically correct.

Notes:
- This also happens with more arguments to `function`, eg. `function x -> x | y -> y`.
- This also happens with a multiline `if-then-else`.
- This also happens with a multiline `function`, including with a newline between `function` and `x -> x`.
- This does not happen for `let id = function x -> x`.
- This does not happen for `let id = if true then Fun.id else function | x -> x` and the `|` remains, while it is removed from `let id = function | x -> x`.

**To Reproduce**

```
nix run github:tweag/topiary/cba1993 -- -l ocaml <<EOF
EOF
```
will produce:
```
[2023-06-22T18:13:51Z ERROR topiary] Parsing error between line 1, column 46 and line 1, column 47
```
while its `-s` counterpart will produce:
```
let id = if true then Fun.id else functionx -> x
```

torhovland:
Fixes #529.

### DIFF
--- a/topiary/tests/samples/expected/ocaml.ml
+++ b/topiary/tests/samples/expected/ocaml.ml
@@ -13,6 +13,24 @@
 (*                                                                        *)
 (**************************************************************************)
 
+(* Issue 529 *)
+
+let id1 = if true then Fun.id else function x -> x
+
+let id2 = if true then Fun.id else function x -> x | y -> y
+
+let id3 =
+  if true then
+    Fun.id
+  else
+    function x -> x
+
+let id4 = function x -> x
+
+let id5 = if true then Fun.id else function x -> x
+
+let id6 = function x -> x
+
 (* Extensible buffers *)
 
 type t = (* Multi-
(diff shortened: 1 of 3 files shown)


topiary/tests/samples/expected/ocaml.ml
```diff
@@ -13,6 +13,24 @@
...
+
+let id3 =
+  if true then
+    Fun.id
+  else
+    function x -> x
+
+let id4 = function x -> x
+
+let id5 = if true then Fun.id else function | x -> x
```
torhovland, line 30: We should be able to drop the bar here. See issue for discussion.

nbacquey: The bar should be dropped, now that #537 is merged. I think we can also keep the spacing after `fun` and `function`, for additional guarantees.

PR #537, opened 2023-06-27
Fix: open scope "function_definition" more often

### LINKED ISSUE #529: Idempotency error (and other peculiar behaviours) with `function` inside an `if-then-else` construct

nbacquey:
Closes #529
Modifies #536

### DIFF
--- a/topiary/languages/ocaml.scm
+++ b/topiary/languages/ocaml.scm
@@ -789,6 +789,14 @@
   ) @begin_scope @end_scope
   (#scope_id! "function_definition")
 )
+(
+  (_) @begin_scope
+  .
+  (function_expression
+    "function" @append_spaced_scoped_softline
+  ) @end_scope
+  (#scope_id! "function_definition")
+)
 (function_expression
   (match_case)? @do_nothing
   .


nbacquey:
Fun fact: I wanted to write
```
(
  _ @begin_scope
  .
  (function_expression
    "function" @append_spaced_scoped_softline
  ) @end_scope
  (#scope_id! "function_definition")
)
```
With `_` instead of `(_)`, to capture both named and non-named nodes, and get rid of the two cases above which open the `function_definition` scope.
Alas, the snippet above is not syntactically correct, for some reason?

torhovland (approved):

torhovland:
Unfortunately, your "fun" fact is causing problems.

`(_)` is not capturing the `else` block, because it is unnamed. And `_` causes a query error, as you say. This actually works, but is not very general:

```scheme
(
  "else" @begin_scope
  .
  (function_expression
    "function" @append_spaced_scoped_softline
  ) @end_scope
  (#scope_id! "function_definition")
)
```

torhovland:
This also doesn't work, and I'm not sure why:

```scheme
(
  (_) @begin_scope
  .
  (
    (function_expression
      "function" @append_spaced_scoped_softline
    ) @end_scope
  )
  (#scope_id! "function_definition")
)
```

nbacquey:
I've thought about something: even if the following was a correct query, we shouldn't use it:
```
(
  _ @begin_scope
  .
  (function_expression
    "function" @append_spaced_scoped_softline
  ) @end_scope
  (#scope_id! "function_definition")
)
```
That's because we know `.` can skip anonymous nodes, so if we have something like `(let) "=" (function_expression)`, then `_ @begin_scope` would match on both `(let)` and `"="`, and the scope would be opened twice, in different places.

torhovland:
Right, but both of them would also be closed. Wouldn't that be OK?

nbacquey:
That would be okay-ish. Both scopes would be closed indeed, but weird thing could happen if there were matches that use the `function_definition` scope between `(let)` and `"="`, like a scoped newline, or worse, closing a previously opened scope.

PR #550, opened 2023-07-02
Fix handling of antispace in post-processing.

### LINKED ISSUE #549: `Hardline, Antispace, IndentEnd` should be post-processed to `IndentEnd, Empty, Hardline`
**Describe the bug**
Post-processing is trying to put `IndentEnd` before `Hardline`, but fails in this case.

**To Reproduce**
The test suite is formatting Nickel code like this:

```
(
  # Foo
  )
```

**Expected behavior**
The following Nickel code

```
(
  # Foo
  )
```

should be formatted as

```
(
  # Foo
)
```

torhovland:
Fixes #549.

### DIFF
--- a/topiary/src/atom_collection.rs
+++ b/topiary/src/atom_collection.rs
@@ -67,6 +67,26 @@ pub struct AtomCollection {
 }
 
 impl AtomCollection {
+    /// Returns a basic AtomCollection with the supplied atoms. Only used for
+    /// testing. Normally you should use `AtomCollection::collect_leafs`
+    /// instead.
+    pub fn new(atoms: Vec<Atom>) -> Self {
+        Self {
+            atoms,
+            prepend: HashMap::new(),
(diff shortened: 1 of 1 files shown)


topiary/src/atom_collection.rs
```diff
@@ -67,6 +67,26 @@ pub struct AtomCollection {
 }
 
 impl AtomCollection {
+    /// Returns a basic AtomCollection with the supplied atoms. Only used for
+    /// testing. Normally you should use `AtomCollection::collect_leafs`
+    /// instead.
+    pub fn new(atoms: Vec<Atom>) -> Self {
```
ErinvanderVeen, line 73: Unless we find a use outside of tests for this one, I think it would be nice to use `#[cfg(test)]` here.

torhovland: Fixed.

Issue #547, opened 2023-06-30
Idempotency errors on the Nickel test suite


yannham:
**Describe the bug**
Running Topiary on the Nickel source files from Nickel's test suite produces a bunch of idempotency errors. In particular on the following files:

- [pass/contracts/contracts.ncl](https://github.com/tweag/nickel/blob/99531e68ad5cf325afeda43948c84fe38fa26d5b/core/tests/integration/pass/contracts/contracts.ncl)
- [pass/merging/metavalues.ncl](https://github.com/tweag/nickel/blob/99531e68ad5cf325afeda43948c84fe38fa26d5b/core/tests/integration/pass/merging/metavalues.ncl)
- [pass/stdlib/string_conversions.ncl](https://github.com/tweag/nickel/blob/99531e68ad5cf325afeda43948c84fe38fa26d5b/core/tests/integration/pass/stdlib/string_conversions.ncl)
- [fail/contracts/poly_record_contract_bad_field.ncl](https://github.com/tweag/nickel/blob/99531e68ad5cf325afeda43948c84fe38fa26d5b/core/tests/integration/fail/contracts/poly_record_contract_bad_field.ncl)
- [fail/contracts/poly_record_contract_higher_order_fn.ncl](https://github.com/tweag/nickel/blob/99531e68ad5cf325afeda43948c84fe38fa26d5b/core/tests/integration/fail/contracts/poly_record_contract_higher_order_fn.ncl)

There is a plain parse error on [pass/strings/string_interpolation.ncl](https://github.com/tweag/nickel/blob/master/core/tests/integration/pass/strings/string_interpolation.ncl), but this is heavy on nested string interpolations, and the tree-sitter grammar might be the culprit here.

**To Reproduce**
Check out the Nickel repo, and then:

```console
$ cd core/tests/integration
$ for f in **/*.ncl; do echo "Checking $f..."; topiary -i -f "$f"; done 
...
errors
...
```

Some tests are not supposed to parse (in particular in `fail/parser`), so those are expected to fail.

**Expected behavior**
Beside tests that aren't parsable, I expect the idempotency check to succeed on the Nickel test suite.

**Environment**
 - Version of the code: b4b3aa9f6fb6f8d220bcae033514327a32557786

ErinvanderVeen:
I took a look at the first one (`contracts.ncl`).

[...]

PR #553, opened 2023-07-03
Nickel: Start infix-chains also for annotated infix_expr

ErinvanderVeen:
This resolves part of #547, in particular the issue investigated by @torhovland.

```
[
  a && (b & c | d),
  a &&
  (b & c | d),
  a &&
  (b &
  c | d)
]
```
Get correctly formatted as:
```
[
  a && (b & c | d),
  a
  && (b & c | d),
  a
  && (
    b
    & c | d
  )
]
```
@torhovland's reasoning was correct, there was a scoping issue. The scope was started by the `[`, ending after the infix expression. Since the `[` was on a previous line, the entire expression was considered to be multi-line, and thus expanded as such. The scope was started there (and not on the next line) because that line was considered a `annotated_infix_expr` instead of an `infix_expr` which we did not include in our query to start the scope.

### DIFF
--- a/topiary/languages/nickel.scm
+++ b/topiary/languages/nickel.scm
@@ -118,7 +118,10 @@
 ; operators, which fall under nodes (infix_b_op_7) and (infix_b_op_8).
 (uni_term
   (#scope_id! "infix_chain")
-  (infix_expr) @begin_scope
+  [
+    (infix_expr)
+    (annotated_infix_expr)
+  ] @begin_scope
 ) @end_scope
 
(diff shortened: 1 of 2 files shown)


ErinvanderVeen:
Also, this PR makes it so that the debug messages for appending and prepending are correct. Previously these messages always reported `Empty` for the Atom, since the Atom was removed. For our sanity it also removed the message entirely when it is empty.

torhovland (approved):

PR #557, opened 2023-07-05
Fall back to compiled-in queries for supported languages.

### LINKED ISSUE #542: Make sure the release binaries have working queries out of the box
We are now releasing binaries with `cargo-dist`. See #538. But we are not distributing any query files. This means that a downloaded package will not work out of the box.

Either we have to find a way to include the query files with `cargo-dist`, or, probably preferably, the Topiary CLI can use the new convenience functions to access compile-time query files. See #528.

The user can always download query files manually.

torhovland:
Fixes #542.

### DIFF
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -136,8 +136,16 @@ async fn run() -> CLIResult<()> {
         )]
     };
 
+    type IoFile = (
+        String,
+        String,
+        Language,
+        Option<PathBuf>,
+        CLIResult<PathBuf>,
+    );
+
     // Add the language and query Path to the io_files
-    let io_files: Vec<(String, String, Language, PathBuf)> = io_files
+    let io_files: Vec<IoFile> = io_files
         .into_iter()
         // Add the appropriate language to all of the tuples
         .map(|(i, o)| {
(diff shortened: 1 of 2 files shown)


topiary-cli/src/main.rs
```diff
@@ -123,30 +123,41 @@ async fn run() -> CLIResult<()> {
...
+                "Could not open query file".into(),
+                Some(CLIError::IOError(e)),
+            )
+        })
+    })
+    .and_then(|query_content: String| Ok(TopiaryQuery::new(&grammar, &query_content)?))
+    .or_else(|e| {
+        // If we weren't able to read the query file, and the user didn't
+        // request a specific query file, we should fall back to the built-in
+        // queries.
```
ErinvanderVeen, line 153: Topiary should log that it is falling back to the builtin query file to avoid future headaches in debugging.

torhovland: Good point!

torhovland: Check now.

torhovland:
@ErinvanderVeen This changed a bit now, after a rebase on main. Please take another look.

ErinvanderVeen (approved):

PR #560, opened 2023-07-06
Keep query between runs when we can in the playground, for performance.

### LINKED ISSUE #545: Playground: Ocaml formatter too slow for on-the-fly formatting
**Describe the bug**
Formatting some Ocaml program in the playground takes ~3s (eyeballed on my machine). This is too slow period. But it's unbearable in on-the-fly formatting mode.

I assume that virtually all the time is spent in `Query::new()` like in #539 (and more generally #443).

**Expected behavior**
Formatting should take less thant 0.5s .

torhovland:
Fixes #545.


ErinvanderVeen (approved):
Hard for me to review, but I don't see anything wrong with it.

PR #562, opened 2023-07-07
Use match predicates to avoid pipe symbols in infix expression formatting

ErinvanderVeen:
Since we can not string match on named nodes and the pipe operators `|>` are two anonymous nodes, this commit uses the `#match?` predicate to separate these from the other `rel_operators`.

### DIFF
--- a/topiary/languages/ocaml.scm
+++ b/topiary/languages/ocaml.scm
@@ -1393,6 +1393,18 @@
 ; top-level one: it is the one that is not preceded by an infix operator.
 ; We only consider logic operators, as not to mess with arithmetic expressions
 (
+  (_)? @do_nothing
+  (#match? @do_nothing "^\\|>$")
+  .
+  (infix_expression) @begin_scope @end_scope
+  (#scope_id! "infix_expression")
+)
+(infix_expression
+  (_) @prepend_spaced_scoped_softline
+  (#match? @prepend_spaced_scoped_softline "^\\|>$")
+  (#scope_id! "infix_expression")
+)
+(
   [
     (and_operator)
     (or_operator)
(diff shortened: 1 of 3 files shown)


topiary/languages/ocaml.scm
```diff
@@ -1392,6 +1392,18 @@
...
+  (#match? @do_nothing "^\\|>$")
+  .
+  (infix_expression) @begin_scope @end_scope
+  (#scope_id! "infix_expression")
+)
+(infix_expression
+  (_) @prepend_spaced_scoped_softline
+  (#match? @prepend_spaced_scoped_softline "^\\|>$")
+  (#scope_id! "infix_expression")
+)
```
Xophmeister, line 1406: I'm not sure I follow this (but if it works, it works :sweat_smile:)

The first rule creates a scope over an `infix_expression` node, providing that node is not immediately preceded by `|>`.

The second rule prepends a softline after the `|>` in the `infix_expression` node, when the scope is active.

Aren't those mutually exclusive? If the `|>` is a sibling the `infix_expression` node when the scope is defined, how can it also be a child of the `infix_expression` when the scope is used?

ErinvanderVeen: The `match?` predicate does is not need to follow the previous node. Instead, it applies the previous `@do_nothing` only if the node that `@do_nothing` was attached to matches the regular expression.

Xophmeister: Does that not mean you have `("|>" . (infix_expression))` when there's a match (i.e., sibling nodes)?

ErinvanderVeen: Ahh, yes. I see. This happens when there is a chain of "|>" operators. Like in this example from the linked issue:
```ocaml
foo
|> bar |> baz
```
Which has this tree:
```
(compilation_unit [0, 0] - [2, 0]
  (expression_item [0, 0] - [1, 13]
    (infix_expression [0, 0] - [1, 13]
      left: (infix_expression [0, 0] - [1, 6]
        left: (value_path [0, 0] - [0, 3]
          (value_name [0, 0] - [0, 3]))
        operator: (rel_operator [1, 0] - [1, 2])
        right: (value_path [1, 3] - [1, 6]
          (value_name [1, 3] - [1, 6])))
      operator: (rel_operator [1, 7] - [1, 9])
      right: (value_path [1, 10] - [1, 13]
        (value_name [1, 10] - [1, 13])))))
```

torhovland:
Like @Xophmeister says, if it works it works. But it needs tons of commenting before you merge 😄

ErinvanderVeen:
I'm leaving this as a draft until we have discussed https://github.com/tweag/topiary/issues/541#issuecomment-1626460317

PR #573, opened 2023-07-10
OCamllex support

ErinvanderVeen:
Adds OCamllex formatting to Topiary, explicitly without support for language injection. That will be a PR for another day.


topiary/tests/samples/expected/ocamllex.mll
```diff
@@ -0,0 +1,67 @@
+(**************************************************************************)
+(*  Taken from https://github.com/colis-anr/morbig/                       *)
```
Niols, line 2: If I recognize it correctly, this is actually from paris-branch/dancelor!

ErinvanderVeen: Whoops! That's a TODO then. I should also take a look at the license and do it correctly.

Niols: Dancelor has no license and therefore re-use is not permitted. It is mostly laziness from my end, but that's how it applies for now. However, I am the sole author of the file in question, so I will happily be giving you free use of it.

314eter:
> There exists an issue right now with `(action)` nodes (which consist of only a `{`, `(ocaml)`, and `}`). The spaces after the `{` are part of the `(action)` scope, but the spaces before the `}` are part of the `(ocaml)`.

This is fixed in tree-sitter-ocamllex [v0.20.2](https://github.com/314eter/tree-sitter-ocamllex/releases/tag/v0.20.2). Spaces are not part of the `(ocaml)` node anymore.

torhovland (approved):

torhovland:
How about adding it to the playground?

torhovland (approved):

PR #575, opened 2023-07-13
create-react-app is dead, long live Vite.

torhovland:


aspiwack:
Can I ask the motivation for such a change?

torhovland:
Well, it's in the title, but sure 🙂

You see, despite its name `create-react-app` is not just used when creating an app. It's really a framework for running a React app. And it introduces a number of dependencies, which we won't be able to upgrade as `create-react-app` stops being maintained.

Also, this switch of tools gives us a more performant development experience.

[More background here.](https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741)

web-playground/src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
ErinvanderVeen, line 1: What does this do?

web-playground/jest.config.js
```diff
@@ -1,6 +1,6 @@
-module.exports = {
+export default {
     preset: "jest-puppeteer",
     transform: {
         "^.+\\.ts?$": "ts-jest"
     },
-};
+  }
```
ErinvanderVeen, line 6: Needs a newline

web-playground/.gitignore
```diff
@@ -1,24 +1,24 @@
...
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
```
ErinvanderVeen, line 24: I don't like to have editor-specific configuration in the repository if I can help it.

ErinvanderVeen (changes requested):

web-playground/.gitignore
torhovland: This is just the default `.gitignore` that Vite created. I can of course change it, but I actually like letting team members add whatever they like to `.gitignore`, so they can use the tools they want with no risk of cluttering up the repo with files that don't belong there.

I even go one step further, and like to have shared config for various editors too, when a project demands certain config settings (which means I wouldn't actually ignore `.vscode/*`).

web-playground/src/vite-env.d.ts
torhovland: Hmm, looks like we don't need this file. Removed.

web-playground/jest.config.js
torhovland: Fixed.

ErinvanderVeen: It wasn't quite, so I force pushed a fix.

ErinvanderVeen (approved):

torhovland:
@ErinvanderVeen I was going to wait for CI before merging, but oh well 😄

web-playground/jest.config.js
torhovland: Oh, I checked in indenting on the last line 🤦‍♂️

Maybe I should start using pre-commit hooks.

Issue #571, opened 2023-07-10
Nickel formatting regression


vkleen:
The following Nickel code used to be formatted correctly according to topiary:

```nickel
{
  kind = 'ReplicationController,
} | KubernetesConfig
```

At least current HEAD 817eb4dfeb82b47038181f9d94c5741add620d60 insists on putting a newline before the contract application `|`:

```nickel
{
  kind = 'ReplicationController,
}
  | KubernetesConfig
```

This should only happen if multiple contract annotations occur, in my opinion.

ErinvanderVeen:
Thank you for the report. There is some internal discussion happening right now about how to solve this best.

This is a regression of https://github.com/tweag/topiary/pull/554

PR #576, opened 2023-07-13
Create append/prepend versions of scope captures

### LINKED ISSUE #572: Scope prepend and append
Issue #571 is best fixed by appending `open_scope` to the expression before the annotation. We should implement these new directives:

- `prepend_open_scope`
- `append_open_scope`
- `prepend_close_scope`
- `append_close_scope`

ErinvanderVeen:
Resolves #572 and unblocks #571

Creates atoms for the Scoping captures and introduces `prepend`/`append` versions of both `scope_begin` and `scope_end`.


nbacquey (approved):
Looks good to me! You managed to expand the feature set while simplifying the implementation :+1:

PR #577, opened 2023-07-13
Append scope to node before annotation

### LINKED ISSUE #571: Nickel formatting regression

ErinvanderVeen:
Resolves #571 by only checking if the annotation itself started on a new line rather than the entire node before it.

### DIFF
--- a/topiary/languages/nickel.scm
+++ b/topiary/languages/nickel.scm
@@ -274,7 +274,7 @@
 ; id | a -> a
 (
   (#scope_id! "annotations")
-  (_) @prepend_begin_scope
+  (_) @append_begin_scope
   .
   (annot) @append_end_scope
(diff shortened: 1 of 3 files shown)


nbacquey (approved):

vkleen (approved):
This solves the issue with formatting in the Nickel repo 👍

PR #578, opened 2023-07-13
Distinction between rel_operator and concat_operator

### LINKED ISSUE #541: Pipe operators aren’t stacked in OCaml
**Describe the bug**
The pipe operator `|>` is appended to the end of the line which is both unexpected and subjectively makes reading harder. Most folks expect the pipes to line up to show what the next operation is. Visually, putting the pipes at the end makes you read to the end of the line to find the operator when in many cases, you just want to know that the output of one is going to the next so you don’t have to read every implementation to the end; the pipe at the beginning lets you ‘skip’ to the next line if it’s not terribly relevant&mdash;which is likely why almost ever language with pipes line it up this way.

**To Reproduce**
```ocaml
foo
|> bar |> baz
```

formats to

```ocaml
foo |>
  bar |>
  baz
```

**Expected behavior**
```ocaml
foo
|> bar |> baz
```

formats to

```ocaml
foo
|> bar
|> baz
```

or (no strong preference either way)

```ocaml
foo
  |> bar
  |> baz
```

**Environment**
 - OS name + version: NixOS 23.11.20230627.e18dc96 (Tapir) x86_64 
 - Version of the code: topiary-cli 0.2.3

ErinvanderVeen:
This PR attempts to resolve #541

### DIFF
--- a/topiary/languages/tree-sitter-query.scm
+++ b/topiary/languages/tree-sitter-query.scm
@@ -36,6 +36,7 @@
     (anonymous_node)
     (capture)
     (grouping)
+    (list)
     (named_node)
     "."
     "("


ErinvanderVeen:
With these recent commits, that same ocaml code is now formatted as:
```ocaml
let one =
  foo
  |> bar
  >> baz
  <|> quux

let two =
  foo @
  bar @@
  baz ^^
  quux

let three =
  raise @@
    Exception
```

ErinvanderVeen:
This just leaves
```ocaml
let blit src srcoff dst dstoff len =
  len < 0 || srcoff < 0 || srcoff > src.position - len
             || dstoff < 0 || dstoff > (Bytes.length dst) - len
```
Which is currently:
```ocaml
let blit src srcoff dst dstoff len =
  len < 0
  || srcoff < 0
  || srcoff > src.position - len
  || dstoff < 0
  || dstoff
  > (Bytes.length dst) - len
```
but would ideally be:
```ocaml
let blit src srcoff dst dstoff len =
  len < 0
  || srcoff < 0
  || srcoff > src.position - len
  || dstoff < 0
  || dstoff > (Bytes.length dst) - len
```

nbacquey:
@ErinvanderVeen you can merge if my modifications make sense to you

ErinvanderVeen:
I approve and will merge this PR, but I worry a little bit about the future of Topiary with regards to precedence and associativity. This is definitely something we need to discuss and research more.

PR #586, opened 2023-07-28
Spacing fixes made by pre-commit.

torhovland:


topiary/tests/samples/input/ocaml.ml
```diff
@@ -35,11 +35,11 @@ let id1 = if true then Fun.id else function x -> x
...
-let id3 = if 
-true 
-then 
-Fun.id 
-else 
+let id3 = if
+true
+then
+Fun.id
+else
```
ErinvanderVeen, line 42: Input files should probably be excluded from the pre-commit stuff. Or at least, we should have a way to test if Topiary correctly removes trailing whitespace.

ErinvanderVeen (changes requested):

toastal:
Wouldn’t a lot of these have been caught by EditorConfig? @ErinvanderVeen shouldn’t another EditorConfig file be dropped into tests folder?

ErinvanderVeen:
> Wouldn’t a lot of these have been caught by EditorConfig? @ErinvanderVeen shouldn’t another EditorConfig file be dropped into tests folder?

Yes! We could definitely commit two editorconfigs, if that would solve the issue.

PR #587, opened 2023-07-28
Extend playground test timeout.

torhovland:
Once in a while, the CI times out on playground formatting testing. So far, it always seems fine to just re-run CI from the GitHub GUI.

But this has happened to me twice today. It's possible that the 20 second timeout is simply too little. It's also possible that we have an actual race condition bug.

I'll try extending the timeout first. If the problem does not go away, we need further investigation.


torhovland:
This PR had a test that succeeded in 23 seconds, so it seems the extending of the timeout has helped.

```
PASS e2e/sample-tester.test.ts (30.372 s)
  test all grammars with puppeteer
    ✓ can format (22812 ms)
    ✓ outputs error messages (989 ms)
```

PR #589, opened 2023-07-28
Syntax highlighting for all languages except Nickel.

### LINKED ISSUE #543: Syntax highlighting in the playground
It would be nice to have syntax highlighting of code samples (both source and formatted) and of the queries in the playground.

Considering that we already load tree-sitter grammar for our language (and for Tree-sitter query files). It should, in principle, be natural to use these grammars for syntax highlighting.

torhovland:
Fixes #543.

### DIFF
--- a/web-playground/package-lock.json
+++ b/web-playground/package-lock.json
@@ -13,28 +13,22 @@
         "react-dom": "^18.2.0"
       },
       "devDependencies": {
-        "@jest/expect": "^29.5.0",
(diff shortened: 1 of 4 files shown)


Xophmeister (approved):
This looks understandable, so I tentatively approve, to get this merged.

The output editor pane can contain a Topiary error message, which doesn't make sense to be highlighted. Would it be easy to switch to `plain_text` in such an event?

PR #593, opened 2023-08-04
CLI Changes Sub-PR: Configuration Subcommand

Xophmeister:
This forms a part of PR #583 and **must not** be merged into `main` alone. It is only the implementation of isolated functionality -- and thus not expected to necessarily pass the CI checks -- to split up the work into logical chunks.

**NOTE** Depends on #591.


topiary-cli/src/configuration.rs
```diff
@@ -1,73 +1,218 @@
...
+    .collect()
+}
+
+/// Consume configuration and collate as specified
+fn configuration_toml(
+    sources: &[ConfigSource],
+    collation: &CollationMode,
+) -> CLIResult<toml::Value> {
+    match collation {
+        CollationMode::Merge => todo!(),
```
ErinvanderVeen, line 178: I think it might be possible for merge to be `merge_toml_values` with a higher `merge_depth`.

ErinvanderVeen (changes requested):

topiary-cli/src/configuration.rs
Xophmeister: Yep: That's what I did in #596 :+1:

PR #599, opened 2023-08-18
CLI Changes Sub-PR: Formatting subcommand

Xophmeister:
This forms a part of PR #583 and **must not** be merged into `main` alone. It is only the implementation of isolated functionality -- and thus not expected to necessarily pass the CI checks -- to split up the work into logical chunks.

### DIFF
--- a/topiary-cli/src/error.rs
+++ b/topiary-cli/src/error.rs
@@ -18,6 +18,7 @@ pub enum TopiaryError {
 pub enum CLIError {
     IOError(io::Error),
     Generic(Box<dyn error::Error>),
+    Multiple,
 }
 
 /// # Safety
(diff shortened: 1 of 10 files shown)


Xophmeister:
@ErinvanderVeen It turns out that cache _was_ working, but the locking is suboptimal. I've so far been unable to fix it, so I've just left it in there, with a `FIXME` note. However, fixing the cache _shouldn't_ change `main.rs`, so once this PR is merged, you can rebase from `chris/cli-changes` to prevent being blocked.

topiary-cli/src/error.rs
```diff
@@ -18,6 +18,7 @@ pub enum TopiaryError {
 pub enum CLIError {
     IOError(io::Error),
     Generic(Box<dyn error::Error>),
+    Multiple,
```
ErinvanderVeen, line 21: Would it be possible to have `Multiple(Vec<CLIError>)`, or some other collection?

ErinvanderVeen (approved):

topiary-cli/src/error.rs
Xophmeister: Maybe... It would need to be `Multiple(Vec<TopiaryError>)` because the errors could be from the library. That shouldn't be a problem, but then what would we do with it?

I'll add this as a feature issue, along with some other thoughts I have around multiple error handling :+1:

Xophmeister: Added as #602

PR #600, opened 2023-08-22
Adds an explicit devShell to our flake

ErinvanderVeen:
Also reworks the way we call `default.nix` a little bit.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,14 @@
+# Allows `nix-shell` without having to go trough the trouble of pinning the same
+# version as is done by the flake.
+{
+  pkgs ? import <nixpkgs> {},
+  checks ? {},
+}:
+pkgs.mkShell {
+  inputsFrom = builtins.attrValues checks;
+
+  buildInputs = with pkgs; [
+    cargo-flamegraph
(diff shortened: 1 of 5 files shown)


shell.nix
```diff
@@ -0,0 +1,14 @@
...
+  pkgs ? import <nixpkgs> {},
+  checks ? {},
+}:
+pkgs.mkShell {
+  inputsFrom = builtins.attrValues checks;
+
+  buildInputs = with pkgs; [
+    cargo-flamegraph
+    rust-analyzer
+  ];
```
Xophmeister, line 13: Should things like `clippy` and `rustfmt` be mentioned here explicitly, or are they implied from the toolchain (which I presume is still a prerequisite)?

Xophmeister (approved):
I should caveat this approval with, "I am not a Nix person" :sweat_smile: This looks coherent, but I don't fully understand the consequences.

shell.nix
ErinvanderVeen: A few lines above I include everything that we use in our tests. Including rustfmt and clippy. So they are indeed implied.

Issue #606, opened 2023-08-24
Completely remove all non formatting tasks from the topiary crate


ErinvanderVeen:
We envision a future where Topiary can be used as a formatting engine in other projects (e.g. text editors, Nickel). Such use cases do not care much about out language detection and configuration parsing stuff. They will handle that on their own.

**What I would like to see**
- `topiary` should be a crate that just contains our wonderful formatting engine.
- `topiary-cli` should be a crate that handles all IO, including parsing the topiary configuration and language detection.
- `topiary-query-files` should be a crate that (behind the features flags) contains all the default query files as static string. Specifying the `nickel` and `ocaml` feature flags will export exactly those query files.

TODO:
- [x] Create `topiary-query-files` (#609)
- [ ] Move all configuration to `topiary-cli`

Xophmeister:
Resolved in #672

PR #609, opened 2023-08-29
Split of Topiary's builtin query files into their own crate

ErinvanderVeen:
This is related to #606.

This PR can only be merged after #583

TODO:
- [x] Wait for #583 
- [x] Rebase on `main`


topiary-cli/src/io.rs
```diff
@@ -11,11 +11,42 @@ use topiary::{Configuration, Language, SupportedLanguage, TopiaryQuery};
...
+    fn from(path: &PathBuf) -> Self {
+        QueryPath::Path(path.clone())
+    }
+}
+
+impl From<String> for QueryPath {
+    fn from(content: String) -> Self {
+        QueryPath::BuiltIn(content)
+    }
+}
```
Xophmeister, line 40: I wonder if this could be genericised over `AsRef<Path>`? The problem is that would also catch `String`, which is handled differently...

topiary-queries/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+]
+bash = []
+json = []
+nickel = []
+ocaml = []
+ocaml_interface = []
+ocamllex = []
+rust = []
+toml = []
+tree_sitter_query = []
```
Xophmeister, line 35: Nice :+1:

Xophmeister (approved):
I really like the idea of enabling language support (or, at least, queries[^1]) behind feature flags. I was a little dubious, at first, about baking the queries into the binary, but it works -- and I guess it's necessary for the feature flags -- so ¯\\_(ツ)_/¯

[^1]: I wonder if it would be useful to extend the features to the included grammars, as well? However, this is probably not worthwhile if "languages" are eventually going to be completely decoupled from the library...

PR #610, opened 2023-08-31
Script to update WASM grammars

ErinvanderVeen:
Updating the grammar WASM file was always a little tricky, with docker being required. This PR introduces a script (automatically added to the `devShell`) that uses `emcc` instead.


tree-sitter-ocaml_interface.wasm
nbacquey, line 1: Are we sure this file, at the project root, should be committed?

update-wasm-grammars.sh
```diff
@@ -0,0 +1,76 @@
...
+WORKDIR=$(mktemp -d)
+BLUE='\033[0;34m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+echo -e "${BLUE}Updating all Topiary grammars. This process can take a few minutes."
+
+json() {
+  echo -e "${BLUE}JSON: Fetching${NC}"
+  git clone --depth=1 https://github.com/tree-sitter/tree-sitter-json.git "${WORKDIR}/tree-sitter-json" &> /dev/null
```
nbacquey, line 16: Maybe the repositories should be fetched from a config file?
Same remark for all other functions.

Xophmeister: Making this general and driving it from config definitely seems worth doing. Using `Cargo.toml` as the config source would be pretty cute and ensure the WASM blobs match the known-to-work revisions. (Of course, some additional source of config would be needed to handle the exceptions; i.e., OCaml.)

update-wasm-grammars.sh
```diff
@@ -0,0 +1,76 @@
...
+
+tree-sitter-query() {
+  echo -e "${BLUE}Query: Fetching${NC}"
+  git clone --depth=1 https://github.com/nvim-treesitter/tree-sitter-query.git "${WORKDIR}/tree-sitter-query" &> /dev/null
+  echo -e "${BLUE}Query: Building${NC}"
+  tree-sitter build-wasm "${WORKDIR}/tree-sitter-query"
+  echo -e "${GREEN}Query: Done${NC}"
+}
+
+(trap 'kill 0' SIGINT; json & nickel & ocaml & ocamllex & rust & toml & tree-sitter-query & wait)
```
Xophmeister, line 72: TIL `kill 0` kills all processes in the group :+1: I was worried it was going to nuke the init process for a second!

update-wasm-grammars.sh
```diff
@@ -0,0 +1,76 @@
...
+
+echo -e "${BLUE}Cleanup...${NC}"
+rm -rf "${WORKDIR}"
```
Xophmeister, line 75: I'd consider putting this behind a signal trap, so the work dir also gets cleaned up in the event of failure, as well as on successful completion. Rather than at this point, where the work dir is created (L7), something like:

```bash
trap 'rm -rf "${WORKDIR}"' EXIT
```

update-wasm-grammars.sh
Xophmeister, line 72: It would be nice if this was parametrised, so you don't have to regenerate every blob each time.

Xophmeister (commented):
Looks good: I just added some "nice-to-have" suggestions and comments :slightly_smiling_face:

update-wasm-grammars.sh
ErinvanderVeen: I won't do that in this PR, because I want to move on towards more urgent stuff. Regardless, running the script only takes about a minute on my PC.

Xophmeister: Makes sense :100:

update-wasm-grammars.sh
ErinvanderVeen: It now uses the `Cargo.lock` file to select the correct grammars.

ErinvanderVeen:
> Making this general and driving it from config definitely seems worth doing. Using Cargo.toml as the config source would be pretty cute and ensure the WASM blobs match the known-to-work revisions. (Of course, some additional source of config would be needed to handle the exceptions; i.e., OCaml.)

I'm leaving that for a future (better) version.

update-wasm-grammars.sh
ErinvanderVeen: With the additional handling needed for grammars like OCaml, I'm leaving it at this for now.

nbacquey (approved):

PR #624, opened 2023-09-09
toml: allow inline comments in multiline arrays

pjjw:
previously, if you had comments on individual items in a multiline
array, e.g.

```toml
children = [
  "michael",
  "gob", # i don't care too much for this one
  "lindsay",
]
```

topiary would confusingly move the comment _below_ the element it was
placed next to-

```toml
children = [
  "michael",
  "gob",
  # i don't care too much for this one
  "lindsay",
]
```
which can destroy the intent of the comment. this rule update now
preserves inline comments and won't insert a linebreak after a separator
if followed by one.

### DIFF
--- a/topiary/languages/toml.scm
+++ b/topiary/languages/toml.scm
@@ -52,9 +52,11 @@
 
 ; Softlines. These become either a space or a newline, depending on whether we
 ; format their node as single-line or multi-line.
-[
-  ","
-] @append_spaced_softline
+(
+  "," @append_spaced_softline
+  .
+  (comment)? @do_nothing
+)
 
 ; Indent arrays. They will only be indented inmulti-line blocks.
 
--- a/topiary/tests/samples/expected/toml.toml
+++ b/topiary/tests/samples/expected/toml.toml
@@ -82,6 +82,14 @@ cve = [ "CVE-2018-13410", "foo", "CVE-2018-1340" ]
 
 comment = "this cve is only valid with attacker-controlled flags to zip"
 
+["inline.comments"]
+multiline_with_comments = [
+  "bar",
+  "blee",
+  "baz", # dont
+  "boo"
+]
+
 [dog."tater.man"]
 type.name = "pug"
 type."speed.mph" = 300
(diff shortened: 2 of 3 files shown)


topiary/tests/samples/expected/toml.toml
```diff
@@ -73,14 +73,13 @@ hosts = [
 cve = [ "CVE-2018-13410", "CVE-2018-1340" ] # garbage CVEs
 comment = "this cve is only valid with attacker-controlled flags to zip"
 
-["unzip"]
-cve = [
-  "CVE-2018-13410",
-  "CVE-2018-1340"
+["inline.comments"]
```
nbacquey, line 76: I think we should add the `inline.comments` section alongside `unzip`, instead of replacing it

topiary/tests/samples/input/toml.toml
```diff
@@ -76,13 +76,11 @@ hosts = [
...
 comment = "this cve is only valid with attacker-controlled flags to zip"
 
-["unzip"]
-cve = ["CVE-2018-13410",
-    "CVE-2018-1340"]
-    cve = ["CVE-2018-13410", "foo", "CVE-2018-1340"]
-
-comment = "this cve is only valid with attacker-controlled flags to zip"
-
+["inline.comments"]
```
nbacquey, line 79: Ditto here

nbacquey (changes requested):

topiary/languages/toml.scm
```diff
@@ -52,9 +52,14 @@
...
 ; format their node as single-line or multi-line.
-[
-  ","
-] @append_spaced_softline
+(
+  "," @append_spaced_softline
+  .
+  [
+    (comment) @do_nothing
+    (_)
```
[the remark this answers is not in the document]
pjjw: oh, thanks for the pointer! absolutely new at tree-sitter queries.

pjjw:
@nbacquey thanks for the pointers! let me know if i've addressed comments to satisfaction.

nbacquey (approved):
Looks good to me. Thanks for contributing!

topiary/tests/samples/expected/toml.toml
nbacquey: I didn't notice that you were the original author of this section ; my bad.

PR #625, opened 2023-09-09
toml: enforce final (oxford) comma in multiline arrays, remove from singleline arrays

pjjw:
this is more a preference, but it comes from a learned (python)
annoyance- if you want to keep a set of things in sorted order, then
every time the last item changes you have to go in and pick around the
commas. in python it's a lot worse due to implicit string coercion,
this is thankfully just illegal in toml, but still feels a good move for
consistency and ease of sorting.

### DIFF
--- a/topiary/languages/toml.scm
+++ b/topiary/languages/toml.scm
@@ -58,6 +58,17 @@
   (comment)? @do_nothing
 )
 
+; remove trailing comma from last element of single line array
+(array
+  ("," @delete . "]")(#single_line_only!)
+)
+
+; add trailing comma if absent to last string of multiline array
+(array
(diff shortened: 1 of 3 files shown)


nbacquey:
I have the impression this PR is somewhat mixed up with #624, could you un-entangle those two?

pjjw:
> I have the impression this PR is somewhat mixed up with #624, could you un-entangle those two?

@nbacquey it was, mostly because preserving the comment newline behavior while also adding trailing commas is an additional third thing that takes different queries. now that #624 is landed, i've rebased off main, so this should be a proper isolated diff now.

nbacquey (approved):

PR #629, opened 2023-09-14
Full-form subcommands and completion

Xophmeister:
* Move to using full-form subcommands (e.g., `format` rather than `fmt`), for clarity. (The short-forms are still valid aliases).
* Added a shell completion generation subcommand.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,16 +1,17 @@
-{
-  pkgs,
-  system,
-  advisory-db,
-  crane,
-  rust-overlay,
-  nix-filter,
-}: let
-  wasmRustVersion = "1.67.1";
+{ pkgs
+, system
+, advisory-db
+, crane
+, rust-overlay
+, nix-filter
+,
+}:
+let
+  wasmRustVersion = "1.70.0";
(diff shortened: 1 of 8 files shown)


default.nix
ErinvanderVeen, line 1: What do did you use to format this nix file?

Xophmeister: My editor is set to use `nixpkgs-fmt` for Nix files. I don't remember why it's set that way, so I assume other formatters would also work if that's preferable :sweat_smile:

Xophmeister: ...The formatters that my linter supports for Nix are: Alejandra, `nixfmt` and `nixpkgs-fmt`.

Niols: I remember using `nixfmt` everywhere in the HASG, but I don't know if it's global at Tweag or if each project does their own thing.

Niols: Actually, I think I remember a debate on this on Tweag's Slack being unconclusive.

ErinvanderVeen: This is what happens when an ecosystem has multiple formatters!

ErinvanderVeen (approved):

Issue #615, opened 2023-09-07
Publish crates to crates.io


sdankel:
**Is your feature request related to a problem? Please describe.**

I would like to use the `topiary` crate and build my own CLI for a specific language.

**Describe the solution you'd like**

It would be great if the `topiary` crate was published to crates.io so engineers can safely build on it.

**Describe alternatives you've considered**

Installing via git in Cargo.toml. Not ideal!

**Additional context**

ErinvanderVeen:
Hej! You might have noticed that the `topiary` crate [already exists](https://crates.io/crates/topiary) on crates.io. However, this project isn't seeing active development at the moment, and so we are currently in conversation with the current owner to discuss taking over ownership. Unfortunately, this might take some time.

Niols:
Some news on this: we now transferred ownership of the `topiary` crate from the previous owner to us. What is blocking now is the presence of dependencies that are not on crates.io themselves. This is currently being worked on.

sdankel:
Nice! It would make it much easier to install the topiary CLI as well.

yannham:
Well, I don't think there's any issue with publishing `topiary-cli`. Is there any reason why it's not published @Xophmeister @ErinvanderVeen or is it just an oversight?

PR #631, opened 2023-09-20
Rename topiary to topiary-core

ErinvanderVeen:
Move us forwards on #615 (publishing on crates), but not quite. Especially the `topiary-cli` crate needs a lot of work before it can be published.

Update:
We decided on the `topiary-core` crate name, without renaming the library to `topiary`. If/when we get access to the `topiary` name on crates.io, the name will be reverted to `topiary`.

For future reference:
```bash
sed -i -- 's/topiary\([-_]\)lib/topiary\1core/g' **/*.{rs,toml,md,nix,test.ts,sh}
```


yannham:
I don't know if it's any better, but just for the record, another solution is `topiary-core`, if it's a dependency of all of the other topiary crates. That what we used for Nickel and it seems to be a common practice in the Rust ecosystem

ErinvanderVeen:
> I don't know if it's any better, but just for the record, another solution is `topiary-core`, if it's a dependency of all of the other topiary crates. That what we used for Nickel and it seems to be a common practice in the Rust ecosystem

I would agree, but `topiary-lib` is one of the products that we want to ship. `-core` suggest to me that it is only meant for internal use, but published in case someone does want to use it.

I'll start a quick internal discussion before committing to something.

Niols:
Some news on this: we now have ownership of the `topiary` crate. Do we still need/want this renaming?

PR #633, opened 2023-09-22
Release 0.3.0 Dreamy Dracaena

ErinvanderVeen:
Release 0.3.0, quite a big one this time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,18 +42,36 @@ This name should be decided amongst the team before the release.
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


Xophmeister:
#637 ought to be merged first and then rebased into this PR: it fixes the TTY playground after the CLI changes.

aspiwack:
I don't think #637 should block release: the playground is mostly useful on master and isn't a release artefact I believe.

Xophmeister:
That's true; #637 is definitely not a blocker. That said, the release won't happen today -- I assume this PR has been left in draft for a reason -- and it's a small change, so it will probably end up in the release.

Xophmeister (approved):

PR #637, opened 2023-09-28
Fix terminal playground to work with new CLI

Xophmeister:


ErinvanderVeen:
Thank you!

I don't think we need `playground.sh` anymore, but let's keep it around while it still works.

ErinvanderVeen (approved):

Issue #604, opened 2023-08-24
Demote or distinguish the log messages from the Topiary library


Xophmeister:
In the Topiary CLI, the library's logging can output a lot. Because the library's logging levels can be the same as the CLI's (e.g., `debug`), this has the effect of making the logs too noisy. I propose one (or both) of the following:

1. Demote the logging level in the library for `info` and `debug` (i.e., make them `debug` and `trace`, respectively).
2. The logs indicate their source. However, both the CLI and library are named `topiary`, so they cannot be distinguished. If we rename the library, in the CLI, then it will become clear where the logs are coming from (and also possible to filter them out, if needs be).

Issue #665, opened 2023-12-21
web playground CI fails


evertedsphere:
See e.g. https://github.com/tweag/topiary/actions/runs/7275305596/job/19822959448?pr=664

Xophmeister:
Resolved in #672

Xophmeister:
Looks like the changes in #672 don't guarantee the fix... Reopening.

PR #666, opened 2023-12-21
Distinguish library and CLI log messages

evertedsphere:
Does both things suggested in #604, by
1. renaming the `topiary` CLI binary to `topiary-cli`
2. adding a `LevelFilter` specific to messages from `topiary::*` modules (i.e. the library)

In this PR in its current state, the levels are inferred heuristically from the existing command-line verbosity setting. Both the particular mapping I came up with semi-randomly and the appropriateness of doing this rather than adding a second e.g. `--lib-log-level` flag are up for debate.

Something more important stems from the fact that renaming the library in `topiary-cli/Cargo.toml` (i.e. 
```toml
topiary-core = { package = "topiary" }
```
etc) doesn't seem to work. The module names in the logs seem to depend on the names of modules as they are defined, so to pursue this route we'd have to rename the library crate itself to `topiary-core` or some such. I personally would prefer that approach, as it's idiomatic Rust and **avoids changing the name of the executable**, but it's something that needs to be discussed first and can easily be added to this PR after. If merged as it is, this PR will require changes to release workflows to account for the change in the name of the Topiary binary.


topiary-cli/src/cli.rs
```diff
@@ -184,15 +184,20 @@ pub fn get_args() -> CLIResult<Cli> {
...
+    // before any fallible operations have started.
+    //
+    // Be more strict with logs coming with the library; prefer to display less of them.
+    let (global_filter_level, library_filter_level) = match args.global.verbose {
+        0 => (LevelFilter::Error, LevelFilter::Error),
+        1 => (LevelFilter::Warn, LevelFilter::Warn),
+        2 => (LevelFilter::Info, LevelFilter::Warn),
+        3 => (LevelFilter::Debug, LevelFilter::Warn),
+        _ => (LevelFilter::Trace, LevelFilter::Trace),
+    };
```
Xophmeister, line 196: I see the logic in hiding away the library logs, but I think this might fall under "surprising behaviour" for many. There are plenty of `info!` and `debug!` calls in the library; it would be a shame to obscure them unintentionally. I doubt there's much need for tuning the CLI and library logging levels independently -- adding something like `--lib-log-level` would definitely be too much, imo -- so my vote would be to keep these in parity.

Xophmeister (commented):
> The module names in the logs seem to depend on the names of modules as they are defined, so to pursue this route we'd have to rename the library crate itself to `topiary-core` or some such. I personally would prefer that approach, as it's idiomatic Rust and **avoids changing the name of the executable**

I think I would choose this option too; having the binary named `topiary-cli` is weird for users. It's unfortunate there's no way to rename modules just for the logger's sake; I remember looking, but couldn't find anything.

topiary-cli/src/cli.rs
evertedsphere: That makes sense to me.

Xophmeister:
@ErinvanderVeen It would be a bit cheeky of me to approve this PR myself :wink: Could you give it a yea/nay, please?

Per the discussion above, I have renamed the library to `topiary-core` and made all the necessary adjustments. That allows the logging output to be distinguishable from the library (`topiary_core`) and the CLI (`topiary_cli`).

I've also taken the opportunity to rename the top-level directories to things like `core` and `cli`, to make navigation easier (for lazy people like me!)

The CI will fail, but the meat of it all works fine. The failure is per #665 and unrelated to the changes made here. I think that can be ignored for the sake of this PR.

Xophmeister:
I have reverted the short-form directory names, so this is ready to merge (after #672)

Xophmeister:
...Oh, of course, there's some CI problem :roll_eyes:

Xophmeister:
Just a simple formatting problem...fixed. _Now_ it is ready for merge, once approved.

Xophmeister:
This is ready now, I just need a :heavy_check_mark: @ErinvanderVeen 

(It's still failing #665, but we can look into that elsewhere.)

ErinvanderVeen (approved):
Good stuff!

evertedsphere:
Thanks for seeing this through.

PR #672, opened 2024-01-25
Engine separation keep playground

ErinvanderVeen:
An attempt to implement #643 without having to deprecate the playground.

The main idea here is to vendor the git crates that we absolutely depend on for the Topiary, and eventually publishing them under the `topiary-*` crate.

Since this PR is very large, I want to reason about all the changes I have made:

### Clippy CI phase not running on MacOS
Unfortunately there seems to be something in our Nix config/Cargo.toml that makes `cargo clippy` attempt to try and build the playground on MacOS. This, in combination with a undetermined bug with building the playground on MacOS, led me to disable it for now.

Building the playground on MacOS manually `nix build .#topiary-playground` works fine.

### Removing the .vscode directory
I don't think .vscode has a place in an otherwise editor agnostic repository.

### Removing the patch versions of some of our packages
Generally, Cargo will favour newer versions over older versions anyway. Assuming the packages abide to SemVer (even below 1.0) it should therefore generally not be a problem to not specify the patch level. In some cases this might ensure that we don't build duplicate versions of packages

### removing the `SupportedLanguages` enum
This was done to make separating the `topiary-config` crate easier. With eyes on the future, this will also make it easier for us to maker it easier for external contributors to add language support to Topiary.

### new DevShell
I got annoyed I could not build the Topiary Playground locally with `cargo`, so I exposed a second devShell named `wasm` to build wasm targets. Ideally, we would find a way to combine these.

### Error changes
With `topiary-config` becoming its own crate, it needed its own errors. With this, I changed some of our Errors to be more in line with what is common in the Rust ecosystem, with eyes on the `thiserror` crate in the future.

### Moving the samples and using the cli to test them
Since the core `topiary` crate no longer depends on all the `tree-sitter` grammars, this crate could no longer be responsible for the testing. All the configuration was already done in the `topiary-cli` crate, so I moved the sample testing there.

### Vendoring the `tree-sitter-*` crates
We are moving to a moment where we publish `topiary-core` on crates.io. With this move we need to publish updated version of `web-tree-sitter-sys` and `tree-sitter-facade`. We could ask for ownership of the crates, but generally, I like prefixing them with `topiary-`.

This prevents people from assuming they are useful for anything else.

We might in the future ask for ownership and maintain them, but we don't have the manpower for that right now.

### `npm update`
The puppeteer package that we use uses a package called `browsers` to install specific versions of chrome. Unfortunately, Google made the specific version of chrome that we used unavailable. This was long fixed upstream, so a simple `npm update` resolved the issue.


Xophmeister:
Presumably this supersedes #643

ErinvanderVeen:
> Presumably this supersedes #643

Absolutely, but I want to keep it until we are sure we can merge this.

Xophmeister:
0e9fba9 resolves #665? Nice :+1:

ErinvanderVeen:
> [0e9fba9](https://github.com/tweag/topiary/commit/0e9fba931ab8b7ff994f40a4638d9b052b786d70) resolves #665? Nice 👍

Sure looks like it ;)

topiary-config/src/serde.rs
```diff
@@ -0,0 +1,229 @@
...
+            "ocaml_interface" => "ocaml_interface",
+            "ocamllex" => "ocamllex",
+            "rust" => "rust",
+            "toml" => "toml",
+            "tree_sitter_query" => "query",
+            name => return Err(TopiaryConfigError::UnknownLanguage(name.to_string())),
+        };
+
+        Ok(topiary_web_tree_sitter_sys::Language::load_path(&format!(
+            "/playground/scripts/tree-sitter-{language_name}.wasm"
```
Xophmeister, line 99: I guess hardcoding the path is fine, as it matches the layout of the website/playground, which is still under control of this repo.

topiary-config/src/source.rs
```diff
@@ -0,0 +1,119 @@
+//! Configuration for Topiary can be sourced from either that which is built-in, for from disk.
```
Xophmeister, line 1:
```suggestion
//! Configuration for Topiary can be sourced from either that which is built-in, or from disk.
```

topiary-playground/src/lib.rs
```diff
@@ -1,110 +1,108 @@
-#[cfg(target_arch = "wasm32")]
-use std::sync::Mutex;
-#[cfg(target_arch = "wasm32")]
-use topiary::{formatter, Configuration, FormatterResult, Language, Operation, TopiaryQuery};
-#[cfg(target_arch = "wasm32")]
-use tree_sitter_facade::TreeSitter;
-#[cfg(target_arch = "wasm32")]
-use wasm_bindgen::prelude::*;
+/// We create a privdte module for all the wasm code, and then export every symbol in that module.
```
Xophmeister, line 1:
```suggestion
/// We create a private module for all the wasm code, and then export every symbol in that module.
```

Xophmeister (approved):

PR #682, opened 2024-02-20
Nickel: Don't prepend line break to equals signs in annotations

### LINKED ISSUE #680: Nickel: Incorrectly inserts a newline before `=` in a record field-value pair
**Describe the bug**

**To Reproduce**
I write this
```nickel
{
  config
    | {
      labels | { _ : String } = { a = "a", b = "b" },
    }
    | default
    = {},
}
```
Turns into this after a `nickel format`
```nickel
{
  config
    | {
      labels | { _ : String } = { a
          = "a", b
          = "b" },
    }
    | default
    = {},
}
```
Which then turns into this after a second `nickel format`
```
{
  config
    | {
      labels | { _ : String } = {
          a
          = "a",
          b
          = "b"
        },
    }
    | default
    = {},
}
```

**Expected behavior**
I expect no changes to happen.

**Environment**
 - Version of the code: Nickel v1.4.1

**Additional context**
This is stable
```nickel
{
  config
    | {
      labels | { _ : String } = { a = "a", b = "b" },
    }
    | default
    #
    = {},
}
```

Xophmeister:
Resolves #680

### DIFF
--- a/topiary-queries/queries/nickel.scm
+++ b/topiary-queries/queries/nickel.scm
@@ -287,8 +287,9 @@
 )
 
 (
-  (#scope_id! "annotated_assignment")
-  "=" @prepend_spaced_scoped_softline
+  (annot)
+  .
+  "=" @prepend_spaced_softline
 )
 
 ; Break a multi-line polymorphic type annotation after the type
--- a/topiary-queries/queries/nickel.scm
+++ b/topiary-queries/queries/nickel.scm
@@ -249,15 +249,6 @@
 
 ;; Annotations
 
-; We want the equals sign to be aligned with the annotations, if they
-; coexist, so create a scope that covers them both.
-(
-  (#scope_id! "annotated_assignment")
-  (annot) @prepend_begin_scope
-  .
-  "=" @append_end_scope
-)
-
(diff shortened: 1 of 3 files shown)


topiary-queries/queries/nickel.scm
```diff
@@ -249,15 +249,6 @@
...
 
-; We want the equals sign to be aligned with the annotations, if they
-; coexist, so create a scope that covers them both.
-(
-  (#scope_id! "annotated_assignment")
-  (annot) @prepend_begin_scope
-  .
-  "=" @append_end_scope
-)
-
```
yannham, line 260: So, this is just not needed at all to ensure what the coments say, that `=` and `|` are aligned? Does this impact in anyway the content of the field if it's e.g. multiline (what's after the `=`)?

Xophmeister: Right; I think when I put this together the first time, I had the belief that a scope was necessary. However, on reflection, this doesn't seem to be the case and -- indeed -- the scope is too liberal.

The input-output tests pass -- which include the Nickel stdlib -- so AFAIK, this change doesn't have a deleterious effect elsewhere.

yannham (approved):

PR #685, opened 2024-02-26
Dependency updates and fixes

Xophmeister:
* Updated to Rust 1.74, to maintain compatibility with newer dependencies
* Pinned Tree-sitter to 0.20.10 and forced `tree-sitter-json` to the same to maintain API compatibility amongst all grammars
* Pinned `tree-sitter-rust` to a known working rev
* Pinned `tree-sitter-nickel` to rev with C-based scanner
* Update `wasm-bindgen` to 0.2.91 (I'm not sure why this is needed, but it won't build without :shrug:)
* Fixed `PartialOrd` implementation in the Tree-sitter facade

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -429,6 +429,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "11157ac094ffbdde99aa67b23417ebdd801842852b500e395a45a9c0aac03e4a"
 
 [[package]]
+name = "env_filter"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a009aa4810eb158359dda09d0c87378e4bbb89b5a801f016885a4707ba24f7ea"
+dependencies = [
+ "log",
+]
+
+[[package]]
 name = "env_logger"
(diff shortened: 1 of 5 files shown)


Xophmeister:
`nix -L flake check` works on Linux. There is, however, a weird linker error on macOS, which needs to be investigated...

yannham (approved):

torhovland:
@Xophmeister The `wasm-bindgen` crate version needs to match the `wasm-bindgen-cli` tool, so I guess you needed to upgrade the crate because you upgraded the tool.

Xophmeister:
Ah, right :ok_hand: Yes, that looks to be declared in `default.nix`, but it's not pinned to any nixpkg version.

yannham:
@Xophmeister I haven't looked at your Nix config, but here is what we do to enforce that the versions stay in sync:

https://github.com/tweag/nickel/blob/20809d3161d9c0ba537d7bf871bd7684ef91b3f3/flake.nix#L84-L93

Xophmeister:
Thanks, @yannham -- I will make an issue out of this so it isn't forgotten.

PR #687, opened 2024-03-04
chore: add our scripts to our flake.nix

ErinvanderVeen:
This PR moves all of our utility scripts into their own `bin/` directory, and adds all of them to our shell, as well as our `packages` flake output.

This allows running: `nix run .#update-wasm-grammars` for instance, but also makes them easily available when in our `nix develop`-shell.


bin/default.nix
```diff
@@ -0,0 +1,54 @@
...
+    name = "generate-coverage";
+
+    runtimeInputs = with pkgs; [
+      cacert
+      grcov
+      rustup
+    ];
+
+    text = builtins.readFile ./generate-coverage.sh;
+  };
```
Xophmeister, line 13: Does this still work? I remember getting a working environment to generate the coverage data was quite tricky and it didn't play well with the normal Topiary development environment, as it needs `rustup` to install `llvm-tools-preview`.

bin/generate-coverage.sh
```diff
@@ -1,16 +1,21 @@
...
-#!nix-shell -i bash --pure --packages cacert grcov rustup
+#!/usr/bin/env bash
 #shellcheck shell=bash
 
 set -eu
 
 # Create temporary working directory
-readonly WORKING_DIR="$(mktemp --directory)"
+WORKING_DIR="$(mktemp --directory)"
+readonly WORKING_DIR
```
Xophmeister, line 8: I'm not sure I approve of this stylistic change :smile:

ErinvanderVeen: In the face of an unyielding linter, I, a seasoned programmer, was forced to my knees! "There is no other way!" the dreaded shellcheck boomed, its voice echoing through the halls of code. With a heavy heart, I surrendered, forever marked by this digital duel.

bin/default.nix
ErinvanderVeen: No, it doesn't you are right. I think I'll just remove this script in its entirety. Or refactor it perhaps when I have the time. I'll move this PR into a Draft state for now.

Thank you <3

bin/generate-coverage.sh
Xophmeister: Hehe :slightly_smiling_face: 

I'm curious which Shellcheck rule this violates, though. I've never seen it complain about that before...

bin/default.nix
Xophmeister: AFAIK, no one is using it. I think this PR is valuable, so this being broken probably shouldn't block things... Maybe just comment it out with a `FIXME`?

bin/generate-coverage.sh
ErinvanderVeen:
```
❯ nix log /nix/store/m66lfsaakg49psaq17991s13ayisbx21-generate-coverage.drv
In /nix/store/zx8hq1z1rp9nikksfdzkggmrcpqhinmk-generate-coverage/bin/generate-coverage line 14:
readonly WORKING_DIR="$(mktemp --directory)"
         ^---------^ SC2155 (warning): Declare and assign separately to avoid masking return values.

For more information:
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
```

ErinvanderVeen: Thankfully, now, it is part of our development shell, so it will be checked constantly and in a reproducible way!

PR #689, opened 2024-03-07
Support Nickel patterns

yannham:
The last (yet to be released) version of the Nickel grammar was extended with enum variants and their corresponding patterns, and patterns can now appear on the left-hand side of a match case (instead of just enum tags). This PR updates Topiary to use the latest Nickel grammar and update the Nickel queries to handle enum variants and patterns.

This PR also adds a scope for the body of a match case (the right-hand side of "=>") for consistency reason, to align with how bound expressions (right-hand side of "=") are flowed.


yannham:
The playground is failing the CI, but I don't know if it's expected these days or not.

ErinvanderVeen:
Unfortunately, we don't dynamically generate the wasm grammars (yet). This has to be done manually using the `update-wasm-grammars` script which is part of the devShell. Yes, not ideal.

Anyway, after the update, it CI seems to pass.

PR #694, opened 2024-04-04
chore: override wasm-bindgen-cli with the version from the Cargo.lock

### LINKED ISSUE #686: Pin the Nix `wasm-bindgen-cli` dependencies, so they don't drift from those in Cargo.toml
> @Xophmeister I haven't looked at your Nix config, but here is what we do to enforce that the versions stay in sync:
>
> https://github.com/tweag/nickel/blob/20809d3161d9c0ba537d7bf871bd7684ef91b3f3/flake.nix#L84-L93

_Originally posted by @yannham in https://github.com/tweag/topiary/issues/685#issuecomment-1970987105_

ErinvanderVeen:
Thanks Nickel team for the inpiration: https://github.com/tweag/nickel/blob/6e79f0db56d7a66b70a3f7c6521ab4c4a286e89d/flake.nix#L84

This should avoid problems in the future when we update wasm-bindgen.

Resolves #686

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,9 +34,23 @@
     flake-utils.lib.eachDefaultSystem (
       system:
       let
+        wasm-bindgen-cli-overlay = final: prev:
+          let
+            cargoLock = builtins.fromTOML (builtins.readFile ./Cargo.lock);
+            wasmBindgenCargoVersions = builtins.map ({ version, ... }: version) (builtins.filter ({ name, ... }: name == "wasm-bindgen") cargoLock.package);
+            wasmBindgenVersion = assert builtins.length wasmBindgenCargoVersions == 1; builtins.elemAt wasmBindgenCargoVersions 0;
+          in
+          {
+            wasm-bindgen-cli = prev.wasm-bindgen-cli.override {
+              version = wasmBindgenVersion;
+              hash = "sha256-f/RK6s12ItqKJWJlA2WtOXtwX4Y0qa8bq/JHlLTAS3c=";
+              cargoHash = "sha256-3vxVI0BhNz/9m59b+P2YEIrwGwlp7K3pyPKt4VqQuHE=";
+            };
+          };
+
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [ rust-overlay.overlays.default ];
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -34,9 +34,23 @@
...
+          let
+            cargoLock = builtins.fromTOML (builtins.readFile ./Cargo.lock);
+            wasmBindgenCargoVersions = builtins.map ({ version, ... }: version) (builtins.filter ({ name, ... }: name == "wasm-bindgen") cargoLock.package);
+            wasmBindgenVersion = assert builtins.length wasmBindgenCargoVersions == 1; builtins.elemAt wasmBindgenCargoVersions 0;
+          in
+          {
+            wasm-bindgen-cli = prev.wasm-bindgen-cli.override {
+              version = wasmBindgenVersion;
+              hash = "sha256-f/RK6s12ItqKJWJlA2WtOXtwX4Y0qa8bq/JHlLTAS3c=";
+              cargoHash = "sha256-3vxVI0BhNz/9m59b+P2YEIrwGwlp7K3pyPKt4VqQuHE=";
```
Xophmeister, line 47: Does hardcoding the hashes here pin this, regardless of what's in `Cargo.lock`? Presumably, when there's version drift, Nix will fail -- which will be caught by the CI, at the least -- when the hashes no longer match. Perhaps that's the best that can be done cleanly...

Xophmeister (approved):
I'm not really a Nix person, so I can only superficially approve. However, if it's the same code from Nickel, then that sets a good precedent :+1:

flake.nix
ErinvanderVeen: Exactly, yes! When there is a version mismatch Nix will report an error that looks like this:

```fish
error: hash mismatch in fixed-output derivation '/nix/store/zg1xxw53cm7fny593n7fy8ymns5bk78g-wasm-bindgen-cli-0.2.91.tar.gz.drv':
         specified: sha256-OdK/fVWOpbMBsl37pSWawPqpk5sePqtu1lx1UM+7c9Q=
            got:    sha256-f/RK6s12ItqKJWJlA2WtOXtwX4Y0qa8bq/JHlLTAS3c=
```

Which is trivial to resolve.

Issue #681, opened 2024-02-20
Named rules


Xophmeister:
**Is your feature request related to a problem? Please describe.**
When debugging the formatting queries, it is very difficult to understand which rules are being applied to some subtree. If rules could be given names, which are spewed out in the debug output, that should make things a lot easier.

**Describe the solution you'd like**
Perhaps using predicates, something like:

```scm
(
   (#name "space_between_things")

   (some_node)
   .
   (some_sibling) @prepend_space
)
```

Using names should be encouraged, but not mandatory. If no name is given for a rule, then one could be generated (e.g., `queries/my-language.scm:L20-25`).

**Additional context**
Given names must be unique and query parsing should probably fail when this isn't true.

Xophmeister:
In the debug output there are `pattern_index` values when there's a match. Presumably these are 0-based in the order in which they appear in the query file.

PR #697, opened 2024-04-04
Output the (column, row) of a patternmatch when outputing at the INFO level

ErinvanderVeen:
# Output the (column, row) of a patternmatch when outputing at the INFO level

Issue: #681

## Description
This is part one of issue #681, adding the row and column to the debug output while not supporting named patterns yet.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@ This name should be decided amongst the team before the release.
 
 [Full list of changes](https://github.com/tweag/topiary/compare/v0.3.0...HEAD)
 
+### Added
+- [#697](https://github.com/tweag/topiary/pull/697) Setting the log level to INFO now outputs the pattern locations in a (row, column) way.
+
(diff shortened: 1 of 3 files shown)


Xophmeister (approved):
Nice :slightly_smiling_face: 

I think I personally prefer `Ly:x` format to `(x, y)`, but that's just bike-shedding. Besides, `(x, y)` is consistent with elsewhere, so no worries!

That said, this is recalculating the coords every time it spits out logging output, which is a lot. It would definitely be worth memoising `pattern_position`, if it's straightforward to do so.

ErinvanderVeen:
> That said, this is recalculating the coords every time it spits out logging output, which is a lot. It would definitely be worth memoising `pattern_position`, if it's straightforward to do so.

Sure. Yeah. I'll do that!

Xophmeister (approved):
:heart:

Issue #634, opened 2023-09-24
languages.ncl


toastal:
**Is your feature request related to a problem? Please describe.**

Currently language configuration is done in a TOML file. With Nickel having types, wouldn’t it be a better default option?

**Describe the solution you'd like**

Read/parse a `.topiary/languages.ncl` file.

**Describe alternatives you've considered**

Run `nickel` to output TOML then run `topiary`?

**Additional context**

n/a

PR #703, opened 2024-04-26
feat: switch to nickel for configuration

ErinvanderVeen:
# languages.ncl

Issue: #634 

## Description
To support Nickel adoption, this PR switches Topiary's configuration from TOML to Nickel.

Unfortunately, this removes the collation mode.
We will support a subset of the collation mode using Nickel's merge functionality.

NOTE: Even with nickel's default features disabled, compilation of Topiary is increased by about five-fold with this PR. Don't worry if compilation suddenly takes much longer.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date


README.md
```diff
@@ -571,89 +470,36 @@ A final optional field, called `indent`, exists to define the indentation method
...
-[[language]]
-name = "demo"
-extensions = ["demo"]
-```
-
-#### Revise Mode
+To understand why, one can read the [Nickel documentation on Merging](https://nickel-lang.org/user-manual/merging).
+The short answer is that a priority must be defined. The builtin configuration
+has everything defined with priority 0. Any priority above that will replace
+any other priority. To override the entire bash configuration, use the following
```
Xophmeister, line 480:
```suggestion
any other priority. For example, to override the entire Bash configuration, use the following
```

docs/migration-0.3-0.4.md
Xophmeister, line 1: v0.4 is already released; so this would need to be `docs/migration-0.4-0.5.md`.

docs/migration-0.3-0.4.md
```diff
@@ -0,0 +1,42 @@
+# Topiary CLI Migration Guide
+Topiary 0.4 has switched away from toml to [Nickel](https://nickel-lang.org/) for its configuration language.
```
Xophmeister, line 2:
```suggestion
Topiary 0.5 has switched away from toml to [Nickel](https://nickel-lang.org/) for its configuration language.
```

docs/migration-0.3-0.4.md
```diff
@@ -0,0 +1,42 @@
...
+{
+  "bash" = {
+    extensions = ["sh", "bash"],
+  },
+
+  "rust" = {
+    extensions = ["rs"],
+    indent = "    ", # 4 spaces
+  },
+}
```
Xophmeister, line 30: In the `README` and `topiary-config/languages.ncl`, you do:

```nickel
{
  languages = {
    <LANGUAGE> = {
      extensions = ["foo"],
    }
  }
}
```

i.e., There's a top-level `languages` key. Is it needed here?

docs/migration-0.3-0.4.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+### Keeping TOML
+Nickel comes with builtin TOML importing, so if you want to keep using your old languages.toml file, you can use the following nickel expression.
```
Xophmeister, line 34:
```suggestion
Nickel comes with builtin TOML importing, so if you want to keep using your old `languages.toml` file, you can use the following Nickel expression.
```

topiary-cli/src/main.rs
```diff
@@ -143,8 +139,8 @@ async fn run() -> CLIResult<()> {
         }
 
         Commands::Config => {
-            // Output collated configuration as TOML, with annotations about how we got there
-            print!("{config}");
+            // Output collated configuration as debug
```
Xophmeister, line 142: It kinda sucks that we lose the annotations

Cargo.toml
Xophmeister, line 1: If we're losing annotations, I don't think we need `indoc` any more (L59).

topiary-config/Cargo.toml
```diff
@@ -17,6 +17,7 @@ directories.workspace = true
 indoc.workspace = true
```
Xophmeister, line 17: If we're losing annotations, I don't think we need `indoc` any more.

topiary-config/languages.ncl
Xophmeister, line 1: Maybe it would be worth using the "collapsed-form" when there are only extensions set. For example, instead of:

```nickel
bash = {
  extensions = ["sh", "bash"]
}
```

this would be cleaner:

```nickel
bash.extensions = ["sh", "bash"]
```

etc.

topiary-config/src/source.rs
Xophmeister, line 1: I kinda liked `Source::Missing` (but I'm biased :sweat_smile:). It's not a problem, of course, but I'm curious why you went to the trouble of replacing it?

Xophmeister (approved):
Looks good :+1:

I've added a few suggestions and comments. I think only one is a blocker (the migration guide should be v0.4-0.5, I think), so I'll approve.

It might also make sense to cherry-pick the changes to `bin/update-wasm-grammars.sh` and `web-playground/public/scripts/*.wasm` into their own PR, to make this one a bit cleaner.

topiary-config/src/source.rs
ErinvanderVeen: Honestly, I don't remember. I think because `Option<A>` seems to be the canonical way to achieve this in rust. It also gives you `iter`, `map`, `and_then`, etc.

Xophmeister (approved):
Looks good :+1:

PR #704, opened 2024-05-08
fix: use slice patterns over iterator

ErinvanderVeen:
# Slices for post_processing!

## Description
This PR refactors our post_process_inner function to use slice pattern matching instead.

# Motivation
Reason for this change it to make it easier to extend the match for patterns longer than just the first 2 atoms.

## Checklist
Checklist before merging:
- [ ] CHANGELOG.md updated

### DIFF
--- a/topiary-core/src/lib.rs
+++ b/topiary-core/src/lib.rs
@@ -114,6 +114,20 @@ pub enum Atom {
     },
 }
 
+impl Atom {
+    /// This function is only expected to take spaces and newlines as argument.
+    /// It defines the order Blankline > Hardline > Space > Empty.
+    pub(crate) fn dominates(&self, other: &Atom) -> bool {
+        match self {
+            Atom::Empty => false,
+            Atom::Space => matches!(other, Atom::Empty),
+            Atom::Hardline => matches!(other, Atom::Space | Atom::Empty),
+            Atom::Blankline => matches!(other, Atom::Hardline | Atom::Space | Atom::Empty),
+            _ => panic!("Unexpected character in is_dominant"),
+        }
+    }
+}
+
 /// Used in `Atom::ScopedConditional` to apply the containing Atoms only if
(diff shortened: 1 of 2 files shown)


topiary-core/src/lib.rs
```diff
@@ -114,6 +114,20 @@ pub enum Atom {
...
+    pub(crate) fn dominates(&self, other: &Atom) -> bool {
+        match self {
+            Atom::Empty => false,
+            Atom::Space => matches!(other, Atom::Empty),
+            Atom::Hardline => matches!(other, Atom::Space | Atom::Empty),
+            Atom::Blankline => matches!(other, Atom::Hardline | Atom::Space | Atom::Empty),
+            _ => panic!("Unexpected character in is_dominant"),
+        }
+    }
+}
```
Xophmeister, line 129: I like `Atom::dominates` rather than the `is_dominant` function :+1: I wonder if we could go a step further and implement `PartialOrd` (or whatever it's called), so we can do `atom1 > atom2`? Perhaps that's too magical...

Xophmeister (approved):
There's quite a lot going on here, but it looks good :+1: Maybe expanding the comments would be worthwhile, to make it a bit easier to follow -- to preserve future maintainers' sanity -- but this is not a blocker.

topiary-core/src/lib.rs
ErinvanderVeen: I think we should only implement `PartialOrd` or `Ord` for those type where the ordering is intuitive, and it's implementation obvious. I do not believe that is the case for this type, and would thus propose we keep it as is.

Although, I'm open to being convinced of the contrary ;)

Xophmeister: Yeah, I think it's in the "too magical" realm. I like it for its syntactic sugar, but it's definitely not intuitive in a general sense :sweat_smile:

PR #705, opened 2024-05-28
[Nickel] Add support for extended patterns

yannham:
# Extended patterns

This PR adds support for the last train of new pattern features in current Nickel: wildcard patterns, or-patterns, array patterns, constant patterns and pattern guards.

## Description

Update to latest Nickel tree-sitter grammar and add queries to format the new constructs.

## Checklist

Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

PR #707, opened 2024-06-05
Update WASM generation script

Xophmeister:
# Update WASM generation script

## Description

Cherry-picked from #705, to separate concerns.

## Checklist
Checklist before merging:
- ~[ ] CHANGELOG.md updated~
- ~[ ] README.md up-to-date~

### DIFF
--- a/bin/update-wasm-grammars.sh
+++ b/bin/update-wasm-grammars.sh
@@ -4,8 +4,7 @@ set -eu
 
 cd web-playground/public/scripts/
 
-# We don't use mktemp because <SOMETHING I CAN'T REMEMBER> doesn't work
-# across disparate filesystems
+# We don't use mktemp because Tree-sitter doesn't work across disparate filesystems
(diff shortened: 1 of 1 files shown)


bin/update-wasm-grammars.sh
```diff
@@ -4,9 +4,12 @@ set -eu
 
 cd web-playground/public/scripts/
 
-WORKDIR="$(mktemp -d)"
+# We don't use mktemp because <SOMETHING I CAN'T REMEMBER> doesn't work
+# across disparate filesystems
+WORKDIR=./tmp/
+mkdir -p $WORKDIR
```
Xophmeister, line 10: @ErinvanderVeen I don't remember the details :sweat_smile: Something to do with `rename`, rather than `mv`; I assume that's in the Rust code somewhere...

ErinvanderVeen: tree-sitter

<3

PR #708, opened 2024-06-05
Bump version to 0.4.1

yannham:
Prepare a release of `topiary-queries` with the latest Nickel queries to `crates.io`. For the time being, it was decided that it's ok to bump the patch version for such an update. I hesitated to do this just for `topiary-queries`, but that would mean to stop using the workspace version for `topiary-queries`, but the added risk of being out of sync (or to just forget to bump `topiary-queries`'s version during an update) isn't worth it IMHO. Instead, I think we should just accept to bump the whole workspace version.

I don't know if a GitHub release is very useful to do in this case (or a cargo release of other Topiary crates). I think it's fine to keep all the others crates at `0.4.0` on GitHub and crates.io until the next update, and just publish `topiary-queries`. But I'm happy to do otherwise.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,7 +285,7 @@ checksum = "98cc8fbded0c607b7ba9dd60cd98df59af97e84d24e49c8557331cfc26d301ce"
 
 [[package]]
 name = "client-app"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "tokio",
  "topiary-config",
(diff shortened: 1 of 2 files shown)


Xophmeister (approved):
LGTM...although note that I don't have a dog in this race, so YMMV :sweat_smile:

yannham:
Let's get a stamp of approval from @ErinvanderVeen as well, to be sure it's a consensus

ErinvanderVeen (approved):

ErinvanderVeen:
This is good for now. I think in the future we want to decouple the versions if possible.

PR #711, opened 2024-07-04
Feature Gate Grammars

### LINKED ISSUE #710: Gate (at least) Bash support behind a feature flag
We have seen a number of times that building `tree-sitter-bash` fails as it references a (somehow) invalid ref on Savannah (GNU's git service, which hosts the Bash repository). The offending commit does appear to exist, so it's not clear why this occurs.

In the past, we've managed to resolve this by bumping the `tree-sitter-bash` version. However, IIRC, we have reached the point where this is no longer possible; newer versions use TS 0.21 (see #684). It's therefore possible that this problem could reoccur and there'd be no simple fix.

Put Bash support behind a feature gate, which is disabled by default.

Remove Bash support completely, until we can move to a newer version.

**Additional context**
`topiary-queries` already uses feature flags; I believe we'd just have to remove Bash from the default features to achieve this.

We're singling out Bash because it is unstable, but perhaps there is also an argument to be had for gating _all_ experimental languages behind an off-by-default feature flag.

Xophmeister:
# Feature Gate Grammars

Resolves #710

## Description

Put each grammar and its associated dependencies behind a feature gate. By default, the supported and contributed languages will be built; experimental languages (Bash and Rust) will not.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,9 @@ experimental.
 #### Supported
 
 These formatting styles cover their target language and fulfil Topiary's
-stated design goals. They are exposed, in Topiary, through a command
-line flag.
+stated design goals. They are exposed, in Topiary, through the
+`--language` command line flag, or language detection (based on file
+extension).
 
 * [JSON]
 * [Nickel]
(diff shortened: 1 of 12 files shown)


Xophmeister:
It was all going so well, but then...

```
error: failed to parse manifest at `/home/chris/Projects/tweag/topiary/Cargo.toml`

Caused by:
  tree-sitter-bash is optional, but workspace dependencies cannot be optional
```

Xophmeister:
@yannham FYI :slightly_smiling_face:

README.md
```diff
@@ -89,22 +89,31 @@ These formatting styles cover their target language and fulfil Topiary's
...
 
 #### Experimental
 
 These languages' formatting styles are subject to change and/or not yet
-considered production-ready. They can be accessed in Topiary by
-specifying the path to their query files.
+considered production-ready. They are not built by default and are gated
+behind a feature flag (either `experimental`, for all of them, or by
+their individual name). Once included, they can be accessed in Topiary
+by specifying the path to their query files.
```
ErinvanderVeen, line 113: `cargo run -- fmt -l rust`

We don't make a distinction between experimental and non-experimental languages anymore in the code. Just as a policy point. So you can still just use `--language`, no query file needed.

ErinvanderVeen (approved):

PR #732, opened 2024-09-06
Update formatting of function application and parenthesized blocks in Nickel

yannham:
Apply the new formatting proposal detailed in https://github.com/tweag/nickel/issues/2003 for function application and parenthesized expressions in Nickel, to reduce the inflation of indentation when using higher-order functions (and more specifically contract constructors such as `std.contract.from_predicate` or `std.contract.custom`).

This PR required to change a bit the logic of the rules applying to function applications and the like. I was helped by @nbacquey through a pairing session; thanks again for your help!

## Version update

This commit also bumps the patch version of the workspace to prepare for the corresponding release of topiary-queries so that it can be published and use from the nickel-* crates.

## Tests

In addition to regression tests, this PR also updates the Nickel stdlib sample to the latest version, because the new formatting rules are more relevant to the latest additions. The result has been reviewed and diffed and the `expected` sample has been updated accordingly.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/topiary-queries/queries/nickel.scm
+++ b/topiary-queries/queries/nickel.scm
@@ -400,7 +400,7 @@
 ; and we indent it.
 ;
 ; Note that this pattern won't match the very last argument of an applicative
-; chain (the last `t2` of the rule), which needs to be handled sperately.
+; chain (the last `t2` of the rule), which needs to be handled separately.
 (
   (#scope_id! "applicative_chain")
   (applicative
--- a/topiary-queries/queries/nickel.scm
+++ b/topiary-queries/queries/nickel.scm
@@ -79,8 +79,10 @@
 )
 
 ; Don't insert spaces before the following delimiters
+;
 ; NOTE This will destroy the space in a polymorphic record tail. For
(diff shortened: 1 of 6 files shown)


Xophmeister (approved):
Looks great :+1: I particularly approve of the comments in the query file :pray:

topiary-queries/queries/nickel.scm
```diff
@@ -327,27 +376,69 @@
...
 (infix_expr
   (#scope_id! "applicative_chain")
   (applicative) @prepend_begin_scope
 ) @append_end_scope
 
+; In the the mutli-ary application case, we add a softline before each argument,
+; and we indent it.
+;
+; Note that this pattern won't match the very last argument of an applicative
+; chain (the last `t2` of the rule), which needs to be handled sperately.
```
yannham, line 403:
```suggestion
; chain (the last `t2` of the rule), which needs to be handled separately.
```

PR #747, opened 2024-09-19
Allow path for grammar source

ErinvanderVeen:
# Allow specifying grammars through the build parsers

## Description
Previously, Topiary was forced to fetch the grammar repositories and build the grammars manually. This PR adds the option to specify a path to a grammar binary file.

Additionally, it adds the `topiary-cli-nix` build output, that doesn't fetch any grammars manually.

## Checklist
Checklist before merging:
- [ ] CHANGELOG.md updated
- [ ] README.md up-to-date


README.md
```diff
@@ -489,6 +489,31 @@ A final optional field, called `indent`, exists to define the indentation method
...
+},
+```
+
+To specify a prebuilt grammar, specify the `grammar.source.path` attribute:
+```nickel
+nickel = {
+  extensions = ["ncl"],
+  grammar.source.path = "/home/erin/Projects/tweag/tree-sitter-nickel/result/parser",
+},
+```
```
Xophmeister, line 512: What does a "prebuilt grammar" look like; presumably an `.so` file? It might be useful to elaborate on this here, to make it clearer for readers.

README.md
```diff
@@ -489,6 +489,31 @@ A final optional field, called `indent`, exists to define the indentation method
...
+For usage in Nix, a `languages_nix.ncl` file is provided that specifies the
+paths of every language using the `@nickel@` syntax. These can easily be
+replaced nixpkgs' `substituteAll`.
```
Xophmeister, line 515: Are you missing a "with" here?

```suggestion
paths of every language using the `@nickel@` syntax. These can easily be
replaced with nixpkgs' `substituteAll`.
```

Xophmeister (approved):
I've made a few comments about the README, but otherwise it looks good :+1:

Although, I don't really follow the Nix-stuff, so you may want a second opinion.

yannham:
FYI, I'm trying that as part of a `flake.lock` update on the Nickel side.  Will report if there is any issue with the new `-nix` only entry.

yannham:
One major issues is that the grammar from nixpkgs aren't necessarily up-to-date, so the queries don't match. This might not be the case for more stable grammars, but for example this is the case for the Nickel grammar, to `topiary-cli-nix` binary isn't able to format Nickel currently. A dirty fix is to take the `tree-sitter-nickel` grammar as a separate argument that we can pull from the `nickel-lang/tree-sitter-nickel` flake directly, in the meantime, because I tried to update the grammar upstream directly in Nixpkgs but the process is really not obvious.

yannham:
I've pushed my work-around to use the right Nickel grammar. I don't know if this needs to be done for other grammars as well,I haven't tried them all.

Xophmeister:
@yannham Your changes look good to me insofar as I can judge, but I trust you so feel free to consider this approved if it's critical for Nickel.

yannham:
In the end I could get rid of the Nixified version of Topiary from the Nickel build altogether, so this isn't as critical as before, if you prefer to wait for another look from @ErinvanderVeen or someone else.

PR #755, opened 2024-09-30
Introduce measuring scopes

nbacquey:
## Description
This PR introduces a new feature in Topiary: measuring scopes.
They work in conjunction with "regular" custom scopes, to bind the multi-line-ness of the latter to a smaller subsection of itself.

Additionally, this PR solves a regression in OCaml introduced in #745, that couldn't be fixed without measuring scopes.

See the updated README for the precise usage of those measuring scopes.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,10 +47,11 @@ This name should be decided amongst the team before the release.
 ### Added
 - [#705](https://github.com/tweag/topiary/pull/705) Added support for Nickel 1.7 extended pattern formatting
 - [#737](https://github.com/tweag/topiary/pull/737) Added the `prefetch` command, that prefetches and caches all grammars in the current configuration
+- [#755](https://github.com/tweag/topiary/pull/755) Introduce measuring scopes, which can be used in conjunction with regular scopes to add expressivity to the formatting queries.
 
 ### Fixed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1132,6 +1132,40 @@ If we used `@append_spaced_softline` rather than
...
+Measuring scopes are opened/closed with a similar syntax as "regular" custom scopes, with any of the following tags, in conjunction with the `#scope_id!` predicate:
+
+* `@prepend_begin_measuring_scope`
+* `@append_begin_measuring_scope`
+* `@prepend_end_measuring_scope`
+* `@prepend_begin_measuring_scope`
+
+The measuring scopes behave as follows:
+* A measuring scope must always be contained in a regular custom scope with the same `#scope_id!`. There can't be two measuring scopes with the same `#scope_id!` inside the same regular custom scope.
+* If a regular custom scope contains a measuring scope, then all tags contained in the regular scope that depend on its multi-line-ness will instead depend on the multi-line-ness of the measuring scope (hence its name: the inner, measuring scope measures the multi-line-ness of the outer, regular scope).
```
aspiwack, line 1167: With this description, I don't understand why

```ocaml
foo bar (fun x ->
  qux)
```

isn't formatted as

```ocaml
foo
  bar
  (fun x ->
    qux
  )
```

I think I'm missing the point.

yannham: I believe the measuring scope is in this case `foo bar (` (or maybe just `bar (`, but the point is that it stops at `(`). Which means that the whole scope, although it spans multiple lines, is considered to be single line, and the application isn't laid out on multiple lines.

Xophmeister (approved):

PR #759, opened 2024-10-07
Function dangling on multi-line call

### LINKED ISSUE #731: Function dangling on multi-line call looks weird
As of https://github.com/tweag/topiary/commit/1720fa963fdf65e9c767edbe02132ccab4078f5a, Topiary formats:

```ocaml
foo bar baz @@ fun x ->
x
;;
foo
  bar
  baz
  @@ fun x ->
  x
```

as

```ocaml
foo bar baz @@ fun x ->
x
;;
foo
  bar
  baz @@ fun x ->
x
```

The single-line one makes sense, I believe, but the multi-line one is just wrong. I do not really have a better solution, though. Tuareg/ocp-indent indent as:

```ocaml
foo
  bar
  baz @@ fun x ->
x
```
and
```ocaml
foo
  bar
  baz
@@ fun x ->
x
```

I am thinking maybe the latter is the best way to go?

Related to https://github.com/tweag/topiary/issues/726

nbacquey:
## Description

Allows the following to be formatted as such:
```ocaml
let _ =
  foo bar baz @@ fun x ->
  x

let _ =
  foo
    bar
    baz
    @@ fun x ->
    x
```

Closes #731 

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [X] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ This name should be decided amongst the team before the release.
 
 ### Fixed
 - [#720](https://github.com/tweag/topiary/pull/720) [#722](https://github.com/tweag/topiary/pull/722) [#723](https://github.com/tweag/topiary/pull/723) [#724](https://github.com/tweag/topiary/pull/724) [#735](https://github.com/tweag/topiary/pull/735)
-[#738](https://github.com/tweag/topiary/pull/738) [#739](https://github.com/tweag/topiary/pull/739) [#745](https://github.com/tweag/topiary/pull/745) [#755](https://github.com/tweag/topiary/pull/755) Various OCaml improvements
+[#738](https://github.com/tweag/topiary/pull/738) [#739](https://github.com/tweag/topiary/pull/739) [#745](https://github.com/tweag/topiary/pull/745) [#755](https://github.com/tweag/topiary/pull/755) [#759](https://github.com/tweag/topiary/pull/759) Various OCaml improvements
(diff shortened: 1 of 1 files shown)


topiary-cli/tests/samples/expected/ocaml.ml
```diff
@@ -1296,3 +1296,15 @@ type bar = {
...
+let _ =
+  foo bar baz @@ fun x ->
+  x
+
+let _ =
+  foo
+    bar
+    baz
+    @@ fun x ->
+    x
```
aspiwack, line 1310: I think I'd have expected rather something like

```
    @@ fun x ->
      x
```

Any reason for the choice of not indenting the body further?

Xophmeister: Not an OCaml person, but I think the motivation is per [this comment](https://github.com/tweag/topiary/issues/731#issuecomment-2334077177) in #731:

> Yes, it is rather classic in OCaml to write something like
> 
> ```ocaml
> a_function @@ fun x ->
> another_function @@ fun y ->
> (x + y) / 2
> ```

i.e., I guess it's to avoid a Christmas Tree indentation situation.

aspiwack:
Ah yes… indeed. Both situations show up where you want indentation and when you don't. Maybe we'll invent two different symbols eventually, a formatter can't read our mind after all.

PR #760, opened 2024-10-08
Introduce "query_name" predicate

### LINKED ISSUE #681: Named rules

nbacquey:
## Description
This PR introduces an optional predicate `#query_name!`, that will be displayed in logs.

For instance, this query:
```scheme
(
  (#query_name! "comma spacing")
  "," @append_space
  .
  (_)
)
```
will produce the following log line:
```
[2024-10-08T15:48:13Z INFO  topiary_core::tree_sitter] Processing match of query "comma spacing": LocalQueryMatch { pattern_index: 17, captures: [ {Node "," (1,3) - (1,4)} ] } at location (286,1)
```
It also triggers an error when encountering an unknown predicate, instead of silently ignoring it (which caused problems during prior developments).

Closes #681
## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ This name should be decided amongst the team before the release.
 - [#705](https://github.com/tweag/topiary/pull/705) Added support for Nickel 1.7 extended pattern formatting
 - [#737](https://github.com/tweag/topiary/pull/737) Added the `prefetch` command, that prefetches and caches all grammars in the current configuration
 - [#755](https://github.com/tweag/topiary/pull/755) Introduce measuring scopes, which can be used in conjunction with regular scopes to add expressivity to the formatting queries.
+- [#760](https://github.com/tweag/topiary/pull/760) Introduce optional `query_name` predicate, to help with query logging and debugging.
 
 ### Fixed
 - [#720](https://github.com/tweag/topiary/pull/720) [#722](https://github.com/tweag/topiary/pull/722) [#723](https://github.com/tweag/topiary/pull/723) [#724](https://github.com/tweag/topiary/pull/724) [#735](https://github.com/tweag/topiary/pull/735)


README.md
```diff
@@ -1307,6 +1307,40 @@ The example below solves the problem of indenting function application in OCaml
...
+When the logging verbosity is set to `-vv` or higher, Topiary outputs information about which queries are matched, for instance:
+```
+[2024-10-08T15:48:13Z INFO  topiary_core::tree_sitter] Processing match: LocalQueryMatch { pattern_index: 17, captures: [ {Node "," (1,3) - (1,4)} ] } at location (286,1)
+```
+The predicate `#query_name!` takes a string argument, is optional, and can be added to any query.
+It will modify the log line to display its argument.
+
+#### Example
+
+Considering the log line above, and let us assume that the query at `location (286,1)` is:
```
Xophmeister, line 1321: The `location (286,1)` is where the query matches in the source file. It's the `pattern_index` that indicates which query has matched (IIRC: 0-based, counting from the top of the query file).

Xophmeister (approved):
Nice :+1:

I think your comment in the example (in the README) isn't right -- see comment :point_down: -- but it's been a while since I looked into this! Please could you double-check?

PR #764, opened 2024-10-10
fix: [OCaml] Relocate dangling function line break

### LINKED ISSUE #546: Hanging forms
## Is your feature request related to a problem? Please describe.

I'd _really_ like us to format, in Ocaml, multiline `@@ fun x ->` forms as follows (1):

```ocaml
somefun @@ fun x ->
  body (* of two indentation levels, I don't care about that bit *)
```

This streamlines nested-loop type code quite a lot. Compare to the current form (2):

```ocaml
somefun @@
  fun x ->
    body
```

I'd even say that the purpose of the `@@` operator is largely to be able to write code like the first example.

Before discussing further, I want to point out the core difficulty. That is we want (3):

```ocaml
somefun @@
  fun x -> body
```

to format as (1). We want to move the `fun x ->` to the line above, it makes the function multiline. So we need to format the function in the wider context of the entire infix expression.

---

The most straightforward way to do something like it, today, is to do something like

```scheme
(infix_expression
  [
    (concat_operator) ; current operator type for @@, probably should be changed
  ] @append_spaced_softline @append_indent_start
  .
  (fun_expression)? @do_nothing
)

(infix_expression
  [
    (concat_operator)
  ] @append_space
  .
  (fun_expression ; duplicates from floating fun expressions
  .
  "fun" @append_indent_start @append_indent_start
  "->" @prepend_indent_end
  (_) @append_indent_end
  .
  )
)

; and so on for the other formatting rules for "fun" and "function"
```

This is a lot of duplication. It's silly. Bad!

## Describe the solution you'd like

So in order to achieve this type of hanging forms, we need better technology.

### Is it possible with scopes?

I don't fully understand scopes yet, and it's possible that scopes are enough to achieve our desired solution with something like

```scheme
(infix_expression
  [
    (concat_operator) ; current operator type for @@, probably should be changed
  ] @append_spaced_softline @append_indent_start
  .
  (fun_expression)? @do_nothing
)

(infix_expression @begin_scope @end_scope
  [
    (concat_operator)
  ] @append_space 
  .
  (fun_expression)
  (#scope_id! "hanging_scope")
)

; and then all the `(fun_expression)` formatting rule happen in "hanging_scope"
```

(I don't know what I'm doing so I don't expect this to work first try)

### Native support

Another option is to add some special support for hanging symbols. It's a standard enough piece of formatting logic that it may be worth it. I'd imagine the whole thing, then, to just be

```scheme
(infix_expression
  [
    (concat_operator) ; current operator type for @@, probably should be changed
  ] @append_spaced_softline @append_indent_start
  .
  (fun_expression)? @hanging
)
```

The semantic of `pattern @hanging` would be that any newline immediately before `pattern` is changed to a space, and `pattern` is formatted as multiline if the enclosing query is multiline.

nbacquey:
## Description
Allows the following
```ocaml
let _ =
  somefun @@
    fun x -> body
```
to be formatted as
```ocaml
let _ =
  somefun @@ fun x ->
  body
```

Closes #546

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ This name should be decided amongst the team before the release.
 - [#760](https://github.com/tweag/topiary/pull/760) Introduce optional `query_name` predicate, to help with query logging and debugging.
 
 ### Fixed
-- [#720](https://github.com/tweag/topiary/pull/720) [#722](https://github.com/tweag/topiary/pull/722) [#723](https://github.com/tweag/topiary/pull/723) [#724](https://github.com/tweag/topiary/pull/724) [#735](https://github.com/tweag/topiary/pull/735) [#738](https://github.com/tweag/topiary/pull/738) [#739](https://github.com/tweag/topiary/pull/739) [#745](https://github.com/tweag/topiary/pull/745) [#755](https://github.com/tweag/topiary/pull/755) [#759](https://github.com/tweag/topiary/pull/759) Various OCaml improvements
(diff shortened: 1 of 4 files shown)


Xophmeister (approved):
:bow:

PR #769, opened 2024-10-17
Playground now follows its own branch

nbacquey:
## Description
This PR makes the [playground](https://topiary.tweag.io/playground/) follow its own `playground` branch, instead of `main`.

This has become necessary because some fixes we want to add to `main` break the playground, because of some upstream issues in the `tree-sitter` library. Unfortunately, we currently lack the expertise to address those issues, yet we still want to keep updating `main`.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,27 +46,27 @@ jobs:
         run: nix run .#verify-documented-usage
 
       - name: Build web playground Wasm app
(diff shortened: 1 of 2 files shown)


Xophmeister (approved):
Thanks for doing this :pray:

PR #787, opened 2024-10-30
Add documentation section about adding new languages

### LINKED ISSUE #652: The "Suggested Workflow" documentation is out-of-date
**Describe the bug**
In the README, the "Suggested Workflow" documentation no longer corresponds with what needs to be done to, e.g., add a new language. It also refers to relics that we no longer use (e.g., `(#language!)` pragma).

**To Reproduce**
Read the README.

**Expected behaviour**
The README should be aligned with the state of the code.

**Environment**
 - OS name + version: NixOS 23.05
 - Version of the code: 79b93527d9bd59533f9a79fe490567963193fafd

**Additional context**
Given the significant changes being made in #643, which are directly related to this issue, any fix should probably wait for this PR (or, better yet, be incorporated into the PR).

nbacquey:
## Description
This PR adds a documentation section about supporting a new language to Topiary, and updates the "Suggested workflow" section to reflect the latest state of the code base.

It also reinstates the option to have "verbose" tests with `--nocapture`, which was removed during #643.

Closes #652

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/topiary-cli/tests/sample-tester.rs
+++ b/topiary-cli/tests/sample-tester.rs
@@ -22,12 +22,17 @@ fn io_test(file: &str) {
 
     // Run Topiary against the staged input file
     let mut topiary = Command::cargo_bin("topiary").unwrap();
-    topiary
+    let output = topiary
         .env("TOPIARY_LANGUAGE_DIR", "../topiary-queries/queries/")
         .arg("fmt")
+        .arg("-vvv")
         .arg(&staged)
-        .assert()
-        .success();
+        .output()
+        .expect("Failed to run `topiary fmt`");
+
+    // Print the invocation output, so that it can be inspected with --nocapture
+    let output_str = String::from_utf8(output.stderr).expect("Failed to decode Topiary output");
+    println!("{output_str}");
 
     // Read the file after formatting
     let formatted = fs::read_to_string(&staged).unwrap();


README.md
```diff
@@ -1388,27 +1388,148 @@ Then the log line will become:
...
+```
+
+#### Create the query file
+```bash
+touch topiary-queries/queries/c.scm
+```
+
+#### Testing
+
+You can now check that Topiary is able to format your new language with:
```
Xophmeister, line 1420: Maybe it would be worth adding scare quotes to manage the readers expectations :sweat_smile: 

```suggestion
You can now check that Topiary is able to "format" your new language with:
```

Maybe that's too subtle... What do you think?

Xophmeister (approved):
Looks good :pray:

I made a few minor suggestions, but not enough to block.

README.md
```diff
@@ -1388,27 +1388,148 @@ Then the log line will become:
...
+
+```bash
+$ echo 'void main();' | cargo run -- format --language c
+voidmain();
+```
+
+```bash
+$ echo 'void main();' > foo.c && cargo run -- format foo.c && cat foo.c
+voidmain();
```
yannham, line 1429: Why doesn't it trigger the idempotency check? (given that the output doesn't look like valid C anymore)

yannham (approved):

README.md
Xophmeister: It is valid C...although pretty gross. `voidmain();` on its own would be an untyped forward declaration for the `voidmain` function.

PR #794, opened 2024-11-27
Bump tree-sitter to 0.24 (not a release)

### LINKED ISSUE #684: Tree-sitter grammars must maintain compatibility with specific Tree-sitter
`tree-sitter` is a direct dependency of Topiary. Each Tree-sitter grammar (e.g., `tree-sitter-json`, `tree-sitter-rust`, etc.) has a transitive dependency on `tree-sitter`. The direct dependency and transitive dependencies _must_ maintain API compatibility, otherwise Topiary won't build.

Currently, we are using Tree-sitter 0.20. v0.21 is now available and some grammars allow this, which causes Cargo to build both 0.20 and 0.21 for Topiary and subsequently fail. We can use the `[patch]` directive in our `Cargo.toml` to pin the transitive dependency (e.g., e91534b0aa).

This might be more of a "Cargo problem", than a Topiary one, but it would be nice if there were some tooling around this. Perhaps, at least, a CI step that checks there's only one version of `tree-sitter` in the `Cargo.lock`...

(#4 will definitely be impacted by this.)

ZedThree:
# Bump tree-sitter to 0.24

Closes #684 

## Description

Biggest difficulty is that `tree_sitter::QueryMatches` is now a `StreamingIterator`, returning references rather than values. This makes it very difficult to store in a new type.

Instead make `topiary_tree_sitter_facade::QueryMatch` a trait rather than a new type. This seems to be the easiest way to preserve as much of the existing structure as possible.

Not tested for the wasm32 build, so possibly some changes required there to match.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [ ] README.md up-to-date

### DIFF
--- a/topiary-config/src/language.rs
+++ b/topiary-config/src/language.rs
@@ -156,7 +156,7 @@ impl Language {
         };
 
         let language = unsafe {
-            let language_fn: Symbol<unsafe extern "C" fn() -> *const()> =
+            let language_fn: Symbol<unsafe extern "C" fn() -> *const ()> =
                 library.get(language_fn_name.as_bytes())?;
             tree_sitter_language::LanguageFn::from_raw(*language_fn)
         };
--- a/topiary-config/src/language.rs
+++ b/topiary-config/src/language.rs
@@ -156,9 +156,9 @@ impl Language {
         };
 
         let language = unsafe {
-            let language_fn: Symbol<unsafe extern "C" fn() -> tree_sitter::Language> =
+            let language_fn: Symbol<unsafe extern "C" fn() -> *const()> =
                 library.get(language_fn_name.as_bytes())?;
-            language_fn()
+            tree_sitter_language::LanguageFn::from_raw(*language_fn)
         };
         std::mem::forget(library);
         Ok(topiary_tree_sitter_facade::Language::from(language))
--- a/topiary-core/src/lib.rs
+++ b/topiary-core/src/lib.rs
@@ -188,7 +188,7 @@ pub enum Operation {
 /// let input = "[1,2]".to_string();
 /// let mut input = input.as_bytes();
 /// let mut output = Vec::new();
-/// let json = tree_sitter_json::language();
+/// let json = topiary_tree_sitter_facade::Language::from(tree_sitter_json::LANGUAGE);
 ///
 /// let mut query_file = BufReader::new(File::open("../topiary-queries/queries/json.scm").expect("query file"));
 /// let mut query_content = String::new();
--- a/topiary-core/src/lib.rs
+++ b/topiary-core/src/lib.rs
@@ -389,7 +389,7 @@ mod tests {
         let mut input = r#"{"foo":{"bar"}}"#.as_bytes();
         let mut output = Vec::new();
         let query_content = "(#language! json)";
-        let grammar = topiary_tree_sitter_facade::Language::from(tree_sitter_json::language());
+        let grammar = topiary_tree_sitter_facade::Language::from(tree_sitter_json::LANGUAGE);
         let language = Language {
             name: "json".to_owned(),
             query: TopiaryQuery::new(&grammar, query_content).unwrap(),
@@ -425,7 +425,7 @@ mod tests {
 
         let mut output = Vec::new();
         let query_content = fs::read_to_string("../topiary-queries/queries/json.scm").unwrap();
-        let grammar = tree_sitter_json::language().into();
+        let grammar = tree_sitter_json::LANGUAGE.into();
         let language = Language {
             name: "json".to_owned(),
             query: TopiaryQuery::new(&grammar, &query_content).unwrap(),
--- a/topiary-tree-sitter-facade/src/language.rs
+++ b/topiary-tree-sitter-facade/src/language.rs
@@ -80,10 +80,10 @@ mod native {
         }
     }
 
-    impl From<tree_sitter::Language> for Language {
+    impl From<tree_sitter_language::LanguageFn> for Language {
         #[inline]
-        fn from(inner: tree_sitter::Language) -> Self {
-            Self { inner }
+        fn from(inner: tree_sitter_language::LanguageFn) -> Self {
+            Self { inner: tree_sitter::Language::new(inner) }
         }
     }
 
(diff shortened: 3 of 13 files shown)


Xophmeister:
Thank you for your contribution, @ZedThree :pray:

Will bumping to Tree-sitter 0.24 affect the grammars we support that, currently, haven't themselves upgraded to 0.24 (or maybe 0.23, I think you said, was where this transitive dependency problem was dropped)? As of writing, it looks like the default branches for the following grammars use:

| Grammar | Version |
| :- | :- |
| `tree-sitter-bash` | `0.24` (dev-dependency) |
| `tree-sitter-css` | `0.24` (dev-dependency) |
| `tree-sitter-json` | `0.24` (dev-dependency) |
| `tree-sitter-nickel` | `>=0.21` |
| `tree-sitter-ocaml` | `0.24` (dev-dependency) |
| `tree-sitter-ocamllex` | `0.20` |
| `tree-sitter-query` | `0.24.3` (dev-dependency) |
| `tree-sitter-rust` | `0.24` (dev-dependency) |
| `tree-sitter-toml` | `~0.20` (repo archived) |

I suspect that the dynamic loading may have resolved this problem regardless, as `tree-sitter-ocamllex` and `tree-sitter-toml`[^1] both predate the current `tree-sitter` dependency (0.22.6) and seem to work regardless.

I would try this out myself, but unfortunately, I cannot get your fork to build:

```console
$ cargo build
+ command cargo build
   Compiling topiary-config v0.5.1 (/home/chris/Projects/tweag/topiary/forks/PlasmaFAIR/topiary-config)
error[E0277]: the trait bound `topiary_tree_sitter_facade::Language: From<tree_sitter::Language>` is not satisfied
   --> topiary-config/src/language.rs:164:12
    |
164 |         Ok(topiary_tree_sitter_facade::Language::from(language))
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<tree_sitter::Language>` is not implemented for `topiary_tree_sitter_facade::Language`
    |
    = help: the trait `From<tree_sitter_language::LanguageFn>` is implemented for `topiary_tree_sitter_facade::Language`
    = help: for that trait implementation, expected `tree_sitter_language::LanguageFn`, found `tree_sitter::Language`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `topiary-config` (lib) due to 1 previous error

$ rustc --version
rustc 1.79.0 (129f3b996 2024-06-10) (built from a source tarball)
```


[^1]: I believe `~0.20` means `>= 0.20.0 && < 0.21`, although I'm vaguely aware that "all bets are off" when dealing with `0.*` releases...so who knows!?

yannham:
There's something wrong with the CI. I wonder if it thinks it's a release because of the version number in the title?

ZedThree:
@Xophmeister Fixed! Sorry, I was just being lazy and only building `topiary-core` locally, as that was the bit I was trying to use in another project :smile: Now all the `topiary-cli` tests pass locally for me.

I'm pretty sure it's using all of the languages, so it doesn't seem to be a problem that some of the languages are on earlier tree-sitter versions? Just had a quick look into this, and the `TSLanguage` struct returned by the `tree_sitter_<language>()` contains a `version` field, and the tree-sitter library checks that the language is compatible with the library. So, it seems to run fine, and I think it should give a sensible error message if it isn't!

Xophmeister:
@yannham IIRC not all CI jobs will run against forks. I believe this is to prevent bad actors from DDOS'ing a GitHub org's CI minutes.

topiary-config/src/language.rs
```diff
@@ -156,9 +156,9 @@ impl Language {
         };
 
         let language = unsafe {
-            let language_fn: Symbol<unsafe extern "C" fn() -> tree_sitter::Language> =
+            let language_fn: Symbol<unsafe extern "C" fn() -> *const()> =
```
Xophmeister, line 159: The formatter wants this to be formatted as:

```suggestion
            let language_fn: Symbol<unsafe extern "C" fn() -> *const ()> =
```

topiary-tree-sitter-facade/src/language.rs
```diff
@@ -80,10 +80,10 @@ mod native {
         }
     }
 
-    impl From<tree_sitter::Language> for Language {
+    impl From<tree_sitter_language::LanguageFn> for Language {
         #[inline]
-        fn from(inner: tree_sitter::Language) -> Self {
-            Self { inner }
+        fn from(inner: tree_sitter_language::LanguageFn) -> Self {
+            Self { inner: tree_sitter::Language::new(inner) }
```
Xophmeister, line 86: The formatter wants this to be formatted as:

```suggestion
            Self {
                inner: tree_sitter::Language::new(inner),
            }
```

topiary-core/src/lib.rs
Xophmeister, line 1: On L191 in `topiary-core/src/lib.rs`, you'll need to change the doc-comment to the new `tree-sitter-json` API, otherwise a test fails.

```diff
@@ -188,7 +188,7 @@ pub enum Operation {
 /// let input = "[1,2]".to_string();
 /// let mut input = input.as_bytes();
 /// let mut output = Vec::new();
-/// let json = tree_sitter_json::language();
+/// let json = topiary_tree_sitter_facade::Language::from(tree_sitter_json::LANGUAGE);
 ///
 /// let mut query_file = BufReader::new(File::open("../topiary-queries/queries/json.scm").expect("query file"));
 /// let mut query_content = String::new();
```

Xophmeister (changes requested):
Thanks, @ZedThree :pray:

I've made a few suggestions where the formatter complains. Nothing significant, but it's preventing the CI tests (`nix -L flake check`) from passing locally.

Otherwise, as you say, it looks like all the other language formatters are working properly, despite being from different versions.

Xophmeister:
@ZedThree Please could you also add this change to the `CHANGELOG.md` to mention this update? You are strongly encouraged to @-mention yourself :slightly_smiling_face:

ZedThree:
Formatting done and changelog updated :+1:

ZedThree:
I figured out why your CI isn't running properly: it only runs on `push`: https://github.com/tweag/topiary/blob/4dcdc0a474003166cec8a725286d5ad1aea5bb2f/.github/workflows/ci.yml#L1

You probably want it to also run on `pull_request`. That will give you a warning and a button to push manually on PRs raised from forks.

PR #795, opened 2024-11-28
Remove `tree-sitter-ocaml` as dependency

Xophmeister:
`tree-sitter-ocaml` is no longer needed as a dependency.


Xophmeister:
`tree-sitter-ocaml` is used in the benchmarking (`topiary-core/benches/benchmark.rs`). In which case, it's worth keeping as, I suspect, benchmarking JSON formatting won't be very illuminating.

PR #797, opened 2024-11-28
Trigger CI on PR, as well as push

Xophmeister:
Also updated benchmark to use new `tree-sitter-ocaml` API.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,6 @@
-on: push
+on:
+  - push
+  - pull_request
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 on:
-  - push
-  - pull_request
+  push:
+    branches: main
+  pull_request:
+    branches: main
 
 jobs:
   build:
--- a/topiary-core/src/tree_sitter.rs
+++ b/topiary-core/src/tree_sitter.rs
@@ -1,15 +1,15 @@
+// WASM build doesn't use topiary_tree_sitter_facade::QueryMatch or
+// streaming_iterator::StreamingIterator
+#![cfg_attr(target_arch = "wasm32", allow(dead_code))]
+
 use std::{collections::HashSet, fmt::Display};
 
 use serde::Serialize;
 
-// WASM build doesn't use topiary_tree_sitter_facade::QueryMatch
-#[cfg_attr(target_arch = "wasm32", allow(dead_code))]
 use topiary_tree_sitter_facade::{
     Node, Parser, Point, Query, QueryCapture, QueryCursor, QueryMatch, QueryPredicate, Tree,
 };
 
-// WASM build doesn't use streaming_iterator::StreamingIterator
-#[cfg_attr(target_arch = "wasm32", allow(dead_code))]
 use streaming_iterator::StreamingIterator;
 
 use crate::{
--- a/topiary-core/src/tree_sitter.rs
+++ b/topiary-core/src/tree_sitter.rs
@@ -242,6 +242,7 @@ pub fn apply_query(
     let capture_names = query.query.capture_names();
 
     let mut query_matches = query.query.matches(&root, source, &mut cursor);
+    #[allow(clippy::while_let_on_iterator)] // This is not a normal iterator
     while let Some(query_match) = query_matches.next() {
         let local_captures: Vec<QueryCapture> = query_match.captures().collect();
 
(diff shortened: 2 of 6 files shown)


Xophmeister:
There are still a few broken things... I will fix and resubmit for review.

Xophmeister:
The problems appear to be with the WASM target, if I'm following the log output correctly. It complains about a bunch of stuff in `topiary-core/src/tree-sitter.rs`:

* [ ] `tree_sitter_facade::QueryMatch` is a trait for native code, but it's a struct for WASM. It needs to be imported for native, but in WASM we get a dead code warning.
* [ ] `tree_sitter_facade::Query::matches` returns a `streaming_iterator::StreamingIterator` implementation in native code, hence needing the trait imported, but an `ExactSizeIterator` implementation in WASM. Another dead code warning.
* [ ] `tree_sitter_facade::QueryMatch::pattern_index` returns a `usize` in native code and a `u32` in WASM, so we get a type mismatch error.
* [ ] `From<tree_sitter::Point>` is not implemented for `topiary_tree_sitter_facade::Point`

Xophmeister:
> * `From<tree_sitter::Point>` is not implemented for `topiary_tree_sitter_facade::Point` for WASM. (`From<topiary_web_tree_sitter_sys::Point>` is implemented instead, in WASM.)

I'm not sure how to solve this one. If I add this to the WASM module in `topiary-tree-sitter-facade/src/point.rs`:

```rust
    impl From<tree_sitter::Point> for Point {
        fn from(point: tree_sitter::Point) -> Self {
            topiary_web_tree_sitter_sys::Point::new(point.row, point.column).into()
        }
    }
```

The error changes to:

```
error[E0433]: failed to resolve: use of undeclared crate or module `tree_sitter`
   --> topiary-tree-sitter-facade/src/point.rs:121:15
    |
121 |     impl From<tree_sitter::Point> for Point {
    |               ^^^^^^^^^^^ use of undeclared crate or module `tree_sitter`
```

This is true and by design...

Xophmeister:
...fixed that by avoiding the conversion call when targetting WASM.

topiary-tree-sitter-facade/src/query_match.rs
```diff
@@ -38,8 +39,8 @@ mod wasm {
 
     impl<'tree> QueryMatch<'tree> {
         #[inline]
-        pub fn pattern_index(&self) -> u32 {
-            self.inner.pattern()
+        pub fn pattern_index(&self) -> usize {
+            self.inner.pattern().try_into().unwrap()
```
yannham, line 43: This `unwrap()` could do with a comment. I suppose it's not going to panic for any platform where word size is at least 32 bits, so most of what we care about for Topiary. It's still good to have it written down, I guess

yannham (approved):

topiary-tree-sitter-facade/src/query_match.rs
Xophmeister: Done in 5c35e051371bda5228

.github/workflows/ci.yml
```diff
@@ -1,4 +1,6 @@
-on: push
+on:
+  - push
+  - pull_request
```
avdv, line 3: Note, for pull requests this triggers duplicate builds (one for the push, another one for the pr event).

Also, it builds for any push to any branch.

Not sure if you want this, usually it is enough to only run CI for PRs targeting the main branch, and run on pushes to the main branch:

```
on:
  push:
    branches: master
  pull_request:
    branches: master
```

Just my 2 ¢

Xophmeister: Thanks @avdv :pray:

topiary-core/src/tree_sitter.rs
```diff
@@ -12,6 +11,12 @@ use crate::{
     FormatterResult,
 };
 
+#[cfg(not(target_arch = "wasm32"))]
+use streaming_iterator::StreamingIterator;
+
+#[cfg(not(target_arch = "wasm32"))]
+use topiary_tree_sitter_facade::QueryMatch;
+
```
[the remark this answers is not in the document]
Xophmeister: Right: It's dead code for WASM builds. To disable the dead code warning on these imports without triggering a "useless lint" warning, you have to use the following pragma:

```rust
#[cfg_attr(target_arch = "wasm32", allow(dead_code))]
```

I'm not sure that's clearer :sweat: However, done in 250555f039ee5fdb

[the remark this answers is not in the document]
Xophmeister: Attempt 2: Try at the module level... :x:
Attempt 3: Use `allow(unused_imports)` rather than `allow(dead_code)`... :scream_cat:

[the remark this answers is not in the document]
Xophmeister: Yippee kai yay! It worked :tada:

.github/workflows/ci.yml
```diff
@@ -1,4 +1,8 @@
-on: push
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
```
aspiwack, line 5: I don't think there's any reason not to run CI on pull requests which don't target master. Especially since there's this playground branch that I believe we still have, which is also in production.

Issue #613, opened 2023-09-06
Cut down on flake input dependencies


toastal:
**Is your feature request related to a problem? Please describe.**
When I include this flake as an input, I get a number of dependencies that I now have to manage. I don’t think these are necessary: `numtide:flake-utils`, `numtide:nix-filter` and can be inlined for less code than the import+usage+flake.lock lines. The default systems should be `nixpkgs.lib.systems.flakeExposed` & flake-utils doesn’t include some of these systems.

**Describe the solution you'd like**
Inline the for loop from `flake-utils`. Use `builtins.path { path = …; filter = path: type: …; }` to replace `nix-filter` *or* use `lib.fileset`.

**Describe alternatives you've considered**
tolerating the dependency tree exploding as each input necessitates different versions with flakes

**Additional context**

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

PR #802, opened 2024-12-10
Remove Nix dependency: flake-utils + fixups

toastal:
This dependency isn’t useful since it is just a loop, doesn’t capture all supported systems, & isn’t worth propagating it to downstream users (currently it still is required for Nickel, but that is on them). Related #613.

Also caught a typo.

Small refactors:

* using `inputs` as an attrset to more easily track dependencies (allows removing `*-input` from one of the inputs)
* expose overlays
* attrset “caches” for `nixpkgs`, `topiaryPkgs`, `binPkgs`
* use `self.packages` where possible
* don’t calculate `checksLight` for all systems

Normally removing `flake-utils` the diff results in less line of code, but refactors added a little bit.

Note: reviewing without whitespace changes makes this easier to review (nesting level changed)

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,20 +25,52 @@
       flake = false;
     };
 
-    flake-utils.url = "github:numtide/flake-utils";
-
     tree-sitter-nickel = {
       url = "github:nickel-lang/tree-sitter-nickel";
       inputs.nixpkgs.follows = "nixpkgs";
-      inputs.flake-utils.follows = "flake-utils";
     };
   };
 
   outputs = { self, nixpkgs, ... }@inputs:
-    inputs.flake-utils.lib.eachDefaultSystem (
-      system:
-      let
-        wasm-bindgen-cli-overlay = final: prev:
+    let
+      supportedSystems = nixpkgs.lib.systems.flakeExposed;
+
+      pkgsFor = nixpkgs.lib.genAttrs supportedSystems (system: rec {
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [
+            self.overlays.tree-sitter-grammars
+            self.overlays.wasm-bindgen-cli
+            inputs.rust-overlay.overlays.default
+          ];
+        };
+
+        topiaryPkgs = pkgs.callPackage ./default.nix {
+          inherit (inputs) advisory-db crane rust-overlay;
+          inherit (pkgs.tree-sitter-grammars) tree-sitter-nickel;
+          craneLib = inputs.crane.mkLib pkgs;
+        };
+
+        binPkgs = pkgs.callPackage ./bin/default.nix { };
+      });
+
+      forAllSystems = fn: nixpkgs.lib.genAttrs supportedSystems (system: fn rec {
+        inherit system;
+        inherit (pkgsFor.${system}) pkgs topiaryPkgs binPkgs;
+        inherit (pkgs) lib;
+        craneLib = inputs.crane.mkLib pkgs;
+      });
+    in
+    {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -56,15 +55,16 @@
...
-        craneLib = crane.mkLib pkgs;
+        craneLib = inputs.crane.mkLib pkgs;
 
-        tree-sitter-nickel = tree-sitter-nickel-input.packages.${system}.default;
+        tree-sitter-nickel = inputs.tree-sitter-nickel.packages.${system}.default;
 
         topiaryPkgs = pkgs.callPackage ./default.nix {
-          inherit advisory-db crane rust-overlay nix-filter craneLib tree-sitter-nickel;
+          inherit (inputs) advisory-db crane rust-overlay nix-filter tree-sitter-nickel;
+          inherit craneLib;
```
Niols: This changes the semantics of the previous code. You probably want to move `tree-sitter-nickel` on the second `inherit` line.

Niols: This will clash with #803.

Niols (commented):
The two first commits definitely bring some nice improvements to the flake; thank you for catching the typo! As for the third one, removing `flake-utils` and replacing it by repeated calls to `nixpkgs.lib.genAttrs`, I am much less convinced by the motivation and the gain in usability and readability, but it does seem clean to me.

flake.nix
Niols: I KNEW IT ;-)

yannham (commented):
LGTM. I don't know how the Topiary people feel about the `flake-utils` part - I personally don't have any strong opinion. However, in practice this part of the PR feels more guided by taste and doesn't have much impact currently, as flake-utils is still part of the inputs transitively and it's not sure that we'll get rid of it in Nickel.

@toastal: would that be ok to take out the last commit as a separate PR, where we can discuss the `flake-utils` question independently? I think the first three commits of this PR are uncontroversial and ready to merge.

Niols:
I agree with @yannham to say that the first three commit seem uncontroversial when the last one is a matter of taste (or potentially threat model). I would be up for merging the three ones today and spend a bit of time deciding what to do with the last one.

> We all remember leftpad or see these Python & npm supply chain attacks surface for not wanting to write 5 lines of code. Normalize adding a couple of line over adding an entire dependency.

This is a very one-sided story. Sure we do remember or see those, but we also know of all the bugs introduced by people deciding to reinvent the wheel (even only for 5 lines). I am not saying that this is the case here — the change looks sound to me — but it is more of a personal choice and therefore I do think that the Topiary team should make the call.

Xophmeister (approved):
I'm not a Nix person, hence deferring to @Niols and @yannham -- thank you both :pray: -- but I think a decision needs to be made on this...so I'm going to stick my neck out :see_no_evil:

I'm going to approve this, including the "opinionated" part of the PR, for the following reasons:

1. If it replaces a dependency with something straightforward -- and certainly clear (and, therefore, maintainable) -- then that seems like a good thing.
2. Given that @toastal maintains Topiary in Nixpkgs for us, if this makes that process easier for them -- or, indeed, anyone -- then that also seems like a good thing.

Niols:
Sounds good to me. The patch definitely looks safe to me; let's merge!

PR #826, opened 2025-01-09
fix(topiary-queries/scm): Fixed invalid formatting of `field_definition`s 

mkatychev:
# Fixed invalid formatting of `field_definition`s

## Description

* prevent invalid formatting of `field_definition` that followed an `identifier`
  ```scm
  (function_call function: (identifier) @function)
  ```
  Currently the line above would format to the invalid output below:
  ```scm
  (function_callfunction: (identifier) @function)
  ```
* `predicates` parameters are now separated by a `spaced_softline` allowing long predicates to be broken up
  ```scm
  (
    (identifier) @function.builtin
    (#any-of?
    @function.builtin
    "union"
    "difference"
    "intersection"
    "circle"
    "square"
    "polygon"
    "text"
    "import"
    "projection"
    "sphere"
    "cube"
    "cylinder"
    "polyhedron"
    "linear_extrude"
    "rotate_extrude"
    "surface"
    "translate"
    "rotate"
    "scale"
    "resize"
    "mirror"
    "multmatrix"
    "color"
    "offset"
    "hull"
    "minkowski")
  )
  ```
  Currently the code would be formatted into a very long predicate line:
  ```scm
  (
    (identifier) @function.builtin
    (#any-of? @function.builtin "union" "difference" "intersection" "circle" "square" "polygon" "text" "import" "projection" "sphere" "cube" "cylinder" "polyhedron" "linear_extrude" "rotate_extrude" "surface" "translate" "rotate" "scale" "resize" "mirror" "multmatrix" "color" "offset" "hull" "minkowski")
  )
  ```

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,8 +59,7 @@ This name should be decided amongst the team before the release.
 ### Fixed
 - [#779](https://github.com/tweag/topiary/pull/779) Load relevant grammars before CLI tests
 - [#799](https://github.com/tweag/topiary/pull/799) Line break after table-less pairs in TOML
-- tree-sitter-query: Long predicate parameters now retain newline separation
-- tree-sitter-query: Prevent invalid concatenation of `field_definition`s that followed an `identifier`
+- [#826](https://github.com/tweag/topiary/pull/799) Various Tree-sitter query fixes, thanks to @mkatychev
 
 ## v0.5.1 - Fragrant Frangipani - 2024-10-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ This name should be decided amongst the team before the release.
 ### Fixed
 - [#779](https://github.com/tweag/topiary/pull/779) Load relevant grammars before CLI tests
 - [#799](https://github.com/tweag/topiary/pull/799) Line break after table-less pairs in TOML
+- tree-sitter-query: Long predicate parameters now retain newline separation
+- tree-sitter-query: Prevent invalid concatenation of `field_definition`s that followed an `identifier`
 
 ## v0.5.1 - Fragrant Frangipani - 2024-10-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@ This name should be decided amongst the team before the release.
 ### Fixed
 - [#779](https://github.com/tweag/topiary/pull/779) Load relevant grammars before CLI tests
 - [#799](https://github.com/tweag/topiary/pull/799) Line break after table-less pairs in TOML
-- [#826](https://github.com/tweag/topiary/pull/799) Various Tree-sitter query fixes, thanks to @mkatychev
+- [#826](https://github.com/tweag/topiary/pull/826) Various Tree-sitter query fixes, thanks to @mkatychev
 
 ## v0.5.1 - Fragrant Frangipani - 2024-10-22
 
--- a/topiary-queries/queries/tree_sitter_query.scm
+++ b/topiary-queries/queries/tree_sitter_query.scm
@@ -52,9 +52,13 @@
 ; The grammar always includes a (parameters) child node under
 ; (predicate), even when there are none. Topiary will deal with the
 ; zero-byte nodes, so we just need to set the spacing correctly.
+; Add softlines and indentation after parameter capture groups and identifiers.
 (predicate
   (parameters
-    _ @prepend_spaced_softline
+    [(identifier) (capture)]+ @append_spaced_softline @prepend_space @append_indent_start
+    (_)* @append_spaced_softline
+    (_) @append_indent_end @append_empty_softline
+    .
   )
 )
 
--- a/topiary-queries/queries/tree_sitter_query.scm
+++ b/topiary-queries/queries/tree_sitter_query.scm
@@ -55,11 +55,15 @@
 ; Add softlines and indentation after parameter capture groups and identifiers.
 (predicate
   (parameters
-    [(identifier) (capture)]+ @append_spaced_softline @prepend_space @append_indent_start
+    [(identifier) (capture)]+ @append_spaced_softline @append_indent_start
     (_)* @append_spaced_softline
+    .
     (_) @append_indent_end @append_empty_softline
     .
   )
+)
+(predicate
+  (predicate_type) @append_space
 )
 
 ; Line breaks and indentation in multi-line lists and nodes
--- a/topiary-queries/queries/tree_sitter_query.scm
+++ b/topiary-queries/queries/tree_sitter_query.scm
@@ -38,6 +38,7 @@
     (grouping)
     (list)
     (named_node)
+    (field_definition)
     "."
     "("
     "["
@@ -53,7 +54,7 @@
 ; zero-byte nodes, so we just need to set the spacing correctly.
 (predicate
   (parameters
-    _ @prepend_space
+    _ @prepend_spaced_softline
   )
 )
 
--- a/topiary-queries/queries/tree_sitter_query.scm
+++ b/topiary-queries/queries/tree_sitter_query.scm
@@ -72,9 +72,6 @@
     (_) @prepend_space
   )
 )
-(predicate
-  (predicate_type) @append_space
-)
 
 ; Line breaks and indentation in multi-line lists and nodes
 (list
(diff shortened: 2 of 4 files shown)


mkatychev:
I wasn't sure where to put these two test cases as `topiary-cli/tests/samples/input/tree_sitter_query.scm` seems to be a carbon copy of the ocaml query

mkatychev:
> The fact that `topiary-cli/tests/samples/input/tree_sitter_query.scm` is a _de facto_ copy of the OCaml query file is because it was our biggest query file at the time. You're more than welcome to add your test cases to it: the file is never actually used to format OCaml code.

Ok, should I just add test cases to it arbitrarily. it may be better to expand test cases into their own files and have the test suite use a set of files. I suppose for now I can keep it in a single file and add corpus style comment headers do delimit various cases.

mkatychev:
I also see that topiary formats the tree-sitter s-expressions as 
```scm
(foo
  (bar)
)
```

instead of the trailing parentheses I see everywhere else:
```scm
(foo
  (bar))
```

Is that intentional?

Xophmeister:
I think this was just the style we adopted at the time and it's stuck. There was no particular reason, as far as I remember.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ This name should be decided amongst the team before the release.
 ### Fixed
 - [#779](https://github.com/tweag/topiary/pull/779) Load relevant grammars before CLI tests
 - [#799](https://github.com/tweag/topiary/pull/799) Line break after table-less pairs in TOML
+- tree-sitter-query: Long predicate parameters now retain newline separation
+- tree-sitter-query: Prevent invalid concatenation of `field_definition`s that followed an `identifier`
```
Xophmeister, line 63: Usually we bundle fixes to language formatting into one line on the changelog. As you've done the work, you'd get credit too :+1:

Something like this:

```suggestion
- [#826](https://github.com/tweag/topiary/pull/799) Various Tree-sitter query fixes, thanks to @mkatychev
```

topiary-queries/queries/tree_sitter_query.scm
```diff
@@ -53,7 +54,7 @@
 ; zero-byte nodes, so we just need to set the spacing correctly.
 (predicate
   (parameters
-    _ @prepend_space
+    _ @prepend_spaced_softline
   )
 )
```
Xophmeister, line 59: This is nice, but I think it would be worthwhile also adding indentation here as well. So, we get something like:

```scm
  (#any-of? @function.builtin
    "union"
    "difference"
    "intersection"
    "circle"
  )
```

IMO, having the capture name on the same line as the predicate "looks nice", but that's not essential.

Xophmeister (changes requested):
Thank you for your contribution :pray: 

I've made a few suggestions and I have a couple of more general comments, from the PR discussion:

1. It would be great if you could add test cases to `topiary-cli/tests/samples/{input,expected}/tree_sitter_query.scm`. Feel free to add them arbitrarily; it doesn't matter that it diverges from our OCaml rules.

2. For the time being, I think we should stick with the Algol-style indentation, rather than Lisp-style. That is:

   ```
   ; Algol-style
   (foo
     (bar
       (quux)
     )
   )

   ; Lisp-style
   (foo
     (bar
       (quux)))
   ```

   I think there's a strong argument to switch to Lisp-style, but I think that warrants an issue -- feel free to create one -- so the Topiary Team can consider it.

topiary-queries/queries/tree_sitter_query.scm
mkatychev: From the predicates section of the [book](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/3-predicates-and-directives.html):

> Predicate S-expressions start with a _predicate name_ beginning with a `#` character, and ending with a `?` character. After that, they can contain an arbitrary number of `@`-prefixed capture names or strings

I think for now I will go with the assumption that:
*  there is a readable limit to predicate tags and identifiers (such as ` (#set! injection.language "doxygen")`)
* all indentifiers and capture tags precede literals

mkatychev: I've implemented the suggestion with this output:
https://github.com/tweag/topiary/blob/2dea7e7e154bef420f6e8548b25fe5a6c4586e83/topiary-cli/tests/samples/expected/tree_sitter_query.scm#L1516-L1526

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ This name should be decided amongst the team before the release.
 ### Fixed
 - [#779](https://github.com/tweag/topiary/pull/779) Load relevant grammars before CLI tests
 - [#799](https://github.com/tweag/topiary/pull/799) Line break after table-less pairs in TOML
+- [#826](https://github.com/tweag/topiary/pull/799) Various Tree-sitter query fixes, thanks to @mkatychev
```
mkatychev, line 62:
```suggestion
- [#826](https://github.com/tweag/topiary/pull/826) Various Tree-sitter query fixes, thanks to @mkatychev
```

topiary-queries/queries/tree_sitter_query.scm
```diff
@@ -52,11 +52,19 @@
...
 (predicate
   (parameters
-    _ @prepend_spaced_softline
+    [(identifier) (capture)]+ @append_spaced_softline @append_indent_start
+    (_)* @append_spaced_softline
+    .
+    (_) @append_indent_end @append_empty_softline
+    .
   )
 )
```
Xophmeister, line 64: This is very close to what you want and, indeed, it works perfectly when there's only a single `(identifier)`/`(capture)`.

However, if there are many, we get output like this:

```
(#foo! a
  b
    c
      "d"
    )
```

I think this fixes the problem:

```scm
(predicate
  (parameters
    [(identifier) (capture)] @append_spaced_softline @append_indent_start
    .
    (string)
  ) @append_indent_end
)

(predicate
  (parameters
    (string) @append_empty_softline
  )
)

(predicate
  (parameters
    (_) @prepend_space
  )
)
```

However, note, this isn't perfect either: in the case when there are no `(identifier)`/`(capture)` nodes -- only `(string)`s -- then the multi-line-ness will not be preserved. If we're making the assumption that there will always be one `(identifier)`/`(capture)` and the `(string)`s are not mixed between, then I think that's probably fine.

Xophmeister (changes requested):
Thanks for your changes :pray:

This is very close to being correct, however there's a small bug when multiple captures/identifiers appear after the predicate in a multi-line context.

topiary-queries/queries/tree_sitter_query.scm
mkatychev: Yes, that's the assumption I went with here: https://github.com/tweag/topiary/pull/826#discussion_r1911871285

Xophmeister: Yep, I think it's a safe assumption, from what I've seen of predicates.

mkatychev: I've made the changes and updated the test cases. I was going to suggest codegolfing the first 2 s-expressions into:
```scm
(predicate
  (parameters
    [(identifier) (capture)] @append_spaced_softline @append_indent_start
    .
    (string)+ @append_empty_softline
  ) @append_indent_end
)
```
but I realized that would limit the formatting where no identifiers/captures exist within a predicate.

Xophmeister: Looks great! Thank you :pray:

Xophmeister (approved):

Xophmeister (changes requested):
Apologies; looks like I was too quick to approve!

There is one more edge case, that the CI uncovered. When a predicate has nothing after it (captures, identifiers or strings), then an errant space ends up in the output:

```diff
    (tag_specification) @prepend_delimiter
    (#delimiter! "| ") ; sic
-   (#multi_line_only!)
+   (#multi_line_only! )
  )
```

Xophmeister:
@mkatychev It's an easy fix; I'd push it myself, but I doubt I have appropriate access to your fork. This rule is both redundant and adding the errant space when there are no parameters to the predicate:

```scm
(predicate
  (predicate_type) @append_space
)
```

It can simply be removed :+1:

mkatychev:
@Xophmeister  I've definitely had maintainers [push directly to my branches before](https://github.com/IndianBoy42/tree-sitter-just/pull/184#discussion_r1845312879).

P.S.
Did this [GHA job](https://github.com/tweag/topiary/actions/runs/12775592654/job/35639610969?pr=826) really run for 24 minutes?

PR #813, opened 2024-12-23
Drop the input before writing to the output

Xophmeister:
# Drop the input before writing to the output
Resolves #812

## Description

When Topiary tries to write its output in place, it does so by writing to a temporary file that is then moved to the same location as the input file when it's complete (known as "persisting"). The hypothesis is that Windows opens an exclusive lock on the input file and therefore this process fails.

This PR causes all references to the open input file to be dropped -- and hence the file handler closed -- before persistence.

> [!NOTE]
> This PR also introduces a minimal Windows CI workflow.

## Checklist
Checklist before merging:
- [x] Changes in #830
  - [x] Rebase from `chris/grammar-build`
  - [x] Rebase once merged to `main`
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

PR #830, opened 2025-01-15
Use `tree-sitter-loader` to build grammars

### LINKED ISSUE #829: Should we shell out to `tree-sitter build` to build grammar libraries?
In `topiary-config/src/language.rs`, we have this:

https://github.com/tweag/topiary/blob/76456f203094abf0cd6b19054fa9a6ef26ded977/topiary-config/src/language.rs#L251-L351

This was a hard slog by @ErinvanderVeen to work out the correct compiler flags for Linux and macOS.

Interestingly, dynamic loading of grammars works on Windows. However, it only seems to work if the grammars are built with `tree-sitter build`; I (quite unsurprisingly) cannot get it to fly by switching to a GNU toolchain and hoping for the best! FWIW, `tree-sitter build` is doing something similar to us:

https://github.com/tree-sitter/tree-sitter/blob/3a85d4d5f3559bb0b25382b987c419b13983a630/cli/loader/src/lib.rs#L806-L881

This is part of `tree-sitter-loader` (specifically `Loader::compile_parser_at_path`), so we might not even need to shell out.

I believe this would give us full platform support without having to work it out ourselves.

Xophmeister:
# Migrate to `tree-sitter-loader` to build grammars

Resolves #829

## Description

Use `tree-sitter-loader::Loader::compile_parser_at_path` to build grammars dynamically, rather than curating our own build pipeline. This should give us cross-platform support.

## Checklist
Checklist before merging:
- [ ] CHANGELOG.md updated
- [ ] README.md up-to-date


Xophmeister:
Hmm... Well this is quite annoying.

It builds fine and `cargo test --all-features` passes with no problem, but `nix flake check` is giving weird Clippy errors in code that hasn't changed :disappointed:

Xophmeister:
It was just a WASM-thing, in the end. Thanks @yannham for the help :pray: 

As the playground has now been detached from `main`, I've also taken the liberty of removing the Clippy WASM and playground checks from the flake. Hopefully that should shed some time from the CI.

Xophmeister:
The CI results in #813 -- whose branch has been rebased off this one -- show that compiling on Windows now works (as well as Linux and macOS) :tada: Although, note, that branch's CI is currently failing for some other reason; under investigation. However, this branch is all :white_check_mark: 

This PR is ready for review and merge.

topiary-config/src/language.rs
```diff
@@ -235,117 +236,24 @@ impl GitSource {
...
-                String::from_utf8_lossy(&output.stderr),
-            )));
-        }
-
-        log::info!("{}: successfully compiled", name);
+        // Build grammar
+        log::info!("{name}: Building grammar");
+        let mut loader =
+            tree_sitter_loader::Loader::new().map_err(TopiaryConfigFetchingError::Build)?;
+        loader.debug_build(false);
```
yannham, line 250: Any reason to do a debug build  here?

yannham (approved):

topiary-config/src/language.rs
Xophmeister: AIUI, `Loader::debug_build` doesn't do anything besides set a flag. So `debug_build(false)` means "don't do a debug build".

flake.nix
```diff
@@ -106,7 +106,7 @@
       });
 
       checks = forAllSystems ({ system, pkgs, topiaryPkgs, ... }: {
-        inherit (topiaryPkgs) clippy clippy-wasm fmt topiary-core topiary-playground audit benchmark;
+        inherit (topiaryPkgs) clippy/* clippy-wasm */ fmt topiary-core/* topiary-playground */ audit benchmark;
```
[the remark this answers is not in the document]
Xophmeister: I've already merged, but I'll put in a comment in a subsequent PR :+1:

topiary-config/src/language.rs
yannham: Ah, I see. A peculiar choice of API :upside_down_face:

PR #831, opened 2025-01-15
fix(docs): fix outdated example that uses `@begin_scope` tag

mkatychev:
## Description
the current `README.md` has an example that referred to an old variant of starting scopes

## Checklist
Checklist before merging:
- [ ] CHANGELOG.md updated
- [X] README.md up-to-date

### DIFF
--- a/README.md
+++ b/README.md
@@ -1231,8 +1231,8 @@ This Tree-sitter query:
 (#language! ocaml)
 
 (parenthesized_expression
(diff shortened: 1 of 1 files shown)


mkatychev:
I don't know what plans there are for https://github.com/tweag/topiary/pull/750 but I'm assuming this change will have to be propagated if that branch is planned to be used.

Xophmeister (approved):
Good catch :+1: Thanks for this.

Issue #836, opened 2025-01-16
Add support for enforcing upper/lower case


ctdunc:
First off, thank you for developing this awesome tool. I went from fearing the process of writing a code formatter to getting a prototype up in an afternoon. 

**Is your feature request related to a problem? Please describe.**
I am trying to write a formatter for a proprietary dialect of SQL, and would like to be able to enforce that `identifiers` be either lower or upper case depending on the users configuration. As far as I can tell, this is not supported.

**Describe the solution you'd like**
Something like
```scm
(some_loud_symbol) @upper_case
(some_quiet_symbol) @lower_case
```
would take the text captured in `some_{loud,quiet}_symbol` and make it upper/lower case respectively.

**Describe alternatives you've considered**
We were using `sqlfluff` before, but it doesn't like the dsl-specific features of this language.

I would be interested in trying to contribute this if it's something that seems like a good fit for the project.

Xophmeister:
This seems like an interesting idea.

My immediate thoughts are that its usefulness will:

1. ...require languages that aren't case sensitive. These are quite rare; SQL is probably the "go to" example. These captures would therefore have limited utility.
2. ...depend on the Tree-sitter grammar. Anecdotally, many grammars don't use nodes to mark keywords, so to apply the case changing properly, one would have to target an appropriate anonymous node. In a fictitious SQL grammar, for example, we may have:

   ```scm
   (select_statement
     "select" @upper_case
   )
   ```

   However, in SQL's case, I presume the anonymous node is sufficiently specified to allow any casing in the keyword. As such, the correct query would probably be closer to:

   ```scm
   (select_statement
     .
     _ @upper_case

     (#not-eq? @upper_case "SELECT")
   )
   ```

   (Note: We have to use `@upper_case` in the predicate because there's currently no other way; see #824.)

That's not to say that these are barriers -- not at all -- but just to demonstrate that its usage mightn't be so clear-cut.

ctdunc:
Thanks! I think these barriers are probably globally solved by adding a "watch out for potential footguns" note in the documentation. 

As for the specifics:

1. I still think its useful for case sensitive languages. e.g. for python, one may want to rename constangs declared at the beginning of a module to be upper case. This use case is out of scope since youd need a way to find all instances of the variable throughout a project, but still might be useful in a `#warn!` directive. 

2. For my use case, the keywords are all their own nodes, so this is a really basic alternation of those nodes. 

[...]

PR #838, opened 2025-01-17
Add support for @upper_case @lower_case captures

ctdunc:
# Add support for @upper_case @lower_case captures.

Issue: #836 

## Description
This adds really basic support for the feature discussed in the linked issue.
I tested it with my SQL capture, and it works, but I don't know if adding SQL as a language is in scope for this PR and there are a lot of other issues with my query (it's still kinda ugly).

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ This name should be decided amongst the team before the release.
 [Full list of changes](https://github.com/tweag/topiary/compare/v0.5.1...HEAD)
 
 ### Added
-- [#838](https://github.com/tweag/topiary/pull/838) Added `@upper_case` and `@lower_case` captures by [@ctdunc](https://github.com/ctdunc/).
+- [#838](https://github.com/tweag/topiary/pull/838) Added `@upper_case` and `@lower_case` captures, thanks to @ctdunc
 - [#785](https://github.com/tweag/topiary/pull/785) Added the `coverage` command, that checks how much of the query file is used by the input.
 - [#786](https://github.com/tweag/topiary/pull/786) Re-introduce tests to check that all of the language queries are useful.
 - [#747](https://github.com/tweag/topiary/pull/747) Added support for specifying paths to prebuilt grammars in Topiary's configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ This name should be decided amongst the team before the release.
 [Full list of changes](https://github.com/tweag/topiary/compare/v0.5.1...HEAD)
 
 ### Added
+- [#838](https://github.com/tweag/topiary/pull/838) Added `@upper_case` and `@lower_case` captures by [@ctdunc](https://github.com/ctdunc/).
 - [#785](https://github.com/tweag/topiary/pull/785) Added the `coverage` command, that checks how much of the query file is used by the input.
 - [#786](https://github.com/tweag/topiary/pull/786) Re-introduce tests to check that all of the language queries are useful.
 - [#747](https://github.com/tweag/topiary/pull/747) Added support for specifying paths to prebuilt grammars in Topiary's configuration
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -9,7 +9,7 @@ use std::{
 use topiary_tree_sitter_facade::Node;
 
 use crate::{
-    tree_sitter::NodeExt, Atom, FormatterError, FormatterResult, HowCapitalize, ScopeCondition,
+    tree_sitter::NodeExt, Atom, Capitalisation, FormatterError, FormatterResult, ScopeCondition,
     ScopeInformation,
 };
 
@@ -274,18 +274,21 @@ impl AtomCollection {
                 self.prepend(Atom::Softline { spaced: true }, node, predicates);
             }
             // Skip over leaves
-            "leaf" => {}
+            "leaf" => {
+                self.prepend(Atom::CaseBegin(Capitalisation::Pass), node, predicates);
+                self.append(Atom::CaseEnd, node, predicates);
+            }
             // Deletion
             "delete" => {
                 self.prepend(Atom::DeleteBegin, node, predicates);
                 self.append(Atom::DeleteEnd, node, predicates);
             }
             "upper_case" => {
-                self.prepend(Atom::CaseBegin(HowCapitalize::UpperCase), node, predicates);
+                self.prepend(Atom::CaseBegin(Capitalisation::UpperCase), node, predicates);
                 self.append(Atom::CaseEnd, node, predicates);
             }
             "lower_case" => {
-                self.prepend(Atom::CaseBegin(HowCapitalize::LowerCase), node, predicates);
+                self.prepend(Atom::CaseBegin(Capitalisation::LowerCase), node, predicates);
                 self.append(Atom::CaseEnd, node, predicates);
             }
             // Scope manipulation
@@ -548,7 +551,7 @@ impl AtomCollection {
                 original_position: node.start_position().into(),
                 single_line_no_indent: false,
                 multi_line_indent_all: false,
-                how_capitalize: HowCapitalize::Pass,
+                capitalisation: Capitalisation::Pass,
             });
             // Mark all sub-nodes as having this node as a "leaf parent"
             self.mark_leaf_parent(node, node.id());
@@ -895,31 +898,25 @@ impl AtomCollection {
         }
     }
 
+    /// Separate post processing of capitalisation, to avoid confusion around whitespacing.
     fn post_process_capitalization(&mut self) {
-        let mut case_context: Vec<HowCapitalize> = Vec::new();
+        let mut case_context: Vec<Capitalisation> = Vec::new();
         for atom in &mut self.atoms {
             match atom {
-                Atom::CaseBegin(case) => {
-                    match case {
-                        HowCapitalize::UpperCase => {
-                            case_context.push(HowCapitalize::UpperCase);
-                            *atom = Atom::Empty;
-                            /*
-                            if lower_case_level > 0 {
-                                panic!("Cannot use @lower_case inside of a node captured by @upper_case!");
-                            }*/
-                        }
-                        HowCapitalize::LowerCase => {
-                            case_context.push(HowCapitalize::LowerCase);
-                            *atom = Atom::Empty;
-                            /*
-                            if upper_case_level > 0 {
-                                panic!("Cannot use @upper_case inside of a node captured by @lower_case!");
-                            }*/
-                        }
-                        _ => {}
+                Atom::CaseBegin(case) => match case {
+                    Capitalisation::UpperCase => {
+                        case_context.push(Capitalisation::UpperCase);
+                        *atom = Atom::Empty;
                     }
-                }
+                    Capitalisation::LowerCase => {
+                        case_context.push(Capitalisation::LowerCase);
+                        *atom = Atom::Empty;
+                    }
+                    Capitalisation::Pass => {
+                        case_context.push(Capitalisation::Pass);
+                        *atom = Atom::Empty;
+                    }
+                },
                 Atom::CaseEnd => {
                     case_context.pop();
                     *atom = Atom::Empty;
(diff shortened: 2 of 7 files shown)


topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +894,85 @@ impl AtomCollection {
...
+                            *atom = Atom::Empty;
+                            if lower_case_level > 0 {
+                                panic!("Cannot use @lower_case inside of a node captured by @upper_case!");
+                            }
+                        }
+                        HowCapitalize::LowerCase => {
+                            lower_case_level += 1;
+                            *atom = Atom::Empty;
+                            if upper_case_level > 0 {
+                                panic!("Cannot use @upper_case inside of a node captured by @lower_case!");
```
ctdunc, line 915: This is probably not ideal, but I wasn't sure what the preferred way of handling such errors is. There may also be times where users want to capitalize all but a subset of a node, which this would need to be changed to support.

topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +895,71 @@ impl AtomCollection {
...
+                        content,
+                        id,
+                        original_position,
+                        single_line_no_indent,
+                        multi_line_indent_all,
+                        ..
+                    } => {
+                        // TODO don't be stupid with derefs
+                        *atom = Atom::Leaf {
+                            content: (*content).to_string(),
```
ctdunc, line 938: the derefs here feel bad. I'm sure there's a rustier way to do this.

topiary-core/src/lib.rs
```diff
@@ -68,6 +75,7 @@ pub enum Atom {
         single_line_no_indent: bool,
         // if the leaf is multi-line, each line will be indented, not just the first
         multi_line_indent_all: bool,
+        how_capitalize: HowCapitalize,
```
Xophmeister, line 78: This is a bit of a bike-shedding comment, so feel free to take it as such :sweat_smile:

1. `how_capitalize` and `HowCapitalize` sound a bit odd to me. I would go with something like `capitalisation: Option<Capitalisation>`, where...
2. ...`Capitalisation` is an enum with two states (`Uppercase` and `Lowercase`)

topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +895,71 @@ impl AtomCollection {
         }
     }
 
+    fn post_process_capitalization(&mut self) {
```
Xophmeister, line 898: Could this function have a doc-comment?

topiary-core/src/atom_collection.rs
Xophmeister: While your use-case is restricted to SQL keywords, I agree that nested case changes are entirely plausible; that may be the easiest way to target the specific node whose case you wish to effect.

Xophmeister (changes requested):
Thanks for your contribution :pray:

My immediate concern with this is that it does not respect leaf nodes (i.e., those annotated with `@leaf`, which are meant to remain unchanged by formatting). For example:

```scheme
; test.scm
(string) @leaf

(command
  argument: (_) @prepend_space
)

(command) @upper_case
```

```console
$ topiary format -lbash -qtest.scm <<'SH'
> echo "hello"
> SH
ECHO "HELLO"
```

Topiary is meant to be general: adding `@{upper,lower}_case` already feels a bit overly specific,[^rewrite] but regardless, they should follow the other assumptions. As such, if you could modify your code to respect `@leaf` nodes, then I think it would meet that bar.

[^rewrite]:
    It occurs to me that one could implement case changes with a mixture of `@delete` and `@append_delimiter`. Granted, it's not as clean as your solution, but it's doable. Something like this _might_ work:

    ```scheme
    (select_keyword
      _ @delete @append_delimiter

      (#not-eq? @delete "SELECT")
      (#delimiter! "SELECT")
    )
    ```

    Untested, but we use similar tricks in queries for other languages.

ctdunc:
> Thanks for your contribution 🙏
> 
> My immediate concern with this is that it does not respect leaf nodes (i.e., those annotated with `@leaf`, which are meant to remain unchanged by formatting). For example:
> 
> ```scheme
> ; test.scm
> (string) @leaf
> 
> (command
>   argument: (_) @prepend_space
> )
> 
> (command) @upper_case
> ```
> 
> ```
> $ topiary format -lbash -qtest.scm <<'SH'
> > echo "hello"
> > SH
> ECHO "HELLO"
> ```
> 
> Topiary is meant to be general: adding `@{upper,lower}_case` already feels a bit overly specific,[1](#user-content-fn-rewrite-5338fc0ec3c8f848b186ccebc7cb7573) but regardless, they should follow the other assumptions. As such, if you could modify your code to respect `@leaf` nodes, then I think it would meet that bar.
> ## Footnotes
> 
>     1. It occurs to me that one could implement case changes with a mixture of `@delete` and `@append_delimiter`. Granted, it's not as clean as your solution, but it's doable. Something like this _might_ work:
>        ```
>        (select_keyword
>          _ @delete @append_delimiter
>        
>          (#not-eq? @delete "SELECT")
>          (#delimiter! "SELECT")
>        )
>        ```
>        
>        
>        Untested, but we use similar tricks in queries for other languages. [↩](#user-content-fnref-rewrite-5338fc0ec3c8f848b186ccebc7cb7573)

Will add respect for leaf nodes. I based my implementation off of `@delete`, which also does not respect `@leaf`. If there were one capture I'd expect to have higher precedence than `@leaf`, `@delete` would probably be it, but wanted to confirm that this is intended behavior.

```
; test.scm
(string) @leaf

(command
  argument: (_) @prepend_space
)

(command) @delete
``` 
Also, cannot find docs for `@leaf` in the README, which may be worth adding.


As for using `delete` and `delimiter`, I thought about this, but (A) there are over 300 keywords in the SQL grammar I am using, and (B) I would like to be able to apply this to column aliases/identifiers as well, for which the node content is not necessarily known.

topiary-core/src/lib.rs
ctdunc: Was thinking it may be useful to have the ability to camel/pascal/snake case things in the future as well, so having a `Pass` variant feels more consistent than having `Pass` as an option. Plus, it means the `match` statement handling these cases has one less level of nesting, which IMO is desirable.

Xophmeister:
> Will add respect for leaf nodes. I based my implementation off of `@delete`, which also does not respect `@leaf`. If there were one capture I'd expect to have higher precedence than `@leaf`, `@delete` would probably be it, but wanted to confirm that this is intended behavior.
> 
> ```
> ; test.scm
> (string) @leaf
> 
> (command
>   argument: (_) @prepend_space
> )
> 
> (command) @delete
> ```

Touché :wink: Yes, `@delete` will delete nodes that are marked as leaves, so it does indeed have higher precedence than `@leaf`. It wouldn't make much sense to delete nodes and somehow leave any leaf-children in the source. I suppose what I was getting at is that `@{upper,lower}_case` affecting the contents of a string is rather surprising behaviour, which `@leaf` is designed to address.

> Also, cannot find docs for `@leaf` in the README, which may be worth adding.

Excellent point :+1: I'll add an issue.

> As for using `delete` and `delimiter`, I thought about this, but (A) there are over 300 keywords in the SQL grammar I am using, and (B) I would like to be able to apply this to column aliases/identifiers as well, for which the node content is not necessarily known.

That's fair; this method is a bit unwieldy. There was talk of a more generic rewriting capture some time ago, but even that wouldn't solve your problem as effectively as your `@{upper,lower}_case` solution.

ctdunc:
I think this is ready for review again. Given the previous query you gave, the output is now:
```sh
ECHO "hello"
```

topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +896,65 @@ impl AtomCollection {
...
+                _ => match atom {
+                    Atom::Leaf {
+                        content,
+                        id,
+                        original_position,
+                        single_line_no_indent,
+                        multi_line_indent_all,
+                        ..
+                    } => {
+                        // TODO don't be stupid with derefs
```
Xophmeister, line 931: There's a `TODO` left over, here.

Xophmeister (approved):
This is great :pray: Thanks, @ctdunc 

I've approved this, but it's not ready to merge just yet. Nearly, but I have a few small requests!

1. There is a `TODO` floating around in `atom_collection.rs` that either needs to be resolved or removed. (Either way :ok_hand:)
2. A simple test for `@{upper,lower}_case` would be helpful, although I'm not sure where you'd put it (there aren't explicit tests for other capture names). As such, don't sweat on this if it doesn't make sense.
3. Please update the `CHANGELOG` and credit yourself :100:

Not for this PR, but per [this discussion](https://github.com/tweag/topiary/pull/838#discussion_r1922494826), I think your idea of adding other case modifiers (kebab, snake, etc.) is interesting and, with this contribution, will make it easy to pursue in the future. So, thank you :bow:

topiary-core/src/atom_collection.rs
ctdunc: For this one, I am honestly unsure whether this copies these values. I don't think it will ever be a significant performance issue, but wanted to get feedback on whether this is the most "idiomatic" way to do this. I am a relative newcomer to programming in Rust (from Python, where these considerations just don't exist). If you think it's fine, I'll remove the TODO.

ctdunc:
> This is great 🙏 Thanks, @ctdunc
> 
> I've approved this, but it's not ready to merge just yet. Nearly, but I have a few small requests!
> 
>     1. There is a `TODO` floating around in `atom_collection.rs` that either needs to be resolved or removed. (Either way 👌)
> 
>     2. A simple test for `@{upper,lower}_case` would be helpful, although I'm not sure where you'd put it (there aren't explicit tests for other capture names). As such, don't sweat on this if it doesn't make sense.
> 
>     3. Please update the `CHANGELOG` and credit yourself 💯
> 
> 
> Not for this PR, but per [this discussion](https://github.com/tweag/topiary/pull/838#discussion_r1922494826), I think your idea of adding other case modifiers (kebab, snake, etc.) is interesting and, with this contribution, will make it easy to pursue in the future. So, thank you 🙇
> 
> **Edit:** I'll run the CI as well, just to check everything passes...

I have tests for `@{upper,lower}_case` on my feature/sql branch, where this test will actually make sense. Not sure what other language I'd want to put it in, but I can cook one up for sure.

topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +896,62 @@ impl AtomCollection {
...
+                        id: *id,
+                        original_position: *original_position,
+                        single_line_no_indent: *single_line_no_indent,
+                        multi_line_indent_all: *multi_line_indent_all,
+                        capitalisation: case_context
+                            .last()
+                            .unwrap_or(&Capitalisation::Pass)
+                            .clone(),
+                    }
+                }
```
Xophmeister, line 941: @yannham and I both noticed that what you're effectively doing here is an in-place mutation of the `capitalisation` element of `atom` (which is an `&mut Atom`). As such, this is probably a little easier to read:

```suggestion
                Atom::Leaf { capitalisation, .. } => {
                    *capitalisation = case_context.last().unwrap_or(&Capitalisation::Pass).clone();
                }
```

topiary-core/src/atom_collection.rs
```diff
@@ -883,13 +896,62 @@ impl AtomCollection {
...
+                    }
+                    Capitalisation::LowerCase => {
+                        case_context.push(Capitalisation::LowerCase);
+                        *atom = Atom::Empty;
+                    }
+                    Capitalisation::Pass => {
+                        case_context.push(Capitalisation::Pass);
+                        *atom = Atom::Empty;
+                    }
+                },
```
Xophmeister, line 917: Would this be simpler, or is there a need to make each branch explicit?

```suggestion
                Atom::CaseBegin(case) => {
                    case_context.push(case.clone());
                    *atom = Atom::Empty;
                }
```

I don't think it matters too much that we're cloning `case`.

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ This name should be decided amongst the team before the release.
 [Full list of changes](https://github.com/tweag/topiary/compare/v0.5.1...HEAD)
 
 ### Added
+- [#838](https://github.com/tweag/topiary/pull/838) Added `@upper_case` and `@lower_case` captures by [@ctdunc](https://github.com/ctdunc/).
```
Xophmeister, line 48: Just for sake of consistency(ish):

```suggestion
- [#838](https://github.com/tweag/topiary/pull/838) Added `@upper_case` and `@lower_case` captures, thanks to @ctdunc
```

Xophmeister (commented):
I've made a few more suggestions to simplify the code :+1:

> I have tests for `@{upper,lower}_case` on my feature/sql branch, where this test will actually make sense. Not sure what other language I'd want to put it in, but I can cook one up for sure.

That's fine then; no need to bother in this PR. I was thinking about having explicit tests in the code, rather than using the IO tester. Best to hold your IO tester tests back until they make sense, rather than trying to shoehorn them into an unexpecting language!

PR #845, opened 2025-01-21
feat(query): Add OpenSCAD formatter

mkatychev:
# feat(query): Add OpenSCAD formatter

## Description
My work on supporting alternate formatters for [`openscad-lsp`](https://github.com/Leathong/openscad-LSP/pull/33) exposed a lack of idiomatic formatting for the OpenSCAD DSL with clang falling somewhat short of expectations.

The intent of this PR is to build a formatter that is as close as possible to idioms found in popular OpenSCAD libraries such as [BOSL2](https://github.com/BelfrySCAD/BOSL2) and  [standard library](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/). 

[Cheatsheet](https://openscad.org/cheatsheet/) for ref of grammar types.

This PR uses a fork of `tree-sitter-openscad` since the main repo has been
inactive for some time: https://github.com/bollian/tree-sitter-openscad/pull/18

https://github.com/tweag/topiary/blob/8809a240ef04c7b8e4f986a4148d9c45618ddd5d/topiary-config/languages.ncl#L78-L81

The changes done to `tree-sitter-openscad` have been quite large so there has
been at ton of grammar changes behind the scenes to allow these formatting queries to behave idiomatically:
https://github.com/mkatychev/tree-sitter-openscad/blob/master/CHANGELOG.md#version-060

I've done my best to document most of the queries and test in this PR but I'm sure there are things I've missed.

## Checklist
Checklist before merging:
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

PR #848, opened 2025-01-22
fix(clippy): apply `needless_lifetimes` clippy fix

mkatychev:
# fix(clippy): apply `needless_lifetimes` clippy fix

## Description
https://github.com/tweag/topiary/pull/845 [CI run](https://github.com/tweag/topiary/actions/runs/12917298008/job/36023363227?pr=845) complained about unelided lifetimes, ran `cargo clippy --fix` to solve the issue.

## Checklist
Checklist before merging:
- [ ] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/topiary-cli/src/io.rs
+++ b/topiary-cli/src/io.rs
@@ -115,7 +115,7 @@ pub struct InputFile<'cfg> {
     query: QuerySource,
 }
 
-impl<'cfg> InputFile<'cfg> {
+impl InputFile<'_> {
     /// Convert our `InputFile` into language definition values that Topiary can consume
     pub async fn to_language(&self) -> CLIResult<Language> {
         let grammar = self.language().grammar()?;
@@ -149,7 +149,7 @@ impl<'cfg> InputFile<'cfg> {
     }
 }
 
-impl<'cfg> Read for InputFile<'cfg> {
+impl Read for InputFile<'_> {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
         match &mut self.source {
             InputSource::Stdin => stdin().lock().read(buf),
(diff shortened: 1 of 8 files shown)


Xophmeister (commented):
Has this linting rule been triggered due to the change in toolchain in #845?

I think it's a positive change, so I'm inclined to accept it for that reason, but if our current toolchain doesn't pick it up, we can't enforce it. That's only a problem insofar as this might lead to drift.

yannham:
> Has this linting rule been triggered due to the change in toolchain in https://github.com/tweag/topiary/pull/845?

Yes, there is a `nix flake update` there, which updates to a more recent version of the Rust toolchain, which usually comes with a more demanding clippy. If #845 is getting merged, the CI will warn in the future. Even if it's not yet enforced by the CI, it'll be one day or the other, because we want to keep the `flake.lock` up to date as much as possible, I suppose.

Xophmeister:
Thanks, @yannham :pray:

Because the `nix flake update` is preventing the CI from running correctly in #845, what I propose to do is:

1. Accept and merge this PR.
2. Create a new PR that does a `nix flake update` only, to upgrade our toolchains and, thus, get the CI working again for #845.

If that all works out, then @mkatychev, I think you can ignore [this comment](https://github.com/tweag/topiary/pull/845#issuecomment-2609597241) I just made on #845...

Xophmeister (approved):

mkatychev:
@Xophmeister appreciate the rundown, I hadn't thought of the flake updating the toolchain to be an issue, would version pinning a rust toolchain similar to `rust-toolchain.toml` in the flake be of use?

PR #851, opened 2025-01-24
Optionally follow symlinks when specifying input files

### LINKED ISSUE #834: Topiary follows symlinks and "reifies" them, if it makes any changes
**Describe the bug**
If `topiary format` is run in a directory that contain symlinks to files that need formatting, then Topiary will format the contents of the target and then replace the symlink with that formatted target (i.e., as a regular file).

**To Reproduce**
```console
$ mkdir test
$ echo '  {  "foo":   "bar"} ' > test/foo.json
$ ln -s foo.json test/bar.json
$ ls -l test
total 4
lrwxrwxrwx 1 chris users  8 Jan 16 10:13 bar.json -> foo.json
-rw-r--r-- 1 chris users 22 Jan 16 10:13 foo.json
$ topiary format test
$ ls -l test
total 8
-rw------- 1 chris users 17 Jan 16 10:14 bar.json
-rw------- 1 chris users 17 Jan 16 10:14 foo.json
```

**Expected behavior**
Two options, perhaps controllable via a command line option:

1. Skip over symlinks (default)
2. Format any symlinks' _targets_, leaving the symlink intact

**Environment**
 - OS name + version: NixOS 24.11
 - Version of the code: 0.5.1

**Additional context**
A similar problem happens with hardlinks; that is, the link is broken. Unlike the symlink case, I'm not quite sure how this can be fixed.

Xophmeister:
# Optionally follow (and resolve) symlinks when specifying input files
Resolves #834

## Description

Provides an `-L`/`--follow-symlinks` option to `topiary format` which will follow and resolve symlinks when computing the set of input files. If this CLI option is specified, then the filesystem traversal:

* Can "break out" of the specified directories by following any directory symlinks elsewhere.
* Will resolve symlinked files to their canonical path (skipping, if the link is broken).

The deduplication now happens after the traversal, rather than before, to dedup symlinks that point to files that exist within the input file set. (This means that if you do, say, `topiary format directory directory`, then it will do twice as much work before dedup'ing, but that's an edge case that's not worth worrying about.)

## Checklist
Checklist before merging:
- [x] Depends on #849
- [x] CHANGELOG.md updated
- [x] README.md up-to-date

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,8 +211,7 @@ Arguments:
   [FILES]...
           Input files and directories (omit to read from stdin)
 
-          Language detection and query selection is automatic, mapped from file extensions defined
-          in the Topiary configuration.
+          Language detection and query selection is automatic, mapped from file extensions defined in the Topiary configuration.
 
 Options:
   -t, --tolerate-parsing-errors
@@ -222,10 +221,13 @@ Options:
           Do not check that formatting twice gives the same output
 
   -l, --language <LANGUAGE>
-          Topiary language identifier (for formatting stdin)
+          Topiary language identifier (when formatting stdin)
 
   -q, --query <QUERY>
           Topiary query file override (when formatting stdin)
+
+  -L, --follow-symlinks
+          Follow symlinks (when formatting files)
 
   -C, --configuration <CONFIGURATION>
           Configuration file
@@ -263,8 +265,8 @@ example, using Graphviz's `dot`: `topiary visualise input.ocaml | dot -T png
 ```
 Visualise the input's Tree-sitter parse tree
 
-Visualise generates a graph representation of the parse tree that can be rendered by external
-visualisation tools, such as Graphviz. By default, the output is in the DOT format.
+Visualise generates a graph representation of the parse tree that can be rendered by external visualisation tools, such as
+Graphviz. By default, the output is in the DOT format.
 
 Usage: topiary visualise [OPTIONS] <--language <LANGUAGE>|FILE>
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -248,7 +250,14 @@ the input files' extensions. To format standard input, you must specify
...
 
-Note: `fmt` is a recognised alias of the `format` subcommand.
+> [!TIP]
+> `fmt` is a recognised alias of the `format` subcommand.
+
+> [!WARNING]
+> Topiary performs in-place file formatting by writing the formatted
+> output to a temporary file and then replacing that with the original
+> input file. For files with multiple links, on filesystems where that
+> makes sense, this will break the link.
```
Xophmeister, line 260: We _could_ test for this with `std::fs::Metadata::{nlink,number_of_links}`, but stat'ing the filesystem can be pretty expensive.

Xophmeister: Hmm... On reflection, `is_symlink` and `is_dir` will necessarily be stat'ing the filesystem and I'm doing this pretty indiscriminately. Let me shore this up and add the `nlink` check :+1:

topiary-cli/src/fs.rs
```diff
@@ -0,0 +1,159 @@
...
+                log::debug!(
+                    "{} is a symlink; use --follow-symlinks to follow",
+                    file.to_string_lossy()
+                );
+                continue;
+            }
+
+            if meta.has_multiple_links() {
+                log::warn!(
+                    "Skipping {} as it has multiple links, which Topiary would break",
```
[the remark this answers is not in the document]
Xophmeister: That's a good point, which I missed by habitually adding `-vvv`. I feel like `error` is a bit extreme, but yes, it definitely should be in the default log output. I guess the compromise is increasing the level to error and making sure it's suitably documented.

Issue #868, opened 2025-02-05
Potential dynamic build problem on Windows


Xophmeister:
#859 has demonstrated that dynamic building in Windows can often fail in unusual ways. For example, this error doesn't make a huge amount of sense:

```
C:\Users\RUNNER~1\AppData\Local\Temp\.tmp2VQMTF\toml\src\parser.c : fatal error C1083: Cannot open compiler generated file: 'C:\Users\RUNNER~1\AppData\Local\Temp\.tmp2VQMTF\rust\parser.obj': Permission denied
```

This error -- which was a result of running `topiary prefetch` -- suggests that the TOML grammar couldn't be built because it was trying to link to the Rust grammar. The Rust grammar has been previously built and therefore its artefacts cleaned up. However, AFAIK, there's no dependency between the two grammars, so that's irrelevant.

Intuitively, this feels like a concurrency issue. Maybe the Windows linker keeps state, somehow, and cannot handle multiple builds concurrently :shrug:

We should look into this. (Although it may be as easy as disabling the `parallel` feature on Windows builds.)

Xophmeister:
#869 confirms that concurrent building is problematic in Windows. This issue is not solved, but we disable concurrent grammar builds on Windows to at least move forward.

PR #869, opened 2025-02-05
Investigate and fix parallel grammar builds on Windows

Xophmeister:
# Investigate and fix parallel grammar builds on Windows

References #868

## Description

For reasons, the MSVC linker does not like building multiple grammars concurrently and it hits a race condition :shrug: It's not at all clear why this happens, so the best we can do for the time being is stick a plaster over the problem: No concurrent grammar builds on Windows.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Prefetch grammars
+        run: cargo run --all-features -- prefetch
+
       - name: Run test suite
         run: cargo test --all-features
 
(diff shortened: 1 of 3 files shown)


Xophmeister:
Forcing `prefetch` to build grammars serially in Windows certainly seems to fix the problem. It's not clear why it should be a problem at all, though :shrug:

.github/workflows/ci.yml
```diff
@@ -107,6 +107,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Prefetch grammars
+        run: cargo run --all-features -- prefetch
+
```
Xophmeister, line 112: Should we keep this here? I only put it in to demonstrate the problem, but it's not necessary for the test suite. (Also, it prevents the test suite from fetching-and-building grammars; but it's the same code path, so that shouldn't matter.)

PR #871, opened 2025-02-11
Add mold linker to CI

mkatychev:
# Add mold linker to CI

## Description

This PR intents to speed up the CI process by adding the [`mold`](https://github.com/rui314/mold) linker, removing expensive (and unnecessary) dependencies, and swapping benchmark targets.

## Checklist


mkatychev:
This draft PR is meant to get some data on some CI speedups including:
- GHA `mold` linker
- nixpkgs `mold-wrapped` linker
- changes to what submodules are fetched & what is benchmarked

mkatychev:
Using the mold GHA, build times went down from [34m](https://github.com/tweag/topiary/actions/runs/13259351415/job/37012207649) to 26m https://github.com/tweag/topiary/pull/871/commits/6b57122e48bd2e3c01a1660925faa706381ee67c

topiary-core/benches/benchmark.rs
```diff
@@ -5,17 +5,17 @@ use std::io;
 use topiary_core::{formatter, Language, Operation, TopiaryQuery};
 
 async fn format() {
-    let input = fs::read_to_string("../topiary-cli/tests/samples/input/ocaml.ml").unwrap();
-    let query_content = fs::read_to_string("../topiary-queries/queries/ocaml.scm").unwrap();
-    let ocaml = tree_sitter_ocaml::LANGUAGE_OCAML;
+    let input = fs::read_to_string("../topiary-cli/tests/samples/input/json.json").unwrap();
+    let query_content = fs::read_to_string("../topiary-queries/queries/json.scm").unwrap();
+    let json = tree_sitter_json::LANGUAGE;
```
mkatychev, line 10: performance issues with the ocaml grammar previously discussed in https://github.com/tweag/topiary/pull/859 suggest that swapping the benchmark language to (anything) JSON will reduce time spent in CI.

Cargo.toml
```diff
@@ -64,9 +64,10 @@ tree-sitter-loader = "0.24"  # NOTE Align with tree-sitter
 tree-sitter-language = "0.1"
 # tree-sitter-bash = { git = "https://github.com/tree-sitter/tree-sitter-bash", rev = "1b0321ee85701d5036c334a6f04761cdc672e64c" }
 # tree-sitter-css = { git = "https://github.com/tree-sitter/tree-sitter-css.git", rev = "02b4ee757654b7d54fe35352fd8e53a8a4385d42" }
-tree-sitter-json = { git = "https://github.com/tree-sitter/tree-sitter-json.git", rev = "4d770d31f732d50d3ec373865822fbe659e47c75" }
+tree-sitter-json = "0.24"
+# tree-sitter-json = { git = "https://github.com/tree-sitter/tree-sitter-json.git", rev = "4d770d31f732d50d3ec373865822fbe659e47c75" }
 # tree-sitter-nickel = { git = "https://github.com/nickel-lang/tree-sitter-nickel", rev = "43433d8477b24cd13acaac20a66deda49b7e2547" }
-tree-sitter-ocaml = { git = "https://github.com/tree-sitter/tree-sitter-ocaml.git", rev = "98c2130c59ca7553b47086f91c5d22180151ad55" }
```
mkatychev, line 69: removing the `tree-sitter-ocaml` git dependency has the benefit of avoiding a long sequence of submodule fetching, switching (rather than removing) the dep to a crates.io source should have the same effect.

mkatychev:
I'm going to add notes for potentially integrating `cargo nextest` and `mold` with crane/nix but I will defer it for now as follow up work for now. The mold linker seems to have sped [up the flow](https://github.com/tweag/topiary/pull/871#issuecomment-2651969001) and further tweaks to the workspace Cargo dependencies (removing `tree-sitter-ocal` git ref skipping fetching its many many submodules) and switching to JSON for benchmarking as further dropped the `ubuntu-latest` CI test time to [20m33s](https://github.com/tweag/topiary/actions/runs/13290653220/job/37110361596).

bin/default.nix
```diff
@@ -17,7 +17,7 @@
   playground = writeShellApplication {
     name = "playground";
 
-    runtimeInputs = with pkgs; [
+    runtimeInputs = with pkgs; pkgs.lib.optionals (!stdenv.isDarwin) [
```
Xophmeister, line 20: I don't know Nix well enough, but I wonder if there's a way of making the whole TTY playground Linux-only, rather than putting this safety check here for its dependencies.

topiary-core/benches/benchmark.rs
Xophmeister: Does that make it a good test, though? JSON formatting is quite trivial, whereas OCaml's is really hitting a large surface area of Topiary. My hunch would be therefore, that for benchmarking, pushing JSON through would result in a weaker signal.

(That said: We are not currently making any use of the benchmarking results, so the above point is kind-of-academic compared to the usefulness of reduced CI time.)

shell.nix
```diff
@@ -16,7 +16,12 @@ craneLib.devShell
       packages = with pkgs; with binPkgs; [
         cargo-dist
         cargo-flamegraph
+        # https://crane.dev/API.html?highlight=mold#cranelibcargonextest
+        # cargo-nextest
```
Xophmeister, line 20: I saw, either somewhere in this PR or in #872, that you're avoiding this for the time being because some parallelism doesn't play nice with `cargo-nextest`...

shell.nix
```diff
@@ -16,7 +16,12 @@ craneLib.devShell
...
         rust-analyzer
+        # # Look into mold-wrapped or `pkgs.stdenvAdapters.useMoldLinker`
+        # # https://crane.dev/API.html?highlight=mold#cranelibdevshell
+        # mold-wrapped
```
Xophmeister, line 24: ...however, is this bit also meant to be commented out? Is it intended as a "TODO" for a future PR?

Xophmeister (approved):
This looks great! Thank you :pray:

I've left a couple of questions, but otherwise this is good to merge.

shell.nix
mkatychev: Yes, I'm happy to remove it for now, the TODO would be to use the mold linker elsewhere but whether to use it just for derivations or devshell included 🤷.

bin/default.nix
mkatychev: we could, I've gotten the other parts to run on macOS and some of it helped me catch other issues in `verify-documented-usage`.

mkatychev:
@Xophmeister  agreed for https://github.com/tweag/topiary/pull/871#discussion_r1954140466
I think for now anything other than OCAML works as the issues with benchmarking it compounds upstream problems. However I'm in favour of doing the benchmarks at all (as they are done on push) in CI as it is taking up CPU cycles without catching any issues.

Xophmeister (approved):
This looks good :pray: I think using Nickel for the benchmarking is a good compromise :+1:

PR #877, opened 2025-02-19
Update flake lock, use buildWasmBindgenCli instead of override

yannham:
# Update flake.lock

## Description

This PR updates `flake.lock`. It also takes inspiration from a recent update on Nickel, where we could get rid of the more fragile `override` and use a Nixpkgs-provided helper instead to get a `wasm-bindgen-cli` with a fixed version.

By the way, should Topiary set up a `update-flake-lock` git-hub action, to make sure Nix dependencies get updated frequently and automatically?

## Checklist

This change doesn't need to be mentioned in the changelog.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "advisory-db": {
       "flake": false,
       "locked": {
-        "lastModified": 1738178975,
-        "narHash": "sha256-qrrRzUfIWEmr+XFS4yAS+WOYETqHpshYoXCfkpTX7kA=",
+        "lastModified": 1739520703,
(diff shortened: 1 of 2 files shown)


Xophmeister (approved):
I should start by saying I'm not really equipped to review this! However, I trust you @yannham and if it's the same approach you've used with Nickel, then I'm going to rubber stamp it through :slightly_smiling_face:

Thanks :pray:

Xophmeister:
> By the way, should Topiary set up a `update-flake-lock` git-hub action, to make sure Nix dependencies get updated frequently and automatically?

That's a good idea. I recently updated the release guide to include Nix+Cargo lock update steps, but having it done on a schedule in a GitHub action would be even better :+1:

PR #884, opened 2025-02-24
Topiary Book Super-PR

Xophmeister:
# Topiary Book Super-PR

Supersedes #750 
Resolves #874 
Resolves #875 
Resolves #876 

## Description

This is the base/super-PR for the Topiary Book. Sub-PRs will hang off this and the idea is that they'll all be merged into this, before everything is merged into `main`.

- [x] #885
- [x] #887
- [x] #890
- [x] #891
- [x] #892
- [x] #897
- [x] #898
- [x] Update `README` to point to new documentation
- [x] #906

## Checklist

Checklist before merging, wherever relevant:

- [x] `CHANGELOG.md` updated
- [x] Documentation (The Topiary Book, `README.md`, etc.) up-to-date

PR #885, opened 2025-02-25
Decant `README` into Topiary Book

Xophmeister:
Decant the original `README.md` into the Topiary Book, making adjustments as necessary.

> [!NOTE]
> This PR is only for decanting and **minor** adjustments, to speed the process along. More significant changes to the text can be done in downstream PRs. (Hence, expect a lot of "TODOs" and "FIXMEs".)
>
> It is meant to be rough :sweat_smile:


yannham (commented):
A couple points on the overall approach:

1. Even if we have a nice Mdbook, we should keep a skeletal README with basic information. Other projects (rustc, LALRPOP, etc.) do as well, since this is the landing page of the repo. At least a generic description should stay and a link to the mdbook.
2. I would keep the whole README during the transition, precisely because we don't have a nice book to link to yet, so the README is still the one and only source of technical documentation for users currently. We can always scrap it at the end.

Xophmeister:
> A couple points on the overall approach:
> 
> 1. Even if we have a nice Mdbook, we should keep a skeletal README with basic information. Other projects (rustc, LALRPOP, etc.) do as well, since this is the landing page of the repo. At least a generic description should stay and a link to the mdbook.
> 
> 2. I would keep the whole README during the transition, precisely because we don't have a nice book to link to yet, so the README is still the one and only source of technical documentation for users currently. We can always scrap it at the end.

Of yes, of course, the `README.md` will not disappear. I just removed content from it to go into the book so I could track progress. In the super-PR (#884), there's a note to reinstate some kind of `README`, and I had in mind exactly as you suggest.

That is, I don't plan on merging all this into `main`. I've just split the PR into sub-PRs, to maintain momentum. The sub-PRs will merge into the super-PR and, when that's all ready, then it'll get merged into `main`.

PR #893, opened 2025-03-04
Use `gix` lib instead of the `git` system install

### LINKED ISSUE #886: Investigate `gitoxide`
Topiary currently shells out to Git when fetching grammars dynamically. Originally, IIRC, we used a create that wraps `libgit2`; however we couldn’t get this to build in the macOS CI (hence the change).

[Gitoxide](https://github.com/GitoxideLabs/gitoxide) is a pure-Rust reimplementation of Git and looks mature enough to do everything we need.

Removing the need to shell out, without weird dependency problems, seems like a win.

nbacquey:
Closes #886 

## Description

This PR modifies the module that dynamically loads grammars from git repositories. We used to call the system's install of `git` (while never actually listing it as a dependency). Now we use `gix` instead, which removes that dependency while ensuring cross-platform compatibility with macOS.

## Checklist


Xophmeister (approved):
Nice one :bow: 

Does this still do the equivalent of a blobless clone?

PR #918, opened 2025-03-20
Add a language prefetcher for Nix

Niols:
# Add a language prefetcher for Nix

## Description

Since the introduction of dynamic loading in Topiary, the usability of Topiary in Nix environments has dropped a bit. In particular, it got difficult to run Topiary within a Nix sandbox, because Topiary wants internet access and cannot get it.

- This has blocked me from adding Topiary to [git-hooks.nix](https://github.com/cachix/git-hooks.nix).
- In nixpkgs, this forced the maintainers to [disable all the interesting tests](https://github.com/NixOS/nixpkgs/blob/f37ad1a90b3d3be8d40b386e0afad8a2bcb58ff7/pkgs/by-name/to/topiary/package.nix#L44-L57).
- In this very repository, it has been circumvented by [duplicating `topiary-config/languages.ncl`](https://github.com/tweag/topiary/blob/c62135d775155b53bb7c1a2a9080769d7c6e9918/topiary-config/languages_nix.ncl) and [injecting grammars from nixpkgs](https://github.com/tweag/topiary/blob/c62135d775155b53bb7c1a2a9080769d7c6e9918/default.nix#L123-L138). This solution is somewhat ugly, but, more importantly, it does not use the exact same versions of the grammars in the Nix and non-Nix version (in fact, some grammars are even missing from the Nix version).

This pull request introduces a piece of Nix code able to consume a Topiary configuration (a Nickel file) and produce another Topiary configuration that is the same, except sources that points to a Git repository have been replaced by sources that point to a pre-compiled binary. That binary is obtained by grabbing the Git repository using the regular Nix `fetchgit` and then passing it to nixpkgs's `tree-sitter.buildGrammar`. This allows almost transparent use of Topiary configurations in a Nix environment. The only thing is that one needs to inform Nix of the output hash; a classic in Nix, I'm afraid. I use this to replace the `languages_nix.ncl` hack in the package `topiary-cli-nix`. I am hoping to upstream it to nixpkgs as well, to recover tests there, and in turn use it in git-hooks.

For now, I consider this a proof-of-concept. It does seem to work for me, but I would love to see some other people play with it before. I would also enjoy reviews, in particular from Nix-savvy people, and also in particular from Nix+Nickel-savvy people for the Nix-Nickel interactions.

## Checklist

- [ ] `CHANGELOG.md` updated


Niols:
I should say that this also opens up the possibility to generate the configuration from Nix, which people from git-hooks.nix will enjoy. We could imagine providing something like `topiary-cli-nix.withConfig` that gets a function from config to config and lets the user do whatever they want with it, eg. bumping the version of one grammar, that sort of things.

Also, part of me wonders why we would want to keep `topiary-cli` and `topiary-cli-nix` separate, now that `topiary-cli-nix` uses the exact same grammars, just pre-fetched. Couldn't we just get rid of `topiary-cli` and use `topiary-cli-nix` in its place? (I'm only speaking of the Nix code - but probably if people are already using Nix, they prefer the `topiary-cli-nix` version anyway?)

Xophmeister (commented):
This sounds great; thanks, @Niols :pray:

I've had a look through this, but I'm no Nix person. It's perfectly coherent, as far as I can tell, but it should be reviewed by a Nix person :+1:

Nix usage is not heavily documented in the Topiary Book at the moment, but I think it would be worth mentioning how to generate the `nixHash` values in `docs/book/src/cli/configuration.md` (in the "Specifying the grammar" section).

Xophmeister:
> Also, part of me wonders why we would want to keep `topiary-cli` and `topiary-cli-nix` separate, now that `topiary-cli-nix` uses the exact same grammars, just pre-fetched. Couldn't we just get rid of `topiary-cli` and use `topiary-cli-nix` in its place? (I'm only speaking of the Nix code - but probably if people are already using Nix, they prefer the `topiary-cli-nix` version anyway?)

:100:

Is it idiomatic to have the `-nix` suffix? If not, I think I would choose to replace `topiary-cli` with `topiary-cli-nix`. The only issue I can think of is that it could be seen as a breaking change...but we're still in v0.x territory, so it's not so important.

yannham:
re `topiary-cli`/`topiary-cli-nix`: wouldn't that possible that you just want to build the Topiary CLI from Nix (say you want the last version in your profile for NixOS instead of the one from Nixpkgs), but you don't plan to _use it_ from Nix, so you're fine with grammar download?

prefetchLanguages.nix
```diff
@@ -0,0 +1,114 @@
...
+    readFile
+    toJSON
+    fromJSON
+    baseNameOf
+    ;
+
+  inherit (pkgs) fetchgit nickel;
+  inherit (pkgs.lib) warn;
+  inherit (pkgs.stdenvNoCC) mkDerivation;
+  inherit (pkgs.tree-sitter) buildGrammar;
```
yannham, line 18: Nitpick: it's quite a matter of personal taste, but it seems `buildGrammar` (and may other inherited symbols as well) is used only once in this whole file. In this case I'm a bit skeptical of the benefit of introducing an unqualified symbol (when I read `buildGrammar` for the first time below, it's a little bit less clear if it's coming from Nix or from this file, as opposed to a fully qualified `pkgs.tree-sitter.buildGrammar` - when you use it many times, there's of course a trade-off, but if it's only used once...). Anyway, also feel free to ignore :slightly_smiling_face:

prefetchLanguages.nix
```diff
@@ -0,0 +1,114 @@
...
+      '';
+    };
+
+  toNickelFile =
+    name: e:
+    mkDerivation {
+      name = "${name}.ncl";
+      dontUnpack = true;
+      buildPhase = ''
+        echo "import \"${toJSONFile name e}\"" > $out
```
yannham, line 83: We also sometimes use `nickel eval $file` to produce a Nickel translation of a JSON file (since `eval` prints the result as a Nickel value). But it's not really meant for this, so until we have a proper `nickel convert`, I guess the current approach is probably the safest, although it has the drawback of needing to keep the JSON file around as well. Another future possibility would be to accept `languages.json` file as well

yannham (approved):
I haven't tried it, but the approach and the implementation looks good to me :+1:

prefetchLanguages.nix
```diff
@@ -0,0 +1,114 @@
+{ pkgs, ... }:
```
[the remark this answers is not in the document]
Niols: Happy with 8429690dd3393c3ca5ef2909ee3a462fa3377b77?

prefetchLanguages.nix
Niols: Yeah, I wondered about that while writing it, actually. Fixed in 8429690dd3393c3ca5ef2909ee3a462fa3377b77.

Niols:
> Nix usage is not heavily documented in the Topiary Book at the moment, but I think it would be worth mentioning how to generate the `nixHash` values in `docs/book/src/cli/configuration.md` (in the "Specifying the grammar" section).

Happy with 8429690dd3393c3ca5ef2909ee3a462fa3377b77?

Niols:
> re `topiary-cli`/`topiary-cli-nix`: wouldn't that possible that you just want to build the Topiary CLI from Nix (say you want the last version in your profile for NixOS instead of the one from Nixpkgs), but you don't plan to _use it_ from Nix, so you're fine with grammar downloading?

> Is it idiomatic to have the `-nix` suffix? If not, I think I would choose to replace `topiary-cli` with `topiary-cli-nix`?

The nixpkgs way would be to have only `topiary-cli` with an boolean switch as argument to change its behaviour, and to then use `override` to change the argument if one wants it another way. We're actually almost there in Topiary: 

https://github.com/tweag/topiary/blob/c62135d775155b53bb7c1a2a9080769d7c6e9918/default.nix#L115

but later, in `flake.nix`:

https://github.com/tweag/topiary/blob/c62135d775155b53bb7c1a2a9080769d7c6e9918/flake.nix#L105-L106

I think the option that satisfies me the most at the moment would be to indeed do something like that. We would have only `topiary-cli` and by default that would not include the Nix pre-fetching. That way, it is predictible, Nix people get the same thing as everyone else.

Then, someone could also do `topiary-cli.withPrefetchedConfiguration (defaultConfig: ......)` which would allow them to choose to pre-fetch, and also potentially to modify the default configuration, for instance to filter out languages that they are not interested in.

The only thing that could then be needed would be for Topiary to give a hint on this famous `os error` (can't remember if it's 13, 30, or something else entirely), for people that would stumble upon it in their Nix sandbox.

prefetchLanguages.nix
Niols: I wanted to avoid calling Nickel because I was afraid of `import-from-derivation`, but in fact it was no problem in `fromNickelFile`. I can't really judge which solution is best, but I'll happily follow your decision on this!

prefetchLanguages.nix
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+  prefetchLanguageSource =
+    name: source:
+    if source ? "path" then
+      { inherit (source) path; }
+    else if source ? "git" then
+      { path = "${prefetchLanguageSourceGit name source.git}/parser"; }
+    else
+      throw ("Unsupported Topiary language sources: " ++ concatStringsSep ", " (attrNames source));
```
Niols, line 54: @Xophmeister > Btw, am I right to assume here that the only two types of sources at the moment are `path` and `git`?

prefetchLanguages.nix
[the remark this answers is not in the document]
yannham: Happy :slightly_smiling_face:

prefetchLanguages.nix
```diff
@@ -0,0 +1,105 @@
...
+      languages = mapAttrs (
+        name: languageConfig:
+        languageConfig
+        // {
+          grammar = languageConfig.grammar // {
+            source = prefetchLanguageSource name languageConfig.grammar.source;
+          };
+        }
+      ) topiaryConfig.languages;
+    };
```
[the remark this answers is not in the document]
yannham: I am not really a Nix Guru but I don't know of one, unfortunately. It's not excluded that there's a nice helper somewhere in Nixpkgs `lib`, though. Or maybe we should implement lenses :stuck_out_tongue:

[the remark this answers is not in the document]
Niols: Sad times :'(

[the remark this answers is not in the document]
Niols: Oh!

```
Type: updateManyAttrsByPath :: [{ path :: [String]; update :: (Any -> Any); }] -> AttrSet -> AttrSet
```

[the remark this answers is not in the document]
Niols: Is bfbae27e2b481b504fcb2396ab622d80f28bfd1d more readable?

Xophmeister:
> > Nix usage is not heavily documented in the Topiary Book at the moment, but I think it would be worth mentioning how to generate the `nixHash` values in `docs/book/src/cli/configuration.md` (in the "Specifying the grammar" section).
> 
> Happy with [8429690](https://github.com/tweag/topiary/commit/8429690dd3393c3ca5ef2909ee3a462fa3377b77)?

Very :heart: Thank you

prefetchLanguages.nix
Xophmeister: Yep, that's right :+1:

docs/book/src/cli/configuration.md
```diff
@@ -158,6 +158,33 @@ nickel = {
...
+
+```
+evaluation warning: Language `nickel`: no nixHash provided - using dummy value
+error: hash mismatch in fixed-output derivation '/nix/store/jgny7ll7plh7rfdnvdpgcb82kd51aiyx-tree-sitter-nickel-43433d8.drv':
+         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+            got:    sha256-9Ei0uy+eGK9oiH7y2KIhB1E88SRzGnZinqECT3kYTVE=
+error: 1 dependencies of derivation '/nix/store/0q20rk8l4g0n5fzr0w45agxx0j9qy65v-nickel-grammar-43433d8477b24cd13acaac20a66deda49b7e2547.drv' failed to build
+error: 1 dependencies of derivation '/nix/store/s5phxykjyzqay7gc33hc6f8kw4ndba25-languages-prefetched.json.drv' failed to build
+error: 1 dependencies of derivation '/nix/store/5w15p3b3xfw5nd6mxz58ln09v10kvf8v-languages-prefetched.ncl.drv' failed to build
+error: 1 dependencies of derivation '/nix/store/7zzyha67jw09kc37valp28bp5h6i7dka-topiary-0.6.0.drv' failed to build
```
mkatychev, line 189: I think this is a great addition and is definitely in the right direction, I wonder if there is a task-runner enabled way to expedite the hashing. Intentionally generating an error is one [common way of hashing](https://sekun.net/blog/packaging-prebuilt-binaries-with-nix/#4), however I think this seems a bit tedious and unintuitive unless one is already familiar with how NAR works.

Niols: I suppose it wouldn't be hard to provide a tiny script that would add the `nixHash` for all the sources. Basically a tiny wrapper around the existing sources crawler and `nix-prefetch-git`. This couldn't be integrated with this function, because we want the fixed output, but this could be a useful helper, maybe something nix-runnable?

Niols:
> The nixpkgs way would be to have only `topiary-cli` with an boolean switch as argument to change its behaviour, and to then use `override` to change the argument if one wants it another way. [...]

I believe 6101043c0e22a17872364c0ca18bd2854debdc1a implements the more nixpkgs-y way of doing things. In `default.nix`, `topiary-cli` is already a package, but it is overridable and one can do `topiary-cli.override { prefetchGramars = true }` if they want that. I have kept `topiary-cli-nix` in the flake for compatibility, but it could go away, I suppose.

While I was at it, I considered providing a mechanism to modify the default configuration, eg. with `topiary-cli.withConfiguration (defaultConfig: ...)`. I think it would make plenty sense to have that and it would prove useful if one wants to only prefetch a few grammars, eg. in `git-hooks.nix`. However, I see currently two ways and I don't like either:

1. Go the same route as we are going now, changing the `languages.ncl` file that gets embedded in Topiary at compile time. This means that each default configuration will lead to a different package of Topiary. Changing configuration requires re-compiling everything (which takes foooooorever) and the Nix store will get poluted with as many versions of Topiary as there are default configurations, even though that's the only thing that differs.

2. Keep the default `languages.ncl` as it is, but provide a small wrapper around Topiary that injects the wanted configuration. This allows sharing the `topiary-cli` package, saving compilation time and space on the machine. The “problem” here is that there is currently no way to point Topiary at a different _default_ configuration. I can write a wrapper that points Topiary at a certain configuration, but the result will not behave the same way as Topiary, in particular it will refuse the `-C` argument and (I think) it will ignore repository-specific configuration. This might make sense in some contexts (to enforce reproducibility and avoid a Nix-provided Topiary to pick up one developer's local machine's configuration), but it is an odd caveat.

I need to think some more about this.

Niols:
Having thought about it some more, I think we should just leave the `withConfig` question out of scope. For now, this PR brings an improvement over the current situation. We can iterate later on to bring new features or clean things up, while keeping PRs well defined and nicely contained.

Niols:
So if @Xophmeister and @yannham give the green light, I think we should just merge this.

Xophmeister (approved):
Sounds like a good plan to merge now, in its current state, and iterate on features later :+1:

prefetchLanguages.nix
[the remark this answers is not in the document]
yannham: It is!

yannham (approved):
Sounds good to me as well.

docs/book/src/cli/configuration.md
yannham: It surely could help! I still think the current solution is kinda ok in the meantime, because if someone needs to use the Nix-compatible version of Topiary, they most likely know at least a tiny bit of Nix/are a Nix user?

Niols: I'd also want to say that, at least for this PR, we can keep the current solution. I wrote a tiny experimental Shell script that does this for you, but I'm not sure I want to add it to this PR.

PR #939, opened 2025-03-26
Fix edge cases in dangling behaviours

### LINKED ISSUE #914: Idempotency error on
As of c62135d775155b53bb7c1a2a9080769d7c6e9918, Topiary runs into an idempotency error on:

```ocaml
match foo with
| _ ->
  (
    match bar with
    | [] -> None
    | _ -> foo
  )
```

First time round:
```ocaml
match foo with
| _ ->
  (
    match bar with
    | [
    ] -> None
    | _ -> foo
  )
```

Second time round:
```ocaml
match foo with
| _ ->
  (
    match bar with
    | [
    ] ->
      None
    | _ -> foo
  )
```

Somehow, this is my minimal reproducer: removing the outside `match` or the parentheses gets rid of the error.

### LINKED ISSUE #915: Idempotency error on list singleton in nested match
As of https://github.com/tweag/topiary/commit/c62135d775155b53bb7c1a2a9080769d7c6e9918, Topiary runs into an idempotency error on:
```ocaml
match foo with
| _ ->
  (
    if path = [] then
      ()
    else
      ()
  )
```

First time round:
```ocaml
match foo with
| _ ->
  (
    if path = [
        ] then
      ()
    else
      ()
  )
```

Second time round:
```ocaml
match foo with
| _ ->
  (
    if path
      = [
        ] then
      ()
    else
      ()
  )
```

Somehow, this is my minimal reproducer: removing the outside `match` or the parentheses gets rid of the error. This is probably the same as https://github.com/tweag/topiary/issues/914

### LINKED ISSUE #916: Idempotency error on let-binding with type with parameters
As of https://github.com/tweag/topiary/commit/c62135d775155b53bb7c1a2a9080769d7c6e9918, Topiary runs into an idempotency error on:

```ocaml
let foo : (bar, baz) qux =
  ()
```

First time round:
```ocaml
let foo :
    (bar, baz) qux =
  ()
```

Second time round:
```ocaml
let foo
    :
    (bar, baz) qux
  =
  ()
```

I'm not sure why Topiary tries to multi-line this in the first place; I kind of like it as it is; I'd expect Topiary to multi-line my type only if I already make it multi-line.

It has to do with multiple type arguments. The following piece of code has no issue:
```ocaml
let foo : bar qux =
  ()

let foo : qux =
  ()
```
and is formatted as-is.

### LINKED ISSUE #917: “Trying to close an unopened indentation block” on some multi-line list
Sorry for the weird title, I cannot make sense of this one.

As of c62135d775155b53bb7c1a2a9080769d7c6e9918, Topiary fails on:

```ocaml
let _ =
  div
    ~a: [a_id "page_nav"]
    [
      some stuff
    ]
```

with

```
[2025-03-20T09:57:23Z ERROR topiary] Trying to close an unopened indentation block
```

I was initially tracking an idempotency error where Topiary changed identation of the multi-line list at each run. I will share a reproducer for that too later.

This happens with or without `some stuff` in the list, but it does not happen if the list is single-line.

This only happens with that named first argument. It doesn't happen:
- with just a named argument, eg. `div ~a [some stuff]` (but multiline)
- with a named argument pointing to something that is not a list, eg. `div ~a: b [some stuff]`
It does happen no matter what the list give to `~a` is.

nbacquey:
## Description
Closes #914
Closes #915
Closes #916
Closes #917

Also fixes some minor issues with antispace, that did not appear before.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ This name should be decided amongst the team before the release.
 <!-- - <Bug fixes> -->
 - [#867](https://github.com/tweag/topiary/pull/867) Enable coverage check and add code samples for OpenSCAD
 - [#869](https://github.com/tweag/topiary/pull/869) Disable parallel grammar building on Windows
-- [#908](https://github.com/tweag/topiary/pull/908) [#907](https://github.com/tweag/topiary/pull/907) Various OCaml issues and improvements
+- [#908](https://github.com/tweag/topiary/pull/908) [#907](https://github.com/tweag/topiary/pull/907) [#939](https://github.com/tweag/topiary/pull/939) Various OCaml issues and improvements
 
 <!-- ### Security -->
 <!-- - <Vulnerabilities> -->


topiary-cli/tests/samples/expected/ocaml.ml
```diff
@@ -73,7 +73,7 @@ let create n =
   let n = if n < 1 then 1 else n in
   let n = if n > Sys.max_string_length then Sys.max_string_length else n in
   let s = Bytes.create n in
-  {buffer = s; position = 0; length = n; initial_buffer = s}
+    {buffer = s; position = 0; length = n; initial_buffer = s}
```
Niols, line 76: A bit puzzled by this indentation after a chain of `let ... in`. Is that a side effect or a choice?

topiary-cli/tests/samples/expected/ocaml.ml
```diff
@@ -953,7 +953,7 @@ open Bar
 
 let _ =
   let open Baz in
-  ()
+    ()
```
Niols, line 956: Ditto.

Niols (approved):
I cannot judge on the state of the query, as always. The OCaml example look good to me. It's very practical that we can just include my snippets in the tests every time!

I am just a bit puzzled by the indentation of objects after a `let .. in`. cf comments.

topiary-cli/tests/samples/expected/ocaml.ml
```diff
@@ -1415,3 +1415,57 @@ type foo = {
...
+  | _ ->
+    (
+      if path = [] then
+          ()
+      else
+          ()
+    )
+
+let foo : (bar, baz) qux =
+    ()
```
Xophmeister, line 1440: Are there meant to be double indents before `()` on lines 1434, 1436 and 1440?

Xophmeister (approved):
I noticed a few double-indents, but they may be intentional. Otherwise, LGTM :+1:

topiary-cli/tests/samples/expected/ocaml.ml
Niols: Oh, didn't even catch that one. I feel it might be related to my own comments: a bunch of places where there is one too many indent levels, either because we're expecting only one level (here) or zero (my comments above).

topiary-cli/tests/samples/expected/ocaml.ml
Niols: I don't understand why we want to format # 2 as it is?

Niols: I mean, those are two very different examples. In one case, we're after the `=` and in one case after the `in`. Independently from what's inside (so it isn't really about records and whatnot), we probably want:

```ocaml
let foo = [something] in
[something not indented]
```

and

```ocaml
let foo =
  [something indented by one level]
in
[something not indented]
```

For records, I guess that's:

```ocaml
let foo = { bar = 1; baz = 2 } in
{ qux = 1; quux = 2 }
```

and

```ocaml
let foo = {
  bar = 1;
  baz = 2;
}
in
{
  qux = 1;
  quux = 2;
}
```

and I think that's rather common OCaml style. So I still don't see why we _want_ something different. (But I'm just missing something I think?)

Maybe your answer lies in your answer to @Xophmeister, namely that it is just too hard to get, so although we want that we will still go for the over-indented version? That is, at this point at least, perfectly fine, imo!

Niols: Absolutely, the PR definitely improves the situation!

Niols: In particular, it fixes all the idempotency problem, so we have a working formatter. It's quite a luxury to get to discuss about style!

PR #964, opened 2025-04-07
Generalize scope opens and closes for `dangling_list_like` scope

### LINKED ISSUE #956: Idempotency error on record as argument of a function as argument of another one
These titles make less and less sense, I'm sorry --

As of 50023c6a756ba40df9dc072b3a0a475aba53e551, Topiary hits an idempotency error on:

```ocaml
foo
  (
    bar {baz} qux
  )
```

First round:

```ocaml
foo
  (
    bar {
        baz
      } qux
  )
```

Second round:

```ocaml
foo
  (
    bar
      {
        baz
      }
      qux
  )
```

Looks like the dangling operators are a PITA...

### LINKED ISSUE #957: Idempotency error on record pattern in a `let%`
As of 50023c6a756ba40df9dc072b3a0a475aba53e551, Topiary hits an idempotency error on:

```ocaml
foo
  (
    let%tag {bar} = baz in
    ()
  )
```

First round:

```ocaml
foo
  (
    let%tag {
      bar
    } = baz in
      ()
  )
```

(I don't know why it's trying to make it multi-line. It probably shouldn't. It doesn't without the `%tag`.)

Second round:

```ocaml
foo
  (
    let%tag {
      bar
    }
      =
      baz
    in
      ()
  )
```

nbacquey:
This PR tries to generalize cases where the `dangling_list_like` scope opens. Basically, I tried to cover every instance that looks like `some_syntactic_element { ... }`. I probably missed some, but hopefully this small refactoring will make later cases easier to fix.

Some edge cases were broken by the anchor `.` skipping non-named nodes, so I had to hack around a bit with
`_? @do_nothing` to force the queries not to skip those nodes.

For reviewers: diffs to the query file will probably be hard to read. it may be easier to directly have a look at the query file [here](https://github.com/tweag/topiary/blob/2eb4b82b7b44398ef30475ae9c214c9027530d5b/topiary-queries/queries/ocaml.scm#L753C1-L908C2).

Closes #956
Closes #957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@ This name should be decided amongst the team before the release.
 <!-- - <Bug fixes> -->
 - [#867](https://github.com/tweag/topiary/pull/867) Enable coverage check and add code samples for OpenSCAD
 - [#869](https://github.com/tweag/topiary/pull/869) Disable parallel grammar building on Windows
-- [#908](https://github.com/tweag/topiary/pull/908) [#907](https://github.com/tweag/topiary/pull/907) [#939](https://github.com/tweag/topiary/pull/939) [#955](https://github.com/tweag/topiary/pull/955) Various OCaml issues and improvements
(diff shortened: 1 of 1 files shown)


Niols (approved):
Test samples look good. I can confirm that this seems to be working fine on my codebase. (Only one more source of idempotency to go, I think!)

Niols:
In fact, the idempotency error I found while trying this out might be very much related, and is on `main` as well. See https://github.com/tweag/topiary/issues/965.

Xophmeister (approved):

PR #967, opened 2025-04-08
Open `dangling_list_like` scope after `fun` keyword

### LINKED ISSUE #965: Idempotency error on record pattern in a function argument
As of a82656c26edf588673e2ba622228f182895f9b98 (latest `main`) and 2eb4b82b7b44398ef30475ae9c214c9027530d5b (head of #964), Topiary hits an idempotency error on:

```ocaml
foo
  (fun {bar} ->
    bar
  )
```

First round:

```ocaml
foo
  (fun {
        bar
      } ->
    bar
  )
```

Second round:

```ocaml
foo
  (fun
      {
        bar
      }
    ->
    bar
  )
```

I would expect things to remain similar to the first snippet, that is with `fun {bar} ->` on the same line.

Maybe related to #956 and #957 that both have to do with records handling.

nbacquey:
Closes #965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@ This name should be decided amongst the team before the release.
 <!-- - <Bug fixes> -->
 - [#867](https://github.com/tweag/topiary/pull/867) Enable coverage check and add code samples for OpenSCAD
 - [#869](https://github.com/tweag/topiary/pull/869) Disable parallel grammar building on Windows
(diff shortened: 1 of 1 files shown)


Niols (approved):
Test samples look good. Also, excellent news, as of 73dc67ace24f9dbf6210b8e462a679c11bec22b9 I can format my whole codebase without any idempotency issues. (I think there are some weird behaviours, but this is only style and we can discuss it in subsequent issues.) Nice :tada: 

I should try on more code, though, but this looks excellent :-)

Issue #1003, opened 2025-05-20
Production search paths for queries


Xophmeister:
The search path for Topiary's formatting queries is currently:

1. The `TOPIARY_LANGUAGE_DIR` environment variable as set during runtime.
2. A baked in path, determined by the `TOPIARY_LANGUAGE_DIR` environment variable that was set at build time.
3. `./topiary-queries/queries`
4. `../topiary-queries/queries`

This makes sense for development purposes, but it doesn't feel correct for production use. I would suggest something similar to the configuration search paths:

1. The `TOPIARY_LANGUAGE_DIR` environment variable as set during runtime.
2. A `queries` directory that is a sibling to the configuration file, however the configuration is specified (except for the case when it defaults to the built-in configuration).
3. Ascending from the working dir until a `.topiary/queries` is found
4. `~/.config/topiary/queries`

I'm not convinced about (2); maybe exposing this as a command line option, as with the configuration, would be a clearer solution.

CC @tweag/topiary-core-team

PR #1005, opened 2025-05-20
Correct documentation of formatting query search paths

### LINKED ISSUE #1002: Where does Topiary look for queries?
[this comment](https://github.com/tweag/topiary/blob/a8c8f1bf0ca30820fbb84283e45320cb988f21f0/docs/book/src/cli/dialogue.md?plain=1#L7) says it looks for a folder called "languages" but I'm pretty sure [it doesn't](https://github.com/tweag/topiary/blob/a8c8f1bf0ca30820fbb84283e45320cb988f21f0/topiary-config/src/language.rs#L93).

I'm also kind of confused why it doesn't look in `.topiary`? What's the point of being able to have a project-specific config if you can't also put the query file in there?

Xophmeister:
# Correct documentation of formatting query search paths

Resolves #1002
Also see #1003

## Description

Corrects the Topiary CLI documentation, in the Topiary Book, with the correct search path priority for formatting query files.

## Checklist

- [ ] ~`CHANGELOG.md` updated~

### DIFF
--- a/docs/book/src/reference/capture-names/general.md
+++ b/docs/book/src/reference/capture-names/general.md
@@ -230,7 +230,7 @@ counter-intuitive. Consider, for instance, the following query:
 ```
 
 One might assume that this query only matches the final element in the
-list but this is not true. Since we did not explicitly march a parent
+list but this is not true. Since we did not explicitly match a parent
 node, the engine will match on every `list_entry`. After all, when
 looking only at the nodes in the query, the `list_entry` is indeed the
 last node.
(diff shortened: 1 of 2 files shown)


docs/book/src/reference/capture-names/general.md
```diff
@@ -230,7 +230,7 @@ counter-intuitive. Consider, for instance, the following query:
 ```
 
 One might assume that this query only matches the final element in the
-list but this is not true. Since we did not explicitly march a parent
+list but this is not true. Since we did not explicitly match a parent
```
yannham, line 233: Now I imagine someone marching on a poor innocent parent.

yannham (approved):

PR #1032, opened 2025-06-16
Remove dependencies on Tree-sitter grammars

### LINKED ISSUE #1022: Remove dependency on `tree-sitter-nickel`
`tree-sitter-nickel` is a declared dependency in `Cargo.toml`, but it is only used in the [little-used] benchmark testing (`topiary-core/benches/benchmark.rs`). This benchmark was written before dynamic loading was implemented and, while we don't want to benchmark the dynamic loading -- as it can involve a download-and-build step -- we could still make use of this mechanism.

This would remove the dependency on `tree-sitter-nickel` and make `topiary-config/languages.ncl` the authoritative source.

(For a similar discussion for `tree-sitter-json`, see #1023.)

Xophmeister:
# Remove dependencies on Tree-sitter grammars

Resolves #1022
Resolves #1023

## Description

Remove the development dependencies on `tree-sitter-json` and `tree-sitter-nickel`. Instead, use the dynamic loading from `topiary-config` in the relevant tests.

## Checklist

- [x] `CHANGELOG.md` updated


Xophmeister:
These changes work outside of Nix, but the CI complains inside (e.g., see [here](https://github.com/tweag/topiary/actions/runs/15687381415/job/44193857267?pr=1032)). My best guess is: Now that `topiary-core` depends on `topiary-config` and `topiary-queries`, for testing, I _think_ that the derivation under test is simply missing the query files. That is, the "permission denied" error is misleading and should be, rather, "file not found".

Does that sound plausible, @Niols?

Adding the JSON and Nickel query files to the `topiary-core` package doesn't sound like the correct solution, as they're only needed for testing. What would be the right approach?

Xophmeister:
> > Adding the JSON and Nickel query files to the `topiary-core` package doesn't sound like the correct solution, as they're only needed for testing. What would be the right approach?
> 
> We could disable tests in the Nix package, but that's a bit sad. We could also make `topiary-core` 's Nix package depend on `topiary-queries` (doesn't it sort of already, if there is such a dependency in Rust as well?) and point (don't know how yet) the tests at the right path in the store?

We currently rely on the tests to run in the CI, so disabling them is a bit of a non-starter. `topiary-queries` is a dev-dependency of `topiary-core`, so perhaps there's scope for two Nix packages: one with tests -- which we invoke in the CI -- and one without, which downstream users can benefit from... I don't know how "idiomatic" that would be :shrug: 

@yannham and @jneem: Have you any experience in this, from Nickel or elsewhere?

Niols:
Nix already has this sort of distinction between build dependencies and runtime dependencies. Downstream users would presumably just download the pre-built binaries from a Nix cache, which will also require runtime dependencies, but not build dependencies.

mkatychev:
Hi @Xophmeister, is this PR still needed to be reviewed and relevant?

Xophmeister:
Thanks for looking through these old PRs, @mkatychev :pray:

It's definitely still relevant, but it's so old now that there's some serious divergence and, IIRC, I was never able to address the Nix code. I'll need to look into it a bit more closely, but I suspect I might be able to get this over the line, if time permits and the rebase doesn't drive me crazy :sweat_smile:

I'll add it to my todo list :+1:

Xophmeister:
I've rebased this on current `main` and it looks like the Nix refactoring done in #1026 has solved the problem that was blocking this PR for free! Thanks, @Niols :pray:

This is now ready for review :tada:

Xophmeister:
GitHub won't let me change reviewers, probably because the `topiary-core-team` was a Tweag team and the org migration doesn't account for that... So pinging @topiary/core @topiary/honorary

mkatychev (approved):
What a fantastic change, makes everything more consistent! Thank you for your effort!

Cargo.toml
```diff
@@ -62,9 +62,7 @@ thiserror = "2.0.16"
 tokio = "1.32"
 tokio-test = "0.4"
 toml = "1.0"
-tree-sitter-json = "0.24"
 tree-sitter-language = "0.1"
-tree-sitter-nickel = "0.5"
```
mkatychev, line 67: 💪

PR #1046, opened 2025-07-24
Revert CI back to `macos-latest`

### LINKED ISSUE #1043: Revert to `macos-latest`
## Reminder

Revert to `macos-latest` in the CI when its problems have been fixed upstream.

Depends on:
- [x] https://github.com/actions/runner-images/issues/12588

Xophmeister:
# Revert CI back to `macos-latest`

Resolves #1043
References https://github.com/actions/runner-images/issues/12588

## Description

Reverts the main CI's macOS run to use GitHub's `macos-latest` runner, rather than `macos-14`, as this has been fixed upstream.

## Checklist

- [ ] ~`CHANGELOG.md` updated~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macos-14  # TODO Revert to macos-latest; see #1043
+          - macos-latest
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code


avdv:
Hi, I am wondering... macos-14 and macos-latest are the same thing currently. (migration to macos-15 will start in August)

Xophmeister:
> Hi, I am wondering... macos-14 and macos-latest are the same thing currently. (migration to macos-15 will start in August)

I didn't actually check, but that's the same thing GitHub said. However, the build was failing and there was an error about the new macOS runners, so I changed it to `macos-14` and that fixed the problem. Then GitHub told me the same thing as you, so I reverted it and the build still works... So I'm super-confused :shrug:

Maybe the build broke for a completely unrelated reason, due to some other action, which was fixed very quickly in between the time of me noticing it and making a fix (less than a day). Or perhaps Topiary was subjected to random A/B testing and we got unlucky. I don't know what's the simplest explanation is here :sweat_smile:

Either way, it's working now, so maybe my curiosity doesn't need to be satisfied!

PR #1070, opened 2025-08-27
perf(topiary-core): Implemented parallelism for query coverage

mkatychev:
# implemented parallel iteration for query coverage

References https://github.com/tweag/topiary/issues/386 https://github.com/tweag/topiary/issues/539

## Description

This PR adds rayon to the coverage functionality of topiary, locally this reduced OCaml coverage time from 5 minutes to 30 seconds.

## Checklist

- [x] `CHANGELOG.md` updated


mkatychev:
@Xophmeister this PR as well as the GHA work in https://github.com/tweag/topiary/pull/1069 should reduce CI issues quite a bit, as mentioned in https://github.com/tweag/topiary/issues/1068#issue-3357493418, the OCaml grammar itself is very underperformant so updating that to latest should fix the last of the imminent issues without worrying about formatting parallelism.

Xophmeister (approved):
As always, this looks great @mkatychev. Thank you :pray:

The only comment I would add, which is entirely optional, is that `topiary-config` exposes the `parallel` (default) feature which adds Rayon as a dependency and uses it for grammar prefetching. I'm not sure why anyone would want to disable parallelisation -- although in the `topiary-config` case, we have to switch it off for Windows builds -- but it might be nice to reflect this feature in `topiary-core`. What do you think?

Xophmeister:
> @Xophmeister are you asking for an implementation of rayon in topiary-config for prefetching as part of this PR?

Oh no, not at all; that would be very pushy of me :sweat_smile:

I'm saying Rayon is _already_ used in `topiary-config` which is behind a feature gate (`parallel`, which is a default feature). I'm just thinking it might make sense -- for symmetry's sake -- to also feature gate the work you've done in this PR. However, I'm completely open to it being merged as it currently stands :+1:

mkatychev:
Looking at this comment: https://github.com/tweag/topiary/blob/070eb656e4bce4577e039fd0f06df9d725e4da91/topiary-config/src/lib.rs#L177-L188

The problem isn't necessarily parallelism/rayon but rather MSVC not linking concurrent linker use here:
https://github.com/tweag/topiary/blob/070eb656e4bce4577e039fd0f06df9d725e4da91/topiary-config/src/language.rs#L306-L312

As long as we're not calling `Config::fetch_language`, we should be totally safe on windows targets as long as we don't cross any system boundaries (or try to call the linker a bunch), I don't mind adding it in but the CI windows test isn't encountering any regressions.

Xophmeister:
Sorry, @mkatychev: I think I'm still being unclear. This isn't about linking or `Config::fetch_language`; those were just to try to illustrate my point... It's been a long day! Apologies!!

All I'm asking is: Do you think it's worthwhile putting your parallel coverage implementation behind the `parallel` feature flag?

PR #1071, opened 2025-08-28
perf(topiary-core): Query coverage performance v2

mkatychev:
# Query coverage performance v2

References https://github.com/tweag/topiary/issues/1068

## Description

After looking at the code some more, turns out we were doing a lot of excessive work in the query section here:
https://github.com/tweag/topiary/blob/f4b67a025a32dc1b6265de6b2d19b6a518124fdd/topiary-core/src/tree_sitter.rs#L572-L574

By removing only one query for _n_ patterns and then running a query with _n - 1_ patterns _n_ times we were handling a complexity of `O(n^2-n)`, this implementation does coverage by splitting each pattern into their separate queries.

## Checklist

- [x] `CHANGELOG.md` updated


mkatychev:
Windows test time seems to have the biggest gain with all these changes (complexity, parallelism & CI): 14 -> 3

Xophmeister (approved):

PR #1080, opened 2025-09-15
feat(core): Allow formatting existing tree-sitter trees

### LINKED ISSUE #1079: Allow formatting already parsed tree
I'm contributing to [gdscript-formatter](https://github.com/GDQuest/godot-gdscript-formatter-tree-sitter), we are using topiary as a library and we a thinking about adding a preprocess step to handle long lines (and possible reorder some top-level statements). Currently it would require to parse source code one more time, to avoid that it would be nice to have an ability to pass already parsed tree-sitter tree into a `formatter` function

Create two functions `formatter_tree` and `apply_query_tree` that accepts additional `tree: Tree` argument

1) parse source code one more time and take a performance hit
2) manually walk over input source code without a tree, creates more code than first option and requires more manual character tracking than tree-sitter queries

**Additional context**
I looked through code a little bit, `formatter` has two operations in one it calls `apply_query` and in another first thing it does it parses a tree, in `apply_query` on the first line tree is parsed. I think it should be an easy change, hardest thing is to name and document new functions. If maintainers are fine with that change I can make a PR myself.

shadr:
# Allow formatting existing tree-sitter trees

Resolves #1079

## Description

This PR adds two functions `formatter_tree` and `apply_query_tree` to allow users of `topiary-core` to format existing trees and avoid cost of parsing input twice

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/topiary-core/src/lib.rs
+++ b/topiary-core/src/lib.rs
@@ -239,6 +239,31 @@ pub fn formatter(
         ))
     })?;
 
+    let tree = match operation {
+        Operation::Format {
+            tolerate_parsing_errors,
+            ..
+        } => tree_sitter::parse(&content, &language.grammar, tolerate_parsing_errors)?,
+        Operation::Visualise { .. } => tree_sitter::parse(&content, &language.grammar, false)?,
+    };
+
+    formatter_tree(tree, &content, output, language, operation)?;
+
+    Ok(())
+}
+
+/// The function that takes a tree and formats, or visualises an output.
+///
+/// # Errors
+///
+/// If formatting fails for any reason, a `FormatterError` will be returned.
+pub fn formatter_tree(
+    tree: topiary_tree_sitter_facade::Tree,
+    input_content: &str,
+    output: &mut impl io::Write,
+    language: &Language,
+    operation: Operation,
+) -> FormatterResult<()> {
     match operation {
         Operation::Format {
             skip_idempotence,
@@ -247,12 +272,7 @@ pub fn formatter(
             // All the work related to tree-sitter and the query is done here
             log::info!("Apply Tree-sitter query");
 
-            let mut atoms = tree_sitter::apply_query(
-                &content,
-                &language.query,
-                &language.grammar,
-                tolerate_parsing_errors,
-            )?;
+            let mut atoms = tree_sitter::apply_query_tree(tree, input_content, &language.query)?;
 
             // Various post-processing of whitespace
             atoms.post_process();
@@ -276,7 +296,6 @@ pub fn formatter(
         }
 
         Operation::Visualise { output_format } => {
-            let tree = tree_sitter::parse(&content, &language.grammar, false)?;
             let root: SyntaxNode = tree.root_node().into();
 
             match output_format {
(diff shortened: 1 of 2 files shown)


topiary-core/src/lib.rs
```diff
@@ -239,6 +239,31 @@ pub fn formatter(
         ))
     })?;
```
shadr, line 240: Should we create a `formatter_str` to avoid reading input from existing String ?

Xophmeister:
> Here we read from an io::Read into a String, should we add a `formatter_str` funtion for the cases when user already has String ?
> 
> https://github.com/tweag/topiary/blob/82c179d26b1348f1e7779d352b073b4dc55912e6/topiary-core/src/lib.rs#L235-L240

@yannham Is this related what you were talking about regarding the Nickel `--check` option, where it would be better to provide a `&str` interface, rather than `impl io::Read`, for memory efficiency?

`&str` implements the `io::Read` trait, so we never thought this was necessary beforehand, but I don't see any harm in adding this; especially as it's come up twice, now.

yannham:
> @yannham Is this related what you were talking about regarding the Nickel fmt --check option, where it would be better to provide a &str interface, rather than impl io::Read, for memory efficiency?

Yes, I think this is the same demand. Currently, if you already have a big `String` allocated somewhere, or a `&str`, going through the `Read` interface will basically duplicate it entirely in memory, just to take a reference out of it in the end. So I suppose the "low-level" function should take a `&str`, and we can still keep a `impl Read` wrapper that reads some content and call to the low-level version.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ This name should be decided amongst the team before the release.
 - [#1069](https://github.com/tweag/topiary/pull/1069) split CI nix dependent tests into separate jobs, thanks to @mkatychev
 - [#1070](https://github.com/tweag/topiary/pull/1070) Implemented parallelism for query coverage, thanks to @mkatychev
 - [#1070](https://github.com/tweag/topiary/pull/1072) Add source spans to query coverage results, thanks to @mkatychev
+- [#1080](https://github.com/tweag/topiary/pull/1080) Allow formatting existing tree-sitter trees
```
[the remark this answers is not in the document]
shadr: Didn't saw that. I guess my brain parsed these lines as "just PR titles" and didn't see beyond comma :)

topiary-core/src/lib.rs
```diff
@@ -239,6 +239,47 @@ pub fn formatter(
...
+        Operation::Format {
+            tolerate_parsing_errors,
+            ..
+        } => tolerate_parsing_errors,
+        _ => false,
+    };
+
+    let tree = tree_sitter::parse(&input, &language.grammar, tolerate_parsing_errors)?;
+
+    formatter_tree(tree, &input, output, language, operation)?;
```
Xophmeister, line 266: Clippy is saying that both of these `&input`s are needless borrows, so the CI is failing:

```suggestion
    let tree = tree_sitter::parse(input, &language.grammar, tolerate_parsing_errors)?;

    formatter_tree(tree, input, output, language, operation)?;
```

Xophmeister:
@shadr This looks good :+1: Let me know if it's ready to merge.

PR #1092, opened 2025-09-24
fix(cli): log early returns during formatting

mkatychev:
# log early returns during formatting

## Description
Certain errors would not get logged during an early return because they exited the async block rather than the let scope:
https://github.com/tweag/topiary/blob/5f336358fe2ccc9a0a9c2b52b38d60df67f41cf8/topiary-cli/src/main.rs#L62-L63

This resulted in only the match arm below being logged:
https://github.com/tweag/topiary/blob/5f336358fe2ccc9a0a9c2b52b38d60df67f41cf8/topiary-cli/src/main.rs#L102-L104

This PR changes the scoped future to log all errors emitted.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -12,14 +12,14 @@ use std::{
 };
 
 use error::Benign;
-use tabled::{Table, settings::Style};
+use tabled::{settings::Style, Table};
 use topiary_config::source::Source;
-use topiary_core::{Operation, check_query_coverage, formatter};
+use topiary_core::{check_query_coverage, formatter, Operation};
 
 use crate::{
     cli::Commands,
     error::{CLIError, CLIResult, TopiaryError},
-    io::{Inputs, OutputFile, read_input},
+    io::{read_input, Inputs, OutputFile},
     language::LanguageDefinitionCache,
 };
 
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -108,14 +108,12 @@ async fn run() -> CLIResult<()> {
             }
 
             let errs = results
-                .iter()
-                .inspect(|r| match r {
-                    Err(e) => print_error(&e),
-                    Ok(Err(e)) if !e.benign() => print_error(&e),
-                    _ => {}
-                })
-                .filter(|result| matches!(result, Err(_) | Ok(Err(_))))
+                .into_iter()
+                .map(|r| r.map_err(TopiaryError::from).and_then(|inner| inner))
+                .filter_map(|r| r.err())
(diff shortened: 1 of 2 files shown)


topiary-cli/src/main.rs
```diff
@@ -57,60 +57,47 @@ async fn run() -> CLIResult<()> {
             let (_, mut results) = async_scoped::TokioScope::scope_and_block(|scope| {
                 for input in inputs {
                     scope.spawn(async {
-                        let result: CLIResult<()> = match input {
-                            Ok(input) => {
-                                let language = cache.fetch(&input).await?;
-                                let output = OutputFile::try_from(&input)?;
```
mkatychev, line 63: FYI this should work once the `try` block  is stabilized:
https://doc.rust-lang.org/nightly/unstable-book/language-features/try-blocks.html

```diff
-                           Ok(input) => {
+                           Ok(input) => try {
                                let language = cache.fetch(&input).await?;
                                let output = OutputFile::try_from(&input)?;
```

topiary-cli/src/main.rs
```diff
@@ -120,6 +107,11 @@ async fn run() -> CLIResult<()> {
                 results.remove(0)??
             } else if results
                 .iter()
+                .inspect(|r| match r {
+                    Err(e) => print_error(&e),
+                    Ok(Err(e)) if !e.benign() => print_error(&e),
```
mkatychev, line 112: This is actually a bug because `.any` will be lazily evaluated thus only the first error will be printed, this is fixed in the check https://github.com/tweag/topiary/pull/1094 by using `.count()` instead of `.any()`:

https://github.com/mkatychev/topiary/blob/ad580dded8533f76aa73853e32e5c3bc9fc81533/topiary-cli/src/io.rs#L488-L497

Xophmeister (commented):
I confident that your change is correct and solves this problem, but I don't quite follow what that problem is. Could you ELI5?

Xophmeister (approved):
Right: I think I understand now. Thanks :pray: So because the try operator is bubbling up to the `scope_and_block`, that ends the function prematurely, rather than just the async scope in which its invoked. Whereas what we want it to collect any error from each async scope and gather them together after they've all completed.

mkatychev:
@Xophmeister exactly it, I was scratching my head at this myself, to actually catch the try operator in a let binding we'd have to use the unstable `try` keyword:

```rs
fn foo()  {
    // adding the `try` keyword will actually catch the `try_this()?` in `let res`
    let res: Result<(), _> = try {
       try_this()?; 
       Ok(())
    };
}
```

PR #1095, opened 2025-09-25
chore: Update to Rust 2024 edition

mkatychev:
# Update to Rust 2024 edition

References https://github.com/tweag/topiary/pull/1094

## Description
https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/

Handle clippy lints arising from the edition change

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace.package]
 version = "0.6.1"
-edition = "2021"
+edition = "2024"
 authors = ["Tweag"]
 homepage = "https://topiary.tweag.io"
 repository = "https://github.com/tweag/topiary"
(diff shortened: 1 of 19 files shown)


Xophmeister:
I know our Nix code is using Crane to orchestrate the Rust toolchain, but I don't see any references in the source to toolchain version or edition..._except_ for `wasmRustVersion` in `nix/packages/topiary.nix`.

This is a long-shot, because the WASM target isn't being built by the CI, but _maybe_ that is the root of the issue. (It's currently hardcoded to `1.77.2`, whereas Rust 2024 needs at least `1.85`.) I really don't think the disabled WASM builds would affect the native builds, though.

`rust-overlay` is used to provide the Rust toolchain itself, but I have no reason to believe that it doesn't bring in the latest stable version, which would definitely provide edition 2024.

@yannham Do you have any insight into this?

mkatychev:
@Xophmeister nice catch on the `pkgs'`, that seems to have fixed the macOS build issues.

PR #1097, opened 2025-09-29
Bump topiary-tree-sitter-facade to fix wasm-bindgen conflict

Xophmeister:
Just trying to get all our ducks in a row to push the updates Topiary queries and config from #1085


mkatychev (approved):
LGTM

PR #1132, opened 2025-11-26
fix(config) Set `default`s for every relevant language property

mkatychev:
# Set `default`s for every relevant language property

Resolves https://github.com/tweag/topiary/issues/1131

## Description

Fixed <code>missing field `extensions`</code> config merge error when merging the built-in config with a configuration that has only one sub property such as the one below:

```nickel
{ languages = {	bash.indent | priority 1 = "	" } }
```

* added `--merge` alias to `--merge-config` flag
* added CI tests for config regressions

[PR Description]

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -1,4 +1,0 @@
-Cargo.lock
-!.cargo
-!.github
-


topiary-config/languages.ncl
Xophmeister: Is my understanding correct here: The `default` annotation in Nickel is not inherited by its children, which is why this bug has come up?

In which case, wouldn't we also need to annotate, say, `languages.<LANGUAGE>.grammar.source.*` with `default`?

Xophmeister: That's a very good point: It's (probably) never going to happen :+1:

Xophmeister:
Thanks, @mkatychev :slightly_smiling_face: All looks good to me. Feel free to merge and I will do the release and push to crates.io in the morning.

PR #1139, opened 2025-12-03
feat(error,cli): Added `rootcause::Report` handling in topiary-cli

mkatychev:
# Added `rootcause::Report` handling in topiary-cli.

References https://github.com/blindFS/topiary-nushell/issues/45

## Description

Add rich error reporting for failures in the formatting query file for a given language in the CLI.

* Multiple errors can now be collected at once using `rootcause::ReportCollection`:
  https://github.com/topiary/topiary/blob/5046777293110be57d82901104851e1418f67481/topiary-cli/src/io.rs#L619-L623
* Removed `unsafe impl Send/Sync for TopiaryError {}` using a preformatting approach instead to handle thread safety.
* Flattened `TopiaryError` by removing `CLIError` and including the associated errors in the `rootcause::Report` error stack.
* `topiary_cli::fs::traverse` now collects all errors as a direcotry tree is walked.

## Checklist

- [X] `CHANGELOG.md` updated
- [X] Documentation (The Topiary Book, `README.md`, etc.) up-to-date


topiary-tree-sitter-facade/src/range.rs
```diff
@@ -24,6 +26,25 @@ mod native {
...
+        pub fn new_linewise(source: impl AsRef<str>, start_byte: u32, start_point: &Point) -> Self {
+            let end_column = source
+                .as_ref()
+                .lines()
+                .nth(start_point.row() as usize - 1)
+                .map(|l| l.len())
+                .unwrap_or(1);
+            let end_point = Point::new(start_point.row(), end_column as u32);
+            let end_byte =
+                SourceOffset::from_location(source, start_point.row() as usize, end_column)
```
mkatychev, line 42: should this call be inlined so as to avoid bringing miette into facade?

topiary-cli/src/fs.rs
```diff
@@ -43,11 +45,15 @@ impl FileMeta {
...
     fn new<P: AsRef<Path>>(path: &P) -> CLIResult<Self> {
         // Stat a potential symlink
-        let lmeta = fs::symlink_metadata(path)?;
+        let lmeta = fs::symlink_metadata(path).context_to()?;
         let symlink = lmeta.is_symlink();
 
         // Follow the symlink, if necessary
-        let meta = if symlink { fs::metadata(path)? } else { lmeta };
+        let meta = if symlink {
+            fs::metadata(path).context_to()?
```
mkatychev, line 53: See https://github.com/rootcause-rs/rootcause/blob/9a9fc4e89add242401de8af057c1960fb68d53fb/examples/error_coercion.rs
about possibly doing away with `.context_to()` if we want to switch to untyped reports.

topiary-core/src/lib.rs
```diff
@@ -164,7 +165,7 @@ pub enum ScopeCondition {
 }
 
 /// A convenience wrapper around `std::result::Result<T, FormatterError>`.
-pub type FormatterResult<T> = std::result::Result<T, FormatterError>;
+pub type FormatterResult<T> = std::result::Result<T, rootcause::Report<FormatterError>>;
```
mkatychev, line 168: `rootcause::Report` supports from/to conversion with `anyhow::Error` if this is a need at the library level:
https://docs.rs/rootcause/latest/rootcause/compat/anyhow1/trait.IntoAnyhow.html#tymethod.into_anyhow

topiary-tree-sitter-facade/Cargo.toml
```diff
@@ -12,6 +12,9 @@ description = """
   Topiary's uniform interface for `tree-sitter` (rust) and `web-tree-sitter` (wasm-bindgen)
 """
 
+[dependencies]
+miette.workspace = true
+
```
ErinvanderVeen, line 17: Adding miette as a dependency, just for the `SourceOffset::from_location` function is painful, but I understand why you do it.

ErinvanderVeen (commented):
Generally speaking, I think this PR is definitely a good improvement for Topiary. I hesitate to to change type of `FormatterResult`, forcing downstream consumers to incorporate another dependency.

But optionally, we could offer a variant of formatter that produces a `Box<dyn Error>` using https://docs.rs/rootcause/latest/rootcause/struct.Report.html#impl-From%3CReport%3CC,+O,+Local%3E%3E-for-Box%3Cdyn+Error%3E

topiary-cli/src/cli.rs
```diff
@@ -257,25 +258,25 @@ pub fn get_args() -> CLIResult<Cli> {
                 file: Some(file), ..
             },
             ..
-        } => {
+        }
             // Make sure our FILE is not a directory
-            if file.is_dir() {
-                return Err(TopiaryError::Bin(
-                    format!(
+            if file.is_dir() => {
```
Xophmeister, line 271: It took me a while to parse this! Is it normal to put the match guard condition on a new, indented line like this?

topiary-core/src/error/mod.rs
```diff
@@ -105,118 +87,123 @@ impl fmt::Display for FormatterError {
...
-// NOTE: Filesystem-based IO errors _ought_ to become a thing of the past, once
-// the library and binary code have been completely separated (see Issue #303).
-impl From<io::Error> for FormatterError {
-    fn from(e: io::Error) -> Self {
-        IoError::from(e).into()
-    }
-}
+// private convenience macro to do [`rootcause::ReportConversion`]
+// https://docs.rs/rootcause/latest/rootcause/trait.ReportConversion.html
+macro_rules! report_conversion {
```
Xophmeister, line 94: Is this macro the same as the one in `topiary-cli/src/error.rs`? I wonder if there's an opportunity to lift out all your nifty error handling stuff into its own crate in the workspace, so it makes everything a bit more tidy? (Just a thought; I don't know if that would even make sense.)

topiary-tree-sitter-facade/src/point.rs
```diff
@@ -10,19 +9,17 @@ mod native {
...
-            let row = row as usize;
-            let column = column as usize;
-            tree_sitter::Point { row, column }.into()
+            tree_sitter::Point::new(row as usize, column as usize).into()
         }
 
         #[inline]
         pub fn column(&self) -> u32 {
-            u32::try_from(self.inner.column).unwrap()
+            self.inner.column as u32
```
Xophmeister, line 18: It might be worth adding a comment to acknowledge that we don't expect files to have more than $2^{32}$ lines or columns...if we do, then formatting is probably the least of our worries!

Xophmeister (commented):
This is a labour of love! Thank you, @mkatychev for all your work on it.

The downstream usage of the changed error API looks like a pretty simple change, which is good. The new error API definitions though are kind of gnarly and I won't pretend to fully understand them. I made a suggestion about lifting it out into its own crate, in an attempt to declutter/simplify, but I genuinely don't know if that's possible or if it even makes sense.

The improved UX for Topiary is worthwhile, so I don't have any qualms about following this through to completion. However -- and I realise this is hand-wavy -- I think we also need to think about the DX, which I personally think Topiary is already pretty bad for.

Another thing that crossed my mind: This PR is enormous (for obvious reasons). I wonder, when it's complete, whether it would be possible to cherry-pick it into sub-PRs to make it easier to review?

Still, I don't want to diminish this. It's an amazing achievement :bow:

topiary-tree-sitter-facade/src/range.rs
Xophmeister: I think anything to reduce the dependency footprint is generally worth thinking about, if it can be done relatively cheaply.

topiary-tree-sitter-facade/src/point.rs
mkatychev: good callout!

topiary-cli/src/cli.rs
mkatychev: good question, I think that's just `rustfmt` doing its thing.

mkatychev:
> This is a labour of love! Thank you, @mkatychev for all your work on it.
> 
> The downstream usage of the changed error API looks like a pretty simple change, which is good. The new error API definitions though are kind of gnarly and I won't pretend to fully understand them. I made a suggestion about lifting it out into its own crate, in an attempt to declutter/simplify, but I genuinely don't know if that's possible or if it even makes sense.
> 
> The improved UX for Topiary is worthwhile, so I don't have any qualms about following this through to completion. However -- and I realise this is hand-wavy -- I think we also need to think about the DX, which I personally think Topiary is already pretty bad for.
> 
> Another thing that crossed my mind: This PR is enormous (for obvious reasons). I wonder, when it's complete, whether it would be possible to cherry-pick it into sub-PRs to make it easier to review?
> 
> Still, I don't want to diminish this. It's an amazing achievement 🙇

@Xophmeister good callouts, I am hoping we can shave down some of the gnarlyness before this leaves draft. I'm happy to walk through the functionality so we can have a better understanding of what is going on!

mkatychev:
@Xophmeister just to get more context, which parts feel most unergonomic for DX?

Some callouts where DX can be changed:
# Problems

  ## 1. having to invoke event hooks to get automatic attachments:
  * Event hooks allow report creation to automatically attach metadata derived from the contents of the inner `C` in `Report<C>`
    but it requires invoking hook creation before errors are propagated but makes `topiary-core` inconsistent if the end user does not call `Hooks::new()`:
    https://github.com/topiary/topiary/blob/8d07a36bbe860e7ae092374abea7ec2591ea3365/topiary-cli/src/main.rs#L36-L41
  * Hook logic lives separate from the type `QueryError` for instance:
    https://github.com/topiary/topiary/blob/8d07a36bbe860e7ae092374abea7ec2591ea3365/topiary-core/src/error/error_span.rs#L306-L313

  ### possible solution
  Doing explicit method calling and relegating hook behaviour to only miette labeling:
  ```rust
  impl QueryError {
    pub fn report_span(self) -> Report<Self> {
        let range = self.range;
        report!(self)
            .with_range(range)
            .with_language("tree_sitter_query");
    }
  }
  ```
  thus [callisites](https://github.com/topiary/topiary/blob/8d07a36bbe860e7ae092374abea7ec2591ea3365/topiary-core/src/tree_sitter.rs#L67-L69) are explicit and consistent:
  ```rust
  impl TopiaryQuery {
      pub fn new(
          grammar: &topiary_tree_sitter_facade::Language,
          query_content: &str,
      ) -> FormatterResult<TopiaryQuery, QueryError> {
          let query = Query::new(grammar, query_content)
              .map_err(|e: QueryError| -> Report<QueryError> {
                  e.report_span() // can be an impl on Result<_, QueryError> as well
              })
              .attach_source(query_content)?;

          Ok(TopiaryQuery {
              query,
              query_content: query_content.to_owned(),
          })
      }
      // ...
  }
  ```
  ...and move all hook logic into `topiary-cli`.

  ## 2. having to call `.context_to()?` for type conversions:
  Though this is handy to make explicit conversions, without [trait
  specialisation](https://github.com/aturon/rfcs/blob/d7441bf81582591233040161b7c926bb1db746b4/text/0000-impl-specialization.md#overlapping-impls-and-specialization),
  it has to be explicitly invoked and defined:
    https://github.com/topiary/topiary/blob/8d07a36bbe860e7ae092374abea7ec2591ea3365/topiary-cli/src/main.rs#L91-L100
  https://github.com/topiary/topiary/blob/80cd3028f006e6618a0423dd6819a33b982dc1b2/topiary-cli/src/error.rs#L31-L42

  ### possible solution
  Use type erasure similar to the behaviour of
  [`anyhow::Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast_ref)
  in the majority of places and do explicit type conversions only where it
  makes sense:

  ```rust
  use rootcause::{Report, markers::Dynamic, report};

  #[derive(PartialEq, Debug)]
  struct FooError;

  fn foo_typed() -> Result<(), Report<FooError>> {
      Err(report!(FooError))
  }

  fn foo_dynamic() -> Result<(), Report> {
      // implicit conversion from Report<FooError> to Report<Dynamic>; Dynamic is the default error
      // type for reports: Report<C = Dynamic>
      foo_typed()?;
      Ok(())
  }

  fn main() {
      if let Err::<(), Report<Dynamic>>(e) = foo_dynamic() {
          assert_eq!(e.downcast_current_context(), Some(&FooError));
      }
  }
  ```

  ## 3. having to call `.preformat_context()?` for errors that are not `Send + Sync`:
  https://github.com/topiary/topiary/blob/8d07a36bbe860e7ae092374abea7ec2591ea3365/topiary-cli/src/main.rs#L58-L59
  Currently we do explicit `unsafe impl Send + Sync` for `TopiaryError` to bypass the non send/sync nickel_lang_core error types:
    https://github.com/topiary/topiary/blob/80cd3028f006e6618a0423dd6819a33b982dc1b2/topiary-cli/src/error.rs#L31-L42

  ### possible solution
  This can be bypassed as well (with unsafe calls and wrappers) but may warrant discussion.

Xophmeister:
Thanks, @mkatychev for outlining these potential problems and solutions in so much detail :pray: I think the `.context_to` and `.preformat_context` calls are actually not too bad. For (1), I have no context -- pun not intended :wink: -- of Miette, but if the hook logic moves into the CLI, that makes some sense. I suppose my thought is whether `topiary-core` becomes tightly coupled to `topiary-cli`? That should be avoided. Expecting `topiary-core` users to depend on Miette is not unreasonable, as it's popular and well-know, but my general opinion is that our default position ought to be to avoid too many dependencies in core (or make them feature gated).

My original DX comment was pretty vague; apologies! If I can be clearer, I think what I'm trying to say is that my brain's capacity for information is particularly limited, so when I see a big module, it tends to make me uneasy. (In your defence, Rust is quite bad for this, with structs and impls everywhere!) Anyway, as well as what you've suggested here, do you think there's scope for sub-modularising things; or do you think further splitting will be too arbitrary and make things actually harder to follow?

mkatychev:
@Xophmeister I think moving miette and the hooks into `topiary-cli` makes perfect sense. Agreed that core and cli coupling should be avoided.

topiary-tree-sitter-facade/src/range.rs
mkatychev: I've thought through it, since it will be unlikely for facade to be brought in without topiary-core, we (and any lib users) will pay for compiling miette regardless.

I can't think of a scenario where facade would be used by itself, by perhaps I am mistaken.

EDIT: fixed it.

topiary-tree-sitter-facade/Cargo.toml
mkatychev: https://github.com/topiary/topiary/pull/1139/changes#r3155481495

I've thought through it in this comment, do you think it common to use facade without core?

topiary-core/src/error/mod.rs
mkatychev: 🤔 I'm not sure what to do hear, would the overhead of an extra crate be worth it?

topiary-cli/src/error.rs
```diff
@@ -1,68 +1,55 @@
...
-/// Something can safely be Send unless it shares mutable state with something
-/// else without enforcing exclusive access to it. TopiaryError does not have a
-/// mutable state.
-unsafe impl Send for TopiaryError {}
-
-/// # Safety
-///
-/// Something can safely be Sync if and only if no other &TopiaryError can write
-/// to it. Since our TopiaryError contains no mutable data, TopiaryError is Sync.
-unsafe impl Sync for TopiaryError {}
```
mkatychev, line 52: The `ResultPreformat` trait allows us to remove these `unsafe impl`s by using the [`rootcause::Report::preformat`](https://docs.rs/rootcause/0.12.1/rootcause/struct.Report.html#method.preformat) method:
https://github.com/topiary/topiary/blob/5b51db4a2475eb06abc3a5ad4ef7741d692f8446/topiary-cli/src/error.rs#L147-L149

topiary-cli/src/error.rs
```diff
@@ -208,16 +130,31 @@ pub trait Benign {
...
+            Ok(t) => Ok(t),
+            Err(e) => Err(report!(e).preformat()),
+        }
     }
 }
+
+fn iter_downcast_reports<T: 'static>(report: &Report<impl ?Sized>) -> impl Iterator<Item = &T> {
+    report
+        .iter_reports()
+        .filter_map(|r| r.downcast_current_context::<T>())
```
mkatychev, line 217: This will return `true` on _any_ `FormatterError::PatternDoesNotMatch` since we now have the possibility of deeper nesting, I think that's relatively safe as of now.

topiary-cli/src/io.rs
```diff
@@ -375,11 +380,11 @@ pub(crate) fn to_query_from_language(
         // matching file in a default location. As a final attempt, try the
         // builtin ones. Store the error, return that if we
         // fail to find anything, because the builtin error might be unexpected.
-        Err(e) => {
+        Err(_e) => {
             log::warn!(
                 "No query files found in any of the expected locations. Falling back to compile-time included files."
             );
-            to_query(&language.name).map_err(|_| e)?
+            to_query(&language.name).preformat_context()?
```
mkatychev: not sure about this case, now the behaviour is in reverse: the later error is propagated but the earlier is ignored.

Cargo.toml
```diff
@@ -51,7 +51,7 @@ predicates = "3.0"
 pretty_assertions = "1.3"
 prettydiff = { version = "0.8.0", default-features = false }
 rayon = "1.11.0"
-rootcause = "0.13.0"
+rootcause = "0.12.1"
```
mkatychev: Reverting for now since preformat was split into own crate in 0.13: https://docs.rs/rootcause-preformat/latest/rootcause_preformat/

topiary-cli/src/error.rs
```diff
@@ -208,16 +134,30 @@ pub trait Benign {
...
-impl Benign for TopiaryError {
-    #[allow(clippy::match_like_matches_macro)]
+impl<C> Benign for Report<C>
+where
+    C: ?Sized,
+{
     fn benign(&self) -> bool {
-        matches!(self, TopiaryError::Lib(r) if r.current_context() == &FormatterError::PatternDoesNotMatch)
+        iter_downcast_reports::<FormatterError>(self)
+            .any(|fmt_err| *fmt_err == FormatterError::PatternDoesNotMatch)
```
Xophmeister: Doesn't this imply that benignity is predicated on the existence of a single `PatternDoesNotMatch`; what if `Multiple` contains a genuine error?

topiary-cli/src/error.rs
```diff
@@ -72,134 +55,77 @@ impl fmt::Display for TopiaryError {
...
-
-impl From<TopiaryError> for ExitCode {
-    fn from(e: TopiaryError) -> Self {
-        let exit_code = match e {
-            // Things went well but Topiary needs to answer 'false' in a clean way: Exit 1
-            _ if e.benign() => 1,
-
-            // Check mode detected unformatted files: Exit 1
-            // This error is not benign, but we still need to answer `false` without resulting in a typical an error
-            TopiaryError::Bin(_, Some(CLIError::CheckFailed { .. })) => 1,
```
Xophmeister, line 105: This is removed, but not reinstated, so `CheckFailed` now exits 10, I think.

topiary-cli/src/error.rs
```diff
@@ -72,134 +55,77 @@ impl fmt::Display for TopiaryError {
...
+        // for more
+        if let Some(e) = rep.downcast_current_context::<TopiaryConfigError>() {
+            // I/O errors: Exit 3
+            code = match e {
+                TopiaryConfigError::FileNotFound(_)
+                | TopiaryConfigError::QueryFileNotFound(_)
+                | TopiaryConfigError::Io(_)
+                | TopiaryConfigError::Fetching(
+                    FetchError::Io(_) | FetchError::GrammarFileNotFound(_),
+                ) => 3,
```
Xophmeister: Isn't this unreachable, because config errors go through `.preformat_context()`, making the context `PreformattedContext`, rather than some `TopiaryConfigError` variant?

topiary-cli/src/error.rs
```diff
@@ -72,134 +55,77 @@ impl fmt::Display for TopiaryError {
...
+                // Idempotency parsing errors: Exit 8
+                FormatterError::IdempotenceParsing => 8,
+                _ => 10,
+            };
+            break;
+        }
+        if rep.downcast_current_context::<io::Error>().is_some() {
+            // I/O errors: Exit 3
+            code = 3;
+        }
```
Xophmeister, line 118: For the sake of consistency:
```suggestion
        if rep.downcast_current_context::<io::Error>().is_some() {
            // I/O errors: Exit 3
            code = 3;
            break;
        }
```

topiary-cli/src/cli.rs
```diff
@@ -245,13 +245,20 @@ pub fn get_args() -> CLIResult<Cli> {
...
             // (potential) files, as input sources. This is finally deduplicated to avoid
             // formatting the same file multiple times (e.g., in the case that a symlink points to
             // a file within the set, or if the same file is specified twice at the command line).
-            fs::traverse(files, *follow_symlinks)?;
+            fs::traverse(files, *follow_symlinks, &mut errs)?;
+
+            // if there are only errors and no files, we should propagate the given errors
+            if files.is_empty() && !errs.is_empty() {
+                return Err(errs.context(TopiaryError::Multiple).into());
+            }
```
Xophmeister, line 257: If, say, there was still one valid file and a bunch of errors, then whole `errs` collection will be dropped.

topiary-cli/src/fs.rs
```diff
@@ -132,7 +137,7 @@ pub fn traverse(files: &mut Vec<PathBuf>, follow_symlinks: bool) -> CLIResult<()
         if is_dir {
             // Descend into directory, symlink-aware as required
             let mut subfiles = file.read_dir()?.flatten().map(|f| f.path()).collect();
-            traverse(&mut subfiles, follow_symlinks)?;
+            traverse(&mut subfiles, follow_symlinks, errs)?;
```
Xophmeister, line 140: Nitpick: The `?`s on `read_dir` and `traverse` will exit early, rather than collecting _all_ errors during the traversal. That's probably fine, tbh.

topiary-cli/src/cli.rs
mkatychev: Yep, that's how the current behaviour works (though they are logged), should we change that?

topiary-cli/src/error.rs
mkatychev: That's a good point, then we should handle `Multiple` and ensure that it's benign only when:
* `Multiple` with count of 1 and the type is `PatternDoesNotMatch`
* The is an `PatternDoesNotMatch` an no `Multiple`

topiary-cli/src/error.rs
mkatychev: good catch

topiary-cli/src/error.rs
mkatychev: Yes, hence the comment in https://github.com/topiary/topiary/blob/f4769b061d4bd4a04b2a479d5cbbe86aeed50a54/topiary-cli/src/error.rs#L100-L103

I figured this logic was worth preserving and better could be spent preformatting send/sync unsafe errors inside of `topiary_config` proper.

topiary-cli/src/fs.rs
mkatychev: That's how the current behaviour is I thought. Might be worth adding a comment regardless.

ErinvanderVeen (approved):
Approved after the TODO is resolved about the `preformat()` call erasing `TopiaryConfigError`.

topiary-cli/src/error.rs
Xophmeister: Right; so the match arm is just there because the compiler needs it to be exhaustive? Maybe it would be worth using `unreachable!()` to make it clearer.

topiary-cli/src/fs.rs
Xophmeister: Hence the nitpick :sweat_smile: I'm probably going to live to regret saying this, but I suspect `read_dir` doesn't fail all that often...or when it does, formatting code is probably the least of your worries!

topiary-cli/src/cli.rs
Xophmeister: Ah, gotcha -- my mistake -- no, I think this is fine :+1:

Xophmeister (approved):

Xophmeister:
Thanks, @mkatychev :bow: I think it's worth addressing the `.any` thing, but otherwise good to go!

mkatychev:
@Xophmeister @ErinvanderVeen I actually wasn't too happy with the exit code logic being untested after you guys pointed out my gaps, I added three test cases to handle preformatted errors and multi-errors (and handled `ReportCollection` directly in exit code now [removed `TopiaryConfig::Multiple`] as a result) to make sure the PR isn't sloppy. 

The changes are non trivial but the tests should be very readable.

topiary-cli/src/error.rs
```diff
@@ -72,152 +52,217 @@ impl fmt::Display for TopiaryError {
...
-impl From<TopiaryError> for ExitCode {
-    fn from(e: TopiaryError) -> Self {
-        let exit_code = match e {
-            // Things went well but Topiary needs to answer 'false' in a clean way: Exit 1
-            _ if e.benign() => 1,
+pub(crate) fn exit_code<C>(r: &Report<C>) -> ExitCode
+where
+    C: ?Sized,
+{
+    // `Report::as_ref` is a non-trait method (no `AsRef<ReportRef>`)
```
mkatychev, line 69: had to introduce error types because `impl AsRef<ReportRef>` does not exist :(

topiary-cli/src/error.rs
```diff
@@ -72,152 +52,217 @@ impl fmt::Display for TopiaryError {
...
-}
 
-pub(crate) fn print_error(e: &dyn error::Error) {
-    log::error!("{e}");
-    if let Some(source) = e.source() {
-        log::error!("Cause: {source}");
+    #[test]
+    fn preformat_context_exit_code_10() {
+        let err: Result<(), TopiaryConfigError> =
+            Err(TopiaryConfigError::UnknownLanguage("nope".to_string()));
```
mkatychev:
```suggestion
            Err(TopiaryConfigError::UnknownLanguage("other (exit 10)".to_string()));
```

topiary-cli/src/error.rs
```diff
@@ -72,152 +52,217 @@ impl fmt::Display for TopiaryError {
...
+    use std::path::PathBuf;
 
-impl Benign for TopiaryError {
-    #[allow(clippy::match_like_matches_macro)]
-    fn benign(&self) -> bool {
-        matches!(self, TopiaryError::Lib(r) if r.current_context() == &FormatterError::PatternDoesNotMatch)
+    use super::*;
+
+    #[test]
+    fn preformat_context_fetching_io_exits_3() {
```
mkatychev:
```suggestion
    fn preformat_context_nested_io_exits_3() {
```

topiary-cli/src/fs.rs
mkatychev: S we do this as a follow-up? the PR as-is tries (yet fails :D) to preserve existing behaviour where it can.

topiary-cli/src/error.rs
```diff
@@ -77,6 +84,17 @@ where
...
+                break;
+            }
+            if let Some(only) = collection.iter().next() {
+                return exit_code_inner(only.into_uncloneable());
+            }
+            continue;
+        }
         if let Some(e) = rep.downcast_current_context::<FormatterError>() {
             code = match e {
                 // TODO check failed
```
Xophmeister: I think this TODO can be removed now, right?

topiary-cli/src/error.rs
```diff
@@ -182,3 +217,52 @@ where
         .iter_reports()
         .filter_map(|r| r.downcast_current_context::<T>())
 }
+
+#[cfg(test)]
+mod tests {
```
Xophmeister, line 221: These tests are great :+1: I think it's worth adding a quick `CheckFailed => 1` test, but otherwise, very nice 💛

Xophmeister (approved):
Awesome :tada:

topiary-cli/src/fs.rs
Xophmeister: Nah, life's too short :) I think it's fine to gloss over this technicality. I'll resolve.

topiary-cli/src/error.rs
mkatychev: yes it can!

PR #1212, opened 2026-04-18
Clean up unused dependencies and incompatible features

bbannier:
This PR removes unneeded dependencies, and cleans up broken interactions between different features. I looked into this since I was trying to reduce the footprint of `topiary-core` which I use in https://github.com/bbannier/spicy-format, but this PR ended up a minor cleanup.

The commits in this PR should be independent, and if you'd prefer I can submit them separately.


ErinvanderVeen (approved):

ErinvanderVeen:
This is super nice! Thank you!

PR #1214, opened 2026-04-20
Cargo.{lock,toml}: try floating `wasm-bindgen`

### LINKED ISSUE #1128: `wasm-bindgen` dependency is hard-pinned
**Describe the bug**
The fact that it's pinned to `=0.2.100` forces other libraries using eg. `topiary-tree-sitter-facade` to hard-pin this dependency as well. My only option is to pin this version in all my crates in the same workspace.

**To Reproduce**
- Run `cargo install wasm-bindgen-cli`.
- Add `wasm-bindgen` and `topiary-tree-sitter-facade` to a crate's dependencies (or even different crates in the same workspace).
- Try to build the wasm-bindgen project.

**Expected behavior**
Building with wasm-bindgen should work, but instead it fails as the version from wasm-bindgen-cli (currently 0.2.105) doesn't match the pinned version.

**Environment**
 - OS name + version: Pop!_OS 22.04 LTS
 - Version of the code: 0.7.1

twitchyliquid64:
# Cargo.{lock,toml}: try updating `wasm-bindgen`

## Description

Fixes: #1128 

Tries to float `wasm-bindgen` to fix broken deps in places where topiary crates are used, i.e.

```
error: failed to select a version for `wasm-bindgen`.
    ... required by package `js-sys v0.3.85`
    ... which satisfies dependency `js-sys = "^0.3.85"` of package `wasm-streams v0.5.0`
    ... which satisfies dependency `wasm-streams = "^0.5"` of package `reqwest v0.13.2`
    ... which satisfies dependency `reqwest = "^0.13"` of package <snip>
    ... which satisfies path dependency <snip>
versions that meet the requirements `=0.2.108` are: 0.2.108

all possible versions conflict with previously selected packages.

  previously selected package `wasm-bindgen v0.2.100`
    ... which satisfies dependency `wasm-bindgen = "=0.2.100"` of package `topiary-tree-sitter-facade v0.7.3`
    ... which satisfies dependency `topiary-tree-sitter-facade = "^0.7.3"` of package `topiary-core v0.7.3`
    ... which satisfies dependency `topiary-core = "^0.7.3"` of package `nickel-lang-core v0.17.0`
    ... which satisfies dependency `nickel-lang-core = "^0.17.0"` of package `minimal v0.3.0 (/home/xxx/minimal/crates/minimal)`

failed to select a version for `wasm-bindgen` which could resolve this conflict
```

## Checklist

- [ ] `CHANGELOG.md` updated


Xophmeister:
I'm a bit out-of-the-loop with this, but IIRC Nickel is pinned to this old version of `wasm-bindgen`. That is, the source of the problem is upstream.

Xophmeister:
Right: So this is in Nickel's `cargo.toml`:

```toml
# The wasm-bindgen version is pinned using `=` since flake.nix reads the version
# number from Cargo.lock and needs to have matching output hashes for the source
# derivation. Combined with caching this would lead to silent inconsistent builds
# if this version changes without the nix output hashes being updated. It's okay
# to bump the version (for example if not doing so prevents some dependency from
# building) but flake.nix needs to be kept in sync.
wasm-bindgen = "=0.2.100"
```


https://github.com/nickel-lang/nickel/blob/d2a66b23eb8563213da8892384dda944e88dfeb4/Cargo.toml#L41-L47

twitchyliquid64:
Gotcha, my bad. Will try and get it updated there then.

Xophmeister:
> Gotcha, my bad.

Oh no, not at all. Thank you for taking this on :pray:

> Will try and get it updated there then.

:+1:

Xophmeister:
Superseded by #1255

PR #1217, opened 2026-04-23
add `--check` flag to `topiary fmt` for CI formatting verification

### LINKED ISSUE #977: Add `--check` to `topiary format`
I was asked about a `--check` feature that would check the validity of formatting in CI:

> It doesnt look like topiary has a way to check for formatting like `rustfmt --check`, that would be nice to have in CI jobs. do you have any context with that project to know why its absent or if they'd be open to adding it?

Allow `topiary format --check <path/to/file>` to return an exit code and possibly a diff:
```diff
$ topiary format --check ./topiary-cli/src/language.rs
Diff in topiary/topiary-cli/src/language.rs:1:
 use std::{
     collections::{
-        hash_map::{DefaultHasher, Entry},
         HashMap,
+        hash_map::{DefaultHasher, Entry},
     },
     hash::{Hash, Hasher},
     sync::Arc,
Diff in topiary/topiary-cli/src/language.rs:8:
 };

 use tokio::sync::Mutex;
-    use topiary_core::Language;
+use topiary_core::Language;

 use crate::{error::CLIResult, io::InputFile};
```

ErinvanderVeen:
Allow `topiary fmt --check` to verify files are already formatted without modifying them. Exits 0 if formatted, exits 1 with a diff if not. 
BREAKING CHANGE: The `check` alias was removed to avoid confusion with `fmt --check`.

# add `--check` flag

Resolves #977 

## Description
Allow `topiary fmt --check` to verify files are already formatted
without modifying them. Exits 0 if formatted, exits 1 with a diff
if not.

BREAKING CHANGE: The `check` alias was removed to avoid confusion with
`fmt --check`.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ This name should be decided amongst the team before the release.
 
 ### Added
 - [#1200](https://github.com/topiary/topiary/pull/1200) Build and deploy Topiary Docker images to ghcr.io.
+- [#1217](https://github.com/topiary/topiary/pull/1217) Add `--check` flag to `topiary fmt` for CI formatting verification.
+
+### Removed
+- [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
 
 <!--
 ### Changed
--- a/topiary-cli/src/cli.rs
+++ b/topiary-cli/src/cli.rs
@@ -122,6 +122,10 @@ pub enum Commands {
     /// Format inputs
     #[command(alias = "fmt", display_order = 1)]
     Format {
+        /// Verify inputs are already formatted (exit non-zero with diff if not)
+        #[arg(short = 'c', long)]
+        check: bool,
+
         /// Consume as much as possible in the presence of parsing errors
         #[arg(short, long)]
         tolerate_parsing_errors: bool,
@@ -182,7 +186,7 @@ pub enum Commands {
     },
 
     /// Check if an input parses to the respective Tree-sitter grammar
-    #[command(alias = "check", display_order = 6)]
+    #[command(display_order = 6)]
     CheckGrammar {
         #[command(flatten)]
         inputs: AtLeastOneInput,
--- a/topiary-cli/src/error.rs
+++ b/topiary-cli/src/error.rs
@@ -1,4 +1,6 @@
 use std::{error, fmt, io, path::PathBuf, process::ExitCode, result};
+
+use similar::TextDiff;
 use topiary_config::error::{TopiaryConfigError, TopiaryConfigFetchingError};
 use topiary_core::FormatterError;
 
@@ -24,6 +26,13 @@ pub enum CLIError {
     Multiple,
     UnsupportedLanguage(String),
 
+    /// Formatting check failed: input is not already formatted
+    CheckFailed {
+        source_name: String,
+        original: String,
+        formatted: String,
+    },
+
     /// Could not detect the input language from the `(filename, Option<extension>)`
     LanguageDetection(PathBuf, Option<String>),
 }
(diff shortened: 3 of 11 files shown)


topiary-cli/src/error.rs
```diff
@@ -45,6 +54,27 @@ impl fmt::Display for TopiaryError {
...
+                    source_name,
+                    original,
+                    formatted,
+                }),
+            ) => {
+                let diff = diff_lines(original, formatted);
+                write!(
+                    f,
+                    "Diff in {source_name}:\n{}",
+                    diff.format_with_context(
```
mkatychev: One of the issues I encountered with prettydiff is that it does not highlight show whitespace related diffs cleanly (newlines especially).

topiary-cli/src/main.rs
```diff
@@ -45,49 +46,64 @@ async fn run() -> CLIResult<()> {
     // Delegate by subcommand
     match args.command {
         Commands::Format {
+            check,
```
mkatychev: Could we move `format --check` into a separate match arm?
```suggestion
            check: true,
```

topiary-cli/src/cli.rs
```diff
@@ -182,7 +186,7 @@ pub enum Commands {
     },
 
     /// Check if an input parses to the respective Tree-sitter grammar
-    #[command(alias = "check", display_order = 6)]
```
Xophmeister, line 185: TIL that `check-grammar` is not documented... I'll add an issue.

topiary-cli/tests/sample-tester.rs
```diff
@@ -132,6 +132,58 @@ mod test_fmt {
     }
 }
 
+#[cfg(test)]
+mod test_check {
```
Xophmeister, line 136: I think your new `test_check` makes `test_fmt` kind of redundant -- it's doing the same thing -- so it could be removed or somehow consolidated.

CHANGELOG.md
```diff
@@ -54,6 +54,10 @@ This name should be decided amongst the team before the release.
 
 ### Added
 - [#1200](https://github.com/topiary/topiary/pull/1200) Build and deploy Topiary Docker images to ghcr.io.
+- [#1217](https://github.com/topiary/topiary/pull/1217) Add `--check` flag to `topiary fmt` for CI formatting verification.
+
+### Removed
+- [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
```
Xophmeister, line 60: As this is a breaking change, I guess it'll instigate a 0.8 release :tada:

Xophmeister (approved):
LGTM 🙂 I've left a few suggestions and I agree that breaking out the `check: true` match arm would make the code a little easier to read.

Thanks :pray:

mkatychev (approved):
Thanks for getting this over the line!

topiary-cli/src/error.rs
ErinvanderVeen: I hadn't noticed! But I've switched to `similar` now. It's not as pretty, but does follow the "standard" diff format, which is nice.

CHANGELOG.md
ErinvanderVeen: Yep!

topiary-cli/tests/sample-tester.rs
ErinvanderVeen: `test_fmt` formats the input files, and then diffs it against the output files, ensuring the formatting is correct.

`test_check` runs `--check` on the input/expected files, but notably, does not check that the formatting step itself is correct. I.e. `--check input/json.json` can correctly fail without formatting to `output/json.json` even though `--check output/json.json` does succeed.

Xophmeister: Good point :sweat_smile:

PR #1220, opened 2026-04-24
small language fixes i found reading the documentation

prednaz:
## Description

- [small language fixes](https://github.com/topiary/topiary/pull/1220/changes/7219e24cb7d0b069d0355829e61f4d0611d0008a)
- [replaced `--` with `–` because `--` is rendered as "--"](https://github.com/topiary/topiary/pull/1220/changes/ec03433b207c7ff9fec71c45a66c7352ea9ef1a9)

## Checklist

- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/book/src/getting-started/on-tree-sitter.md
+++ b/docs/book/src/getting-started/on-tree-sitter.md
@@ -90,7 +90,7 @@ exceptional rules like this exist, they can easily be accommodated.
 
 ### Topiary language query files
 
-A language query file, usually given then `.scm` extension, is a
+A language query file, usually given the `.scm` extension, is a
 collection of such queries with appropriate formatting capture names.
 Taken in aggregate, when applied to source code files by Topiary, they
 define a formatting style for the language in question.
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -30,11 +30,11 @@ pub struct AtomCollection {
     /// A flat list of all Atoms. This is is updated by some formatting
     /// directives, but most require some more complexity.
     atoms: Vec<Atom>,
-    /// Whenever a formatting directive instructs tree-sitter to prepend
+    /// Whenever a formatting directive instructs topiary to prepend
     /// something to a node, a new Atom is added to this HashMap.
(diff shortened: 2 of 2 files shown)


docs/book/src/getting-started/installation/using-nix.md
```diff
@@ -2,7 +2,7 @@
 
 Topiary provides a flake for easy (re)use via Nix. It is also possible to use
 Topiary's Nix code without going through flakes, although some inputs are harder
-to provide than others -- have a look at the `nix/` directory.
+to provide than others – have a look at the `nix/` directory.
```
yannham: Mmh, `--` is a valid markdown encoding of – (github properly renders this one for example if we view this markdown file in the UI), but it's true that it's not properly rendered in the book. I don't know if this is something we should fix on the book side instead (if fixable)

Xophmeister: Yeah, I think sticking with (mostly) ASCII in the markdown source is probably a better idea, otherwise we'll get a situation where there's a mixture of Unicode endashes and `--`s because, I-for-one am way too lazy to find out how to insert an endash efficiently while I'm typing :sweat_smile:

docs/book/src/index.md
```diff
@@ -27,7 +27,7 @@ Topiary follows this trend by aspiring to be a "universal formatter
 engine", which allows developers to not only automatically format their
 codebases with a uniform style, but to define that style for new
 languages using a [simple DSL][tree-sitter:query]. This allows for the
-fast development of formatters, providing a [Tree-sitter
+fast development of formatters, provided a [Tree-sitter
```
Xophmeister: FWIW, these are both considered correct, these days. However, according to my (very brief) research, "provided" is the OG form and may be considered more formal. So your change works here.

Xophmeister (changes requested):
Thanks, @prednaz :pray:

This looks good, but I think @yannham is right and that the endash/`--` distinction is better left as it was, for practicality's sake.

Xophmeister (approved):
Thanks, @prednaz :pray: Feel free to merge!

Issue #1142, opened 2025-12-04
Implement `topiary test`


mkatychev:
# Proposal for `topiary test`

## Thoughts 

* There's value to having test cases be in one huge file for side effects.
* One should not have to recompile the `topiary` binary to test query changes for built-in languages
* external grammar repositories such as [topiary-nushell](https://github.com/blindFS/topiary-nushell) should have comparable DX to what is happening in topiary

## Ideas 

* initial aim to have queries organized with comment delimiters similar to [corpus tests](https://tree-sitter.github.io/tree-sitter/creating-parsers/5-writing-tests.html#writing-tests)
in tree-sitter proper
* This doesn't establish a complete correlation whether a test query was successful but it will check to see if there was any difference in the before and after test doc for a particular query
  - this is different from `topiary coverage` because it would check if the string value changed when a particular query is run v.s. if a query found matches
* to handle something akin to [topiary-nushell's pattern](https://github.com/blindFS/topiary-nushell/tree/main/test) we can do something similar to nushell's [`parse`](https://www.nushell.sh/commands/docs/parse.html) command where we can define file ranges with a template string for _input_ and _expeceted_:
  `topiary test --input "test/input_{case}.nu" --expected "test/expected_{case}"`

Enumerating queries and test files is a many to many complexity but could be
an acceptable complexity for now since it will improve coverage over present state.

### Mockup

```bash
# ./my_lang/tes1.myl
:' 1. use native language line comment start followed by 4 or more `=` signs to
 2. next line test name
 3. new line equivalent `=` sign comment as 1.
 4. add final linebreak
'

# ==================
# return-statement
# ==================

ok() {
  yeet 1;    
} 

# ==================
# loop
# ==================

$one by-one {
  utter "the good die ${one}";
} 
```

Test report could look like this in stdout:
```sh-session
$ topiary test --dir ./my_lang/test --query ./my_lang/query.scm

FAILURE(S):
- test1.myl
  * return-statement (L10-L##)
  * loop (L19-L##)
```

Instead of comments, one could use [`query_name!`](https://topiary.tweag.io/book/reference/capture-names/general.html#query_name) but DX may be different.

Xophmeister:
> I feel like discussions are for pre-design phase stuff, I feel like this has been discussed

Fair point 👌

PR #1232, opened 2026-04-30
feat(core,error): `ErrorSpan` rework

mkatychev:
# `ErrorSpan` rework

Resolves #XXX
References https://github.com/topiary/topiary/pull/1139/ https://github.com/topiary/topiary/issues/1142
## Description

As a prelude to https://github.com/topiary/topiary/pull/1139/, rework `ErrorSpan` to handle `QueryErrors` as well as error nodes by introducing the [`rootcause::Report`](https://docs.rs/rootcause/latest/rootcause/struct.Report.html) error wrapper to `topiary-core`.

`rootcause::Report` allows us to attach metadata such as filepaths, file contents, and location of an error to any error types we handle through "attachments" in the error stack.

## Checklist

- [x] `CHANGELOG.md` updated


topiary-core/src/error/mod.rs
```diff
@@ -105,118 +67,103 @@ impl fmt::Display for FormatterError {
...
-        ))
-    }
-}
+                }
+            }
+        )+
+    };
+}
+
+report_conversion!(
```
mkatychev, line 105: @Xophmeister on https://github.com/topiary/topiary/pull/1139#discussion_r3138123615 a concern I have is that moving this into a separate crate would expose the logic since we would have to `macro_export`, ideally this should not be in the public API

mkatychev:
https://github.com/topiary/topiary/actions/runs/25179760471/job/73821559031?pr=1232#step:5:145
> error[E0658]: `if let` guards are experimental

Seems like the rust version in nix & windows needs to be bumped, this feature is in stable.

EDIT: I've polyfilled this until our flake & windows support 1.95

topiary-cli/src/error.rs
```diff
@@ -210,7 +212,9 @@ impl Benign for TopiaryError {
     #[allow(clippy::match_like_matches_macro)]
     fn benign(&self) -> bool {
         match self {
-            TopiaryError::Lib(FormatterError::PatternDoesNotMatch) => true,
+            TopiaryError::Lib(r) if r.current_context() == &FormatterError::PatternDoesNotMatch => {
+                true
+            }
             _ => false,
         }
```
Xophmeister: Nitpick: Would this be simpler?
```rust
matches!(self, TopiaryError::Lib(r) if r.current_context() == &FormatterError::PatternDoesNotMatch)
```
...or do you expect additional match arms in the future?

topiary-core/src/error/error_span.rs
```diff
@@ -5,51 +5,244 @@
...
 
-use miette::{Diagnostic, NamedSource, SourceSpan};
+use miette::{LabeledSpan, MietteError, MietteSpanContents, SourceCode, SourceSpan, SpanContents};
+use rootcause::markers::ObjectMarkerFor;
 use topiary_tree_sitter_facade::Range;
 
-use crate::tree_sitter::NodeSpan;
+/// ErrorSpan is meant to represent errors code that lives outside of the topiary
+/// call stack and is rendered with [`miette::Report`].
+/// Examples of files that generate  ErrorSpans (these are typically runtime objects):
```
Xophmeister: nitpick
```suggestion
/// ErrorSpan is meant to represent error codes that lives outside of the topiary
/// call stack and is rendered with [`miette::Report`].
/// Examples of files that generate ErrorSpans (these are typically runtime objects):
```

topiary-core/src/error/mod.rs
Xophmeister: Fair enough. Duplicating just once, I suppose, is not a stimulus for refactoring, so this is fine :+1:

Xophmeister (approved):
Thanks, @mkatychev :pray:

I approve :+1: but as the PR is long and it's late, maybe worth getting a second opinion! @ErinvanderVeen would you be able to spare some time to double check this?

topiary-cli/src/error.rs
mkatychev: I suppose we can deal with more arms when we grow them, thanks for suggestion!

topiary-core/src/error/error_span.rs
```diff
@@ -5,51 +5,244 @@
...
+where
+    ErrorSpan: ObjectMarkerFor<T>,
+{
+    fn attach_filepath(mut self, filepath: &Path) -> Self {
+        if let Some(span) = self.get_span() {
+            span.filepath = Some(filepath.to_owned());
+            return self;
+        }
+        self.attach(ErrorSpan::default().with_filepath(filepath))
+    }
```
mkatychev: We can also handle `Option<&Path>` as well as `&Path` to avoid certain gnarly `map_err` sections:
```rust
fn attach_filepath<'a>(mut self, filepath: impl Into<Option<&'a Path>>) -> Self {
        let Some(filepath) = filepath.into() else {
            return self;
        };
        if let Some(span) = self.get_span() {
            span.filepath = Some(filepath.to_owned());
            return self;
        }
        self.attach(ErrorSpan::default().with_filepath(filepath))
    }
```

Then the section below could be simplified:

## Before

https://github.com/topiary/topiary/blob/7b85bf1e2d47d3041c59352397cc818f751d9989/topiary-cli/src/main.rs#L230-L238

## After

```rust
    let coverage_data =
        check_query_coverage(&input_content, &language.query, &language.grammar)
            .attach_source(&input_content)
            .attach_filepath(buf_input.get_ref().filepath())?;
```


...though I don't know what the compilation time hit would be

topiary-core/src/error/mod.rs
```diff
@@ -105,118 +67,103 @@ impl fmt::Display for FormatterError {
...
+    std::str::Utf8Error | std::string::FromUtf8Error,
+    FormatterError::Io,
+    "Input is not valid UTF-8"
```
mkatychev, line 108: We can also remove these `ReportConversion` impls altogether if we want to propagate the error directly (without adding a `FormatterError` to the stack).

topiary-core/src/error/error_span.rs
Xophmeister: I think that's a nice improvement for downstream code. Is your concern that `impl Into<Option<&'a Path>>` may incur a combinatorial monomorphisation problem that will drag the compiler down?

mkatychev: yeah, though it should not be too bad

Xophmeister: I suspect it pales in comparison to the Nickel parser :wink:

topiary-cli/src/io.rs
```diff
@@ -156,7 +170,8 @@ impl InputFile<'_> {
     pub async fn to_language(&self) -> CLIResult<Language> {
         let grammar = self.language().grammar()?;
         let query_contents = self.query.get_content().await?;
-        let query = TopiaryQuery::new(&grammar, &query_contents)?;
+        let query =
+            TopiaryQuery::new(&grammar, &query_contents).context(FormatterError::Parsing)?;
```
ErinvanderVeen: Invalid Query files now produce the "Tree-sitter could not part input" error. Which is kinda true, but not really what we want to report.

```bash
printf '((document) @does_not_exist' > /tmp/topiary-bad-query.scm
printf '{}\n' | target/debug/topiary format -l json -q /tmp/topiary-bad-query.scm
```

topiary-cli/src/error.rs
ErinvanderVeen, line 223: I don't know if this is intended or not, but right now a failure in parsing using the CLI doesn't actually show a snippet.

```cmd
$ printf '{ bad json' > /tmp/topiary-bad.json
$ topiary format --skip-idempotence /tmp/topiary-bad.json
[2026-05-07T10:39:41Z ERROR topiary::error]
     ● Tree-sitter could not parse the input without errors.
     ├ topiary-core/src/tree_sitter.rs:459
     ╰ Parsing error between line 1, column 0 and line 1, column 10
```

topiary-core/src/tree_sitter.rs
```diff
@@ -61,9 +63,10 @@ impl TopiaryQuery {
     pub fn new(
         grammar: &topiary_tree_sitter_facade::Language,
         query_content: &str,
-    ) -> FormatterResult<TopiaryQuery> {
+    ) -> FormatterResult<TopiaryQuery, QueryError> {
         let query = Query::new(grammar, query_content)
-            .map_err(|e| FormatterError::Query("Error parsing query file".into(), Some(e)))?;
+            .into_report()
+            .attach_source(query_content)?;
```
ErinvanderVeen, line 73: The facade now stores QueryError.range, but here it only does .into_report().attach_source(query_content). So we get "missing range".

```cmd
$ printf '((document) @does_not_exist' > /tmp/topiary-bad-query.scm
$ printf '{}\n' | target/debug/topiary -vv format -l json -q /tmp/topiary-bad-query.scm
[2026-05-07T11:06:22Z ERROR topiary::error]
     ● Tree-sitter could not parse the input without errors.
     ├ topiary-cli/src/io.rs:174
     │
     ● Query error at 1:28. Invalid syntax:
     │ ((document) @does_not_exist
     │                            ^
     ├ topiary-core/src/tree_sitter.rs:68
     ╰ Parsing error: missing range
```

ErinvanderVeen (changes requested):
Hej!

Sorry for taking so long to review this PR. Overall, I'm happy with that the PR achieves, and I'm eager to get this into Topiary. Some concerns.

topiary-cli/src/io.rs
mkatychev: This should be fixed in https://github.com/topiary/topiary/pull/1232/commits/5944f21162069a1044b219718377467c0476c1fc, I think this was an oversight in splitting up https://github.com/topiary/topiary/pull/1139:

```sh-session
$ printf '{}\n' | target/debug/topiary format -l json -q /tmp/topiary-bad-query.scm
[2026-05-07T16:21:03Z ERROR topiary::error]
     ● Tree-sitter could not parse the input without errors.
     ├ topiary-cli/src/io.rs:174
     │
     ● Query error at 1:28. Invalid syntax:
     │ ((document) @does_not_exist
     │                            ^
     ├ topiary-core/src/tree_sitter.rs:68
     ╰ Parsing error between line 1, column 27 and line 1, column 27
```

~EDIT: though the line range should be better 🤔~

EDIT2: also missing filepath

topiary-cli/src/error.rs
mkatychev: This is an interesting thought, this was previously handled in the `Display` impl for `FormatterError` by calling `miette::Report::new` on `ErrorSpan`:
https://github.com/topiary/topiary/blob/35cd149ba7671e885538b5a1d9fb6f08d5001097/topiary-core/src/error/mod.rs#L98-L98

The follow-up PR will do the same here:
https://github.com/topiary/topiary/blob/2f870825901b31bd921c2680e01f0e31cd4f7047/topiary-cli/src/error/hooks.rs#L21-L30

`miette::Report` cannot be called on a `Display` impl on self without overflowing the stack but we do want to have access to the snippet without the CLI...

We could move the `write!(f, "{}", miette::Report::new(self.clone))` into the `Debug` impl, and have the display and debug hook above use the debug version. 

...though I think that's a bit unintuitive now that I think through it

EDIT3: I found [`AttachmentHandler`](https://docs.rs/rootcause/latest/rootcause/handlers/trait.AttachmentHandler.html) to be a good solution

topiary-core/src/tree_sitter.rs
mkatychev: Good catch, here's the new result from https://github.com/topiary/topiary/pull/1232/commits/a42e4fefebaa33bee5507f3f6dae8d0b7d44e44f:

```sh-session
$ printf '{}\n' | target/debug/topiary -vv format -l json -q /tmp/topiary-bad-query.scm
[2026-05-07T18:42:44Z INFO  topiary_config::source] Using unix-home-specified configuration: /Users/mkatychev/.config/topiary/languages.ncl
[2026-05-07T18:42:44Z ERROR topiary::error]
     ● Tree-sitter could not parse the input without errors.
     ├ topiary-cli/src/io.rs:184
     │
     ● Query error at 1:28. Invalid syntax:
     │ ((document) @does_not_exist
     │                            ^
     ├ topiary-core/src/tree_sitter.rs:68
     ╰ See Error Span #1 below

    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
     Error Span #1

       × Parsing error between line 1, column 28 and line 1, column 28
        ╭─[/tmp/topiary-bad-query.scm:1:28]
      1 │ ((document) @does_not_exist
        ╰────

    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

topiary-cli/src/error.rs
mkatychev: New result with https://github.com/topiary/topiary/pull/1232/commits/a42e4fefebaa33bee5507f3f6dae8d0b7d44e44f

```sh-session
$ ./target/debug/topiary format --skip-idempotence /tmp/topiary-bad.json
[2026-05-07T18:52:38Z ERROR topiary::error]
     ● Tree-sitter could not parse the input without errors.
     ├ topiary-core/src/tree_sitter.rs:463
     ╰ See Error Span #1 below

    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
     Error Span #1

       × Parsing error between line 1, column 1 and line 1, column 11
        ╭─[/private/tmp/topiary-bad.json:1:1]
      1 │ ╭─▶ { bad json
        ╰────

    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

mkatychev:
not sure whether nix is being flaky (🥁) or not but the CI failures are exclusive to it.

ErinvanderVeen (approved):

PR #1115, opened 2025-10-14
feat(core): support nested formatting

mkatychev:
# initial support for nested formatting 

Resolves #XXX
References #400, #568, #579
Depends on #ZZZ

## Description

[PR Description]

## Checklist

Checklist before merging, wherever relevant:

- [ ] `CHANGELOG.md` updated
- [ ] Documentation (The Topiary Book, `README.md`, etc.) up-to-date

- [ ] Updated regression tests

PR #1244, opened 2026-05-18
OCamllex/OCaml language injections!

### LINKED ISSUE #579: Language injections for OCamllex
It has been discussed before, but Topiary should be able to handle Language Injections (where Topiary could format a language embedded in another language). This feature was first discussed #400. We recently added ocamllex support in #573, which is the perfect opportunity to explore this further.

Ocamllex is a good candidate to experiment with language injection because it contains only a single, pre-determined, language injection (namely the `(ocaml)`nodes).

After we implement this issue, we could continue looking at other (more complicated) places. i.e. Nickel's `doc` or Markdown codeblocks.

ErinvanderVeen:
# Add language injections for OCamllex

Resolves #579
References #400
References #573

## Description

This adds language injection support to Topiary, with OCamllex as the first supported use case. OCamllex action blocks are detected with a new `injections.scm` query and formatted using the existing OCaml formatter before being spliced back into the host OCamllex output.

The implementation adds injection-query loading, injected-language resolution in the CLI, hard-error behavior when matched injected languages cannot be resolved or formatted, and core tests covering resolution failures, parse failures, idempotence, and successful injected formatting.

This also updates the Topiary Book with language-injection documentation and adds an OCamllex injection benchmark.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/docs/book/src/guides/adding-a-new-language.md
+++ b/docs/book/src/guides/adding-a-new-language.md
@@ -212,7 +212,7 @@ inner language:
 )
 ```
 
-The host language's `formatting.scm` still controls layout around the
+The host language's `formatting.scm` retains control of the layout around the
 captured node. The captured content is formatted independently by the
 inner language and then rendered as a host leaf. See
 [Language injections](../reference/language-injections.md) for the
--- a/docs/book/src/guides/adding-a-new-language.md
+++ b/docs/book/src/guides/adding-a-new-language.md
@@ -193,6 +193,31 @@ fn to_query<T>(name: T) -> CLIResult<QuerySource>
 This will allow your query file to by considered as the default fallback
 query, when no other file can be found at runtime for your language.
 
+## Optional: add language injections
+
+If the new language embeds source code from another Topiary language,
+add an `injections.scm` file next to `formatting.scm`:
+
+```sh
+touch topiary-queries/queries/clang/injections.scm
+```
+
+An injection query captures host syntax that should be formatted by the
+inner language:
+
+```scheme
+(
+  (embedded_code) @injection.content
+  (#injection_language! "inner_language")
+)
+```
+
+The host language's `formatting.scm` still controls layout around the
+captured node. The captured content is formatted independently by the
+inner language and then rendered as a host leaf. See
+[Language injections](../reference/language-injections.md) for the
+full behaviour and limitations.
+
 ## Iterate
 
 Once the above steps have been completed, Topiary will be able to use
--- a/docs/book/src/reference/language-injections.md
+++ b/docs/book/src/reference/language-injections.md
@@ -65,9 +65,11 @@ There is no host reparse after injected text is rewritten.
 
 ## Failure behaviour
 
-Injected formatting is best-effort. If an injected span cannot be
-formatted, Topiary logs a warning and preserves the original injected
-source text. The host formatter continues.
+Injected formatting is all-or-nothing. If an injection query matches,
+Topiary must resolve the injected language and format the captured span
+successfully. If the injected language cannot be resolved, its grammar
+or query files cannot be loaded, or the captured span cannot be
+formatted, formatting fails.
 
 Idempotence is still checked at the outer formatting level by default.
 Injected spans are formatted again during that second pass, so unstable
@@ -81,9 +83,9 @@ are cached by the existing language definition cache, so formatting many
 spans of the same injected language does not recompile the grammar or
 reload the same queries for every span.
 
-If the CLI cannot resolve the injected language at runtime, the span is
-preserved with a warning. This can happen when a language is missing
-from the runtime configuration or its query files cannot be found.
+If the CLI cannot resolve the injected language at runtime, formatting
+fails. This can happen when a language is missing from the runtime
+configuration or its query files cannot be found.
 
 ## Limitations
 
--- a/docs/book/src/reference/language-injections.md
+++ b/docs/book/src/reference/language-injections.md
@@ -1,0 +1,106 @@
+# Language injections
+
+Language injections let a host language delegate part of its input to
+another Topiary language. This is useful for languages that embed code
+from another language in a syntactically distinct region. For example,
+OCamllex embeds OCaml code in `(ocaml)` nodes, and Topiary can format
+that inner OCaml with the OCaml formatter while the OCamllex formatter
+continues to control the surrounding lexer syntax.
+
+Injection declarations live in an `injections.scm` file next to a
+language's `formatting.scm` file:
+
+```text
+topiary-queries/queries/<language>/formatting.scm
+topiary-queries/queries/<language>/injections.scm
+```
+
+The formatting query still describes host-language layout. The
+injection query only describes which host syntax nodes should be
+formatted by another language.
+
+## Query schema
+
+The initial injection schema supports a captured content node and a
+static injected language name:
+
+```scheme
+(
+  (ocaml) @injection.content
+  (#injection_language! "ocaml")
+)
+```
+
+The `@injection.content` capture marks the host node whose source text
+will be formatted by the injected language. The
+`#injection_language!` predicate declares the Topiary language
+identifier to use for that captured content.
+
+Patterns without `#injection_language!` are skipped with a warning.
+Patterns without `@injection.content` do not inject anything.
+
+Dynamic language selection, such as taking the language name from a
+Markdown code fence, is not supported yet. It is future work for
+broader multi-language document support.
+
+## Formatting model
+
+Topiary parses the host input once. If the host language has an
+`injections.scm` file, Topiary runs that query against the host tree and
+collects the captured spans. Those captured nodes are then treated as
+host leaves while the normal host formatting query is applied.
+
+After host atomisation, each injected span is formatted independently
+with its resolved inner language. The resulting text replaces the
+corresponding host leaf before normal atom post-processing and pretty
+printing continue.
+
+Inner formatters produce text from column zero. The host renderer is
+still responsible for applying indentation at the point where the
+injected leaf is rendered. This works well when the host grammar has a
+natural boundary around the injected code, such as OCamllex action
+blocks.
+
+There is no host reparse after injected text is rewritten.
+
+## Failure behaviour
+
+Injected formatting is best-effort. If an injected span cannot be
+formatted, Topiary logs a warning and preserves the original injected
+source text. The host formatter continues.
+
+Idempotence is still checked at the outer formatting level by default.
+Injected spans are formatted again during that second pass, so unstable
+injected formatting can still make the top-level idempotence check fail.
+
+## CLI behaviour
+
+The CLI resolves injected language names through the normal language
+configuration and query discovery paths. Compiled language definitions
+are cached by the existing language definition cache, so formatting many
+spans of the same injected language does not recompile the grammar or
+reload the same queries for every span.
+
+If the CLI cannot resolve the injected language at runtime, the span is
+preserved with a warning. This can happen when a language is missing
+from the runtime configuration or its query files cannot be found.
+
+## Limitations
+
+Injected spans are formatted independently. Current injection formatting
+cannot express layout decisions that need to measure or choose between
+softline layouts across the host/injected boundary.
+
+The model is also stateless. The injected language is determined by the
+injection query match itself, not by earlier host-language context. For
+example, a host syntax where one declaration changes the language of a
+later heredoc is outside the current model.
+
+The injection query should not produce overlapping or nested
+`@injection.content` captures. If multiple patterns match the same
+region of code, Topiary will attempt to format each one. However,
+because an injected span is treated as a forced leaf in the host
+language, any nested captures will fail to find their corresponding leaf
+node during the rewrite phase, resulting in a formatting error. Query
+authors should ensure that injection patterns are mutually exclusive for
+any given span of text.
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -60,7 +60,13 @@ async fn run() -> CLIResult<()> {
                     input.query(),
                 );
 
-                check::check_input(input, &language, skip_idempotence, tolerate_parsing_errors)
+                check::check_input(
+                    input,
+                    &language,
+                    skip_idempotence,
+                    tolerate_parsing_errors,
+                    None, // TODO(Erin): Language Injection
+                )
             })
             .await?;
         }
@@ -101,6 +107,7 @@ async fn run() -> CLIResult<()> {
                             skip_idempotence,
                             tolerate_parsing_errors,
                         },
+                        None, // TODO(Erin): Language Injection
                     )?;
                 }
 
@@ -155,6 +162,7 @@ async fn run() -> CLIResult<()> {
                 Operation::Visualise {
                     output_format: format.into(),
                 },
+                None, // TODO(Erin): Language Injection
             )
             .attach_filepath(buf_input.get_ref().filepath())?;
         }
--- a/topiary-cli/src/main.rs
+++ b/topiary-cli/src/main.rs
@@ -14,8 +14,11 @@ use std::{
 
 use error::Benign;
 use tabled::{Table, settings::Style};
-use topiary_config::source::Source;
-use topiary_core::{Operation, SpanAttachment, check_query_coverage, formatter};
+use topiary_config::{Configuration, source::Source};
+use topiary_core::{
+    FormatterError, FormatterResult, Language, Operation, SpanAttachment, check_query_coverage,
+    formatter,
+};
 
 use crate::{
     cli::Commands,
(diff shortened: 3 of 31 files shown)


docs/book/src/guides/adding-a-new-language.md
```diff
@@ -193,6 +193,31 @@ fn to_query<T>(name: T) -> CLIResult<QuerySource>
...
+add an `injections.scm` file next to `formatting.scm`:
+
+```sh
+touch topiary-queries/queries/clang/injections.scm
+```
+
+An injection query captures host syntax that should be formatted by the
+inner language:
+
+```query
```
Xophmeister: Elsewhere in the book, we've used `scheme` for these code fences:
````suggestion
```scheme
````

File: docs/book/src/reference/language-injections.md [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+The formatting query still describes host-language layout. The
+injection query only describes which host syntax nodes should be
+formatted by another language.
+
+## Query schema
+
+The initial injection schema supports a captured content node and a
+static injected language name:
+
+```query
```
Xophmeister:
````suggestion
```scheme
````

File: topiary-cli/src/main.rs [hunk 3]
```diff
@@ -155,6 +203,7 @@ async fn run() -> CLIResult<()> {
                 Operation::Visualise {
                     output_format: format.into(),
                 },
+                None, // TODO(Erin): Language Injection
```
Xophmeister: Is this `TODO` for a later PR? IMO, visualising injections is going to make the graph _super_ complicated. Maybe it would be useful to have, but I'm not convinced. Perhaps a "stretch goal" 🙂

Xophmeister (commented):
Nice! Thank you, @ErinvanderVeen, this is awesome 💛

I've gone through all of this and I get the gist of it. However, my feeble brain can't follow it in its entirety, so to be conservative, I've not approved. However, I do "tentatively approve", but for such a big change -- and a big PR -- another set of eyes I think would be a good idea.

---

My biggest question is the resolver function. Your comment in `topiary-core/src/lib.rs` says:

> Resolves an injected language name to a Topiary language definition.

So that's a function that maps the language string to a `Language`? Does that mean every potential mapping needs to be hard-coded? I assume I'm missing something obvious, but why not reuse the `str -> Language` mapping used for formatting the top-level (i.e., from the file extension)?

(Minor question: Also is the `Arc` just in there to conform to the interface of the language cache? Doesn't that tie core to the CLI a bit too tightly?)

I also get the feeling that the points I've raised here are just trade-offs, for the time being, to get a proof-of-concept out the door; which is completely understandable and a good approach. As such, take what I say with a suitably large grain of salt 😅

topiary-cli/tests/samples/expected/ocamllex.mll
```diff
@@ -9,12 +9,13 @@
...
-      "and",    AND;
-      "or",     OR;
-      "not",    NOT ]
+  let keywords = [
+    "true", TRUE;
+    "false", FALSE;
+    "and", AND;
+    "or", OR;
+    "not", NOT
+  ]
```
aspiwack, line 18: Maybe you should add some multi-line code action in the input so that we can see formatting happening in code actions as well.

ErinvanderVeen:
> I assume I'm missing something obvious, but why not reuse the `str -> Language` mapping used for formatting the top-level (i.e., from the file extension)?

Conceptually the resolver is `str -> Option<Arc<Language>>`, but in the CLI it is not meant to be a hand-written match table for every injected language.

The split exists because `topiary-core` cannot know about CLI configuration, file discovery, query dirs, built-in query fallbacks, or caching. So core asks the caller: "given the injection query’s language name, can you provide the corresponding `Language`?" That is the resolver.

In this PR, the CLI resolver does reuse the normal config-based language loading path:

- `topiary-cli/src/main.rs:32` calls `cache.fetch_from_config(config, name)`.
- `topiary-cli/src/language.rs:89` resolves that name with `config.get_language(name)`, then loads the normal formatting query and optional injection query.

So the distinction is:

```text
top-level file:
  path -> extension detection -> configured language -> Language

injected span:
  injection query says "ocaml" -> configured language named "ocaml" -> Language
```

We cannot directly reuse file-extension detection for injected content because an injected span usually has no filename or extension. The host grammar/query has already identified "this node is OCaml", so the natural lookup key is the configured language name, not a path.

The only hard-coded-looking part is the built-in fallback for injection query contents in `topiary-cli/src/io.rs:397`: currently only `ocamllex` has a bundled `injections.scm`. But that is analogous to the existing bundled formatting-query fallback. The injected language mapping itself is config-name based, not hard-coded in the resolver.

ErinvanderVeen:
> Minor question: Also is the `Arc` just in there to conform to the interface of the language cache? Doesn't that tie core to the CLI a bit too tightly?

You are probably right here. Honestly, keeping the `Arc` around was done for convenience, I think the language resolver can probably pass `as_ref()` without any issue. I'll see if I can make that work.

ErinvanderVeen:
I tried dropping the Arc in core, but the only solution I could come up with it to make the resolver take a callback function. i.e. `Fn(&str, &mut dyn FnMut(&Language) -> FormatterResult<()>) -> FormatterResult<()>` OR

```
pub trait LanguageResolver {
    fn resolve(
        &self,
        name: &str,
        with_language: &mut dyn FnMut(&Language) -> FormatterResult<()>,
    ) -> FormatterResult<()>;
}
```

which feels worse to me.

Xophmeister:
Thanks, @ErinvanderVeen :pray:

@mkatychev and/or @prednaz: Could you guys please take a look at this, if you have time, so we can get it merged?

mkatychev:
@Xophmeister definitely will tomorrow!

topiary-core/src/tree_sitter.rs
```diff
@@ -105,6 +105,132 @@ impl TopiaryQuery {
...
+}
+
+/// A region of host source text that should be formatted as a different
+/// language, as determined by an [`InjectionQuery`].
+#[derive(Clone, Debug)]
+pub struct InjectionSpan {
+    /// Start byte (inclusive) within the host source.
+    pub start_byte: usize,
+    /// End byte (exclusive) within the host source.
+    pub end_byte: usize,
```
mkatychev: https://github.com/topiary/topiary/blob/7aab5c5b8bb005a1214680fa3763dea3eb32c457/topiary-core/src/error/error_span.rs#L38-L38

Should we use a `Range` like we do in `ErrorSpan`?

topiary-core/src/tree_sitter.rs
```diff
@@ -105,6 +105,132 @@ impl TopiaryQuery {
...
+/// `@injection.content` capture paired with the language declared by its
+/// pattern's `#injection_language!` predicate.
+///
+/// Patterns missing an `#injection_language!` predicate are skipped (with a
+/// warning logged).
+///
+/// # Errors
+///
+/// Returns an error only on internal tree-sitter failures; missing predicates
+/// or unmatched captures are logged, not raised.
```
mkatychev: is there somewhere that we can return an error in `collect_injections`? I was not able to find one 😅.

topiary-core/src/tree_sitter.rs
```diff
@@ -105,6 +105,132 @@ impl TopiaryQuery {
...
+        let Some(language_name) = language_name else {
+            log::warn!(
+                "Injection query pattern {} has no #injection_language! predicate; skipping",
+                query_match.pattern_index()
+            );
+            continue;
+        };
+
+        for capture in query_match.captures() {
+            if capture.name(capture_names.as_slice()) == "injection.content" {
```
mkatychev: would it be worth it to flatten the inner check with a filter?

```suggestion
        for capture in query_match.captures()
            .filter(|c| c.name(capture_names.as_slice()) == "injection.content") {
            // if capture.name(capture_names.as_slice()) == "injection.content" {
```

topiary-core/src/lib.rs
```diff
@@ -324,6 +384,82 @@ pub fn formatter_tree(
...
+) -> FormatterResult<()> {
+    for span in spans {
+        let inner_source = input_content
+            .get(span.start_byte..span.end_byte)
+            .ok_or_else(|| {
+                report!(FormatterError::Internal(format!(
+                    "Injected {} span is not on UTF-8 boundaries",
+                    span.language
+                )))
+            })?;
```
mkatychev: Could we use [`OptionExt`](https://docs.rs/rootcause/latest/rootcause/option_ext/trait.OptionExt.html) here to simplify this block a bit?
```suggestion
            .ok_or_report()
            .context(FormatterError::Internal("Injected span is not on UTF-8 boundaries")
            .attach_language(&span.language)?;
```

Though this well need to have `ErrorSpan` language changed to an `Option<String>`:
https://github.com/topiary/topiary/blob/ab16bfb1f70d93f03c6776df717a2f6b27ca631d/topiary-core/src/error/error_span.rs#L37-L37

topiary-core/src/tree_sitter.rs
```diff
@@ -105,6 +105,132 @@ impl TopiaryQuery {
...
+/// language, as determined by an [`InjectionQuery`].
+#[derive(Clone, Debug)]
+pub struct InjectionSpan {
+    /// Start byte (inclusive) within the host source.
+    pub start_byte: usize,
+    /// End byte (exclusive) within the host source.
+    pub end_byte: usize,
+    /// The injected language name, taken from the `#injection_language!`
+    /// predicate of the matching pattern.
+    pub language: String,
```
mkatychev, line 162: piggybacking off of https://github.com/topiary/topiary/pull/1244/changes#r3278160734, if we add an filepath field, we should be able to do an `impl From<InjectionSpan> for ErrorSpan` to simplify some of the error propagation for injection errors.


```suggestion
    pub language: String,
    /// The optional filepath from which the injection originated.
    pub filepath: Option<PathBuf>,
```

ErinvanderVeen: Thanks for the suggestion! 

I considered adding `filepath` to `InjectionSpan` to implement `impl From<InjectionSpan> for ErrorSpan`, but I think keeping it out is actually cleaner. Currently, core knows nothing about the fs.

Fortunately, the CLI already attaches the correct filepath at the boundary when it catches a library error:

```rust
process_fn(input, language, cache).map_err(|e| {
    if let TopiaryError::Lib(report) = e {
        return report.attach_filepath(location.to_path()).into();
    }
    e
})
```

topiary-core/src/error/error_span.rs
```diff
@@ -74,11 +74,11 @@ impl ErrorSpan {
         self
     }
 
-    pub fn set_language(&mut self, language: &'static str) {
-        self.language = Some(language);
+    pub fn set_language(&mut self, language: impl Into<String>) {
```
mkatychev: we may want to keep the "inner" function parameters type concrete for (compilation) performance reasons:
https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice

topiary-core/src/error/error_span.rs
```diff
@@ -74,11 +74,11 @@ impl ErrorSpan {
...
+        self.language = Some(language.into());
```
mkatychev: we may want to keep the "inner" function parameters type concrete for (compilation) performance reasons:
https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice
```suggestion
    pub fn set_language(&mut self, language: String) {
        self.language = Some(language);
```

moving by reference rather than by value (typically in nested calls) typically results in more compact assembly, this is more of a nit though:
```suggestion
    pub fn set_language(&mut self, language: &str) {
        self.language = Some(language.to_owned());
```

topiary-core/src/tree_sitter.rs
mkatychev: that's a great point

docs/book/src/guides/adding-a-new-language.md
```diff
@@ -193,6 +193,31 @@ fn to_query<T>(name: T) -> CLIResult<QuerySource>
...
+touch topiary-queries/queries/clang/injections.scm
+```
+
+An injection query captures host syntax that should be formatted by the
+inner language:
+
+```scheme
+(
+  (embedded_code) @injection.content
+  (#injection_language! "inner_language")
```
mkatychev, line 211: I take it we're not able to do dynamic injected language support yet similar to [this mockup](https://github.com/topiary/topiary/pull/1115/changes#diff-d12145f118d4bd60516cf7c279b7c8c8fe0d83f64010a427880f4f985c60c02f)?

docs/book/src/guides/adding-a-new-language.md
```diff
@@ -193,6 +193,31 @@ fn to_query<T>(name: T) -> CLIResult<QuerySource>
...
+)
+```
+
+The host language's `formatting.scm` still controls layout around the
+captured node. The captured content is formatted independently by the
```
mkatychev:
```suggestion
The host language's `formatting.scm` retains control of the layout around the
captured node. The captured content is formatted independently by the
```

topiary-core/benches/benchmark.rs
```diff
@@ -18,9 +18,10 @@ async fn format() {
 
     let language: Language = Language {
         name: "nickel".to_owned(),
-        query: TopiaryQuery::new(&nickel.into(), &query_content).unwrap(),
+        formatting_query: TopiaryQuery::new(&nickel.into(), &query_content).unwrap(),
         grammar: tree_sitter_nickel::LANGUAGE.into(),
         indent: None,
+        injection_query: None,
```
mkatychev, line 20: should we use the same form for both formatting and injecting?
* `format_query` & `injection_query`
* `formatting_query` & `injecting_query`

First one sounds better to me but 🤷

topiary-config/src/language.rs
```diff
@@ -120,6 +120,28 @@ impl Language {
...
+        let name = self.name.as_str();
+
+        #[rustfmt::skip]
+        let potentials: [Option<PathBuf>; 5] = [
+            std::env::var("TOPIARY_LANGUAGE_DIR").map(PathBuf::from).ok(),
+            option_env!("TOPIARY_LANGUAGE_DIR").map(PathBuf::from),
+            Source::fetch_one(&None).queries_dir(),
+            Some(PathBuf::from("./topiary-queries/queries")),
+            Some(PathBuf::from("../topiary-queries/queries")),
+        ];
```
mkatychev, line 136: should we create a function to return this so it can be reused in `find_injections_file` and `find_query_file`?

docs/book/src/guides/adding-a-new-language.md
ErinvanderVeen: Nope! I plan to do that in a followup PR!

topiary-core/src/lib.rs
```diff
@@ -274,22 +306,50 @@ pub fn formatter_str(
...
             tolerate_parsing_errors,
         } => {
+            log::debug!("Discovering potentially injected languages");
+            let spans = match &language.injection_query {
+                Some(injection_query) => collect_injections(&tree, input_content, injection_query),
+                None => Vec::new(),
+            };
+
+            // Create a list of nodes that are injection formatted.
+            // These must will be treated as leafs (although, in all likelyhood, they already are).
```
```suggestion
            // These must will be treated as leaves (although, in all likelihood, they already are).
```

topiary-core/src/lib.rs
```diff
@@ -274,22 +306,50 @@ pub fn formatter_str(
...
     match operation {
         Operation::Format {
             skip_idempotence,
             tolerate_parsing_errors,
         } => {
+            log::debug!("Discovering potentially injected languages");
+            let spans = match &language.injection_query {
+                Some(injection_query) => collect_injections(&tree, input_content, injection_query),
+                None => Vec::new(),
+            };
```
mkatychev, line 330:
```suggestion
            let spans = language
                .injection_query
                .as_ref()
                .map(|q| collect_injections(&tree, input_content, q))
                .unwrap_or_default();
```
Dunno if the functional approach reads better or not, feel free to disregard.

mkatychev (approved):
LGTM, just a coupe of nits

Xophmeister (approved):

topiary-core/src/tree_sitter.rs
```diff
@@ -105,6 +105,126 @@ impl TopiaryQuery {
...
+            .into_iter()
+            .find_map(|p| {
+                (&*p.operator() == "injection_language!")
+                    .then(|| p.args().into_iter().next())
+                    .flatten()
+            });
+
+        let Some(language_name) = language_name else {
+            log::warn!(
+                "Injection query pattern {} has no #injection_language! predicate; skipping",
```
prednaz, line 206: even a query without an `@injection.content` will trigger this warning.

Xophmeister:
@Niols With this change, we now have a functionally complete OCamllex formatter. Do you know whether this is new/unique in the OCaml ecosystem? @ErinvanderVeen has a blog post in preparation, so it would be good for us to highlight, if so. Thanks :pray:

Niols:
Pretty cool! To my knowledge, there is nothing. You should check whether ocp-indent has some support for that, but I don't think so, and I'm almost completely sure that ocamlformat has nothing.

That being said, if we're going to advertise this and hope to convince OCaml people to use Topiary, we should probably fix the few outstanding issues that make it unusable in practice for plain OCaml.

Niols:
(“Pretty cool” is an understatement, it really is an amazing feature!)

PR #1250, opened 2026-06-01
refactor!: remove TOPIARY_LANGUAGE_DIR and shift query_dir to Configur…

ErinvanderVeen:
> [!IMPORTANT]
> In draft and awaiting https://github.com/topiary/bud/discussions/9

# Remove `TOPIARY_LANGUAGE_DIR` and simplify query resolution architecture

## Description
Completely remove TOPIARY_LANGUAGE_DIR, and generally cleanup the query resolution strategy. I started implementing this realizing that it was kind of annoying to have topiary-config access environment variables. However, this quickly shifted into realizing that the TOPIARY_LANGUAGE_DIR doesn't really have a point anymore.

So, it has been removed. Additionally, the query resolution has been moved away from Language and into Config. This, IMHO makes more sense.

### Background
`TOPIARY_LANGUAGE_DIR` was introduced sometime 2022 as a workaround for Nix packaging issues to allow Topiary to find detached query files on disk. This is now obsolete because standard queries are compiled directly into the binary, and the Nickel `Configuration` system automatically infers local `queries/` folders dynamically.

### Changes
* **Removed side-effectful state:** Completely stripped `TOPIARY_LANGUAGE_DIR`r from the codebase.
* **Separation of Concerns:** Moved `query_dir` and `find_query_file()` away `Language` and onto `Configuration`. `Language` is now purely a data struct describing parsing details, and is fully agnostic to the fs.
* **New CLI argument:** Introduced a `--query-dir <PATH>` CLI argument to override the directory where queries are found.

## Checklist

- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docs/book/src/cli/usage/check-grammar.md
+++ b/docs/book/src/cli/usage/check-grammar.md
@@ -34,6 +34,9 @@ Options:
   -M, --merge-configuration
           Enable merging for configuration files
 
+      --query-dir <QUERY_DIR>
+          Override the directory where queries are found
+
   -v, --verbose...
           Logging verbosity (increased per occurrence)
 
--- a/topiary-config/src/lib.rs
+++ b/topiary-config/src/lib.rs
@@ -56,11 +56,7 @@ impl Configuration {
     /// If the configuration file exists, but cannot be parsed, this function will return a
     /// `TopiaryConfigError` with the error that occurred.
     #[allow(clippy::result_large_err)]
-    pub fn fetch(
-        merge: bool,
-        file: &Option<PathBuf>,
-        query_dir_override: Option<PathBuf>,
-    ) -> TopiaryConfigResult<(Self, NickelValue)> {
+    pub fn fetch(merge: bool, file: &Option<PathBuf>) -> TopiaryConfigResult<(Self, NickelValue)> {
         // If we have an explicit file, fail if it doesn't exist
         if let Some(path) = file
             && !path.exists()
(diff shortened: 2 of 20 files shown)


Xophmeister:
Thanks for looking into this, @ErinvanderVeen :yellow_heart: This has been one of those stubborn development artefacts that no one's had the energy to tackle.

There's an old issue that is related to this, which may no longer be relevant with your changes but still worth mentioning: #1003

docs/book/src/cli/usage/check-grammar.md
```diff
@@ -34,6 +34,9 @@ Options:
   -M, --merge-configuration
           Enable merging for configuration files
 
+      --query-dir <QUERY_DIR>
```
mkatychev, line 37: are we able to use `languages.ncl` to set a `query-dir` field?

Xophmeister (approved):
LGTM :tada:

docs/book/src/cli/usage/check-grammar.md
Xophmeister: My reading of the code is "Yes, but it's a global for all languages, rather than per-language". Something like this:
```nickel
{
  languages = {
    # ...
  },

  query_dir = "/path/to/my/queries"
}
```

However, I'm not a Nickel aficionado. YMMV :sweat_smile:

mkatychev: maybe there should be a global one for them all, but I would like to eventually  handle a `query.source` (for remote queries) similar to how grammar is done now:
https://github.com/topiary/topiary/blob/928df3c010da6e070dd11490bf7caae170be60e3/topiary-config/languages.ncl#L81-L86

Xophmeister: Oh, so `grammars/ocaml` in this example would be relative to the grammar repo root? That would be pretty cool :+1:

mkatychev: Reposting a discussion for visibility:
> I'm not sure a global `query_dir` is the right approach, maybe have it used as a fallback for `<lang>.query.source` but the way `grammar.source` works right now is easy to reason about, I'm worried a top level `query_dir` does not follow the same pattern and makes reconciling remote/built-in/local queries differently
> if we want to handle variants of queries for cases like this : https://github.com/topiary/topiary/blob/928df3c010da6e070dd11490bf7caae170be60e3/topiary-queries/queries/openscad/formatting.scm#L209-L222
> it will make it harder to be flexible if say there's an `openscad/queries/formatting_2021.scm` and a `openscad/queries/formatting.scm` locally or remote

topiary-config/src/lib.rs
```diff
@@ -304,11 +389,9 @@ impl From<SerdeConfiguration> for Configuration {
             .map(|(name, config)| Language::new(name, config))
             .collect();
 
-        Self { languages }
+        Self {
+            languages,
+            query_dir: value.query_dir,
```
mkatychev, line 394: I'm concerned this might create a scenario where there is competing sources of truth how to resolve a formatting query, if a `--query-dir` were to be passed in, it should be used as a resolution hint to handle creating individual resources in the same way grammar is populated on a per `LanguageConfiguration` basis:
https://github.com/topiary/topiary/blob/928df3c010da6e070dd11490bf7caae170be60e3/topiary-config/src/language.rs#L38-L50

Adding a `pub query: Query` to the struct above feels more consistent and allows us to expand/differentiate explicitly between query types (formatting vs. injection).
A top level `PathBuf` that stays in scope for the duration of `Configuration` feels odd.

mkatychev: A top level `cache_dir` override would be great though so we can cache both `topariy/<language>/queries` and `topariy/<language>/grammars` in the same place and expose that through a config file.

https://github.com/topiary/topiary/blob/928df3c010da6e070dd11490bf7caae170be60e3/topiary-config/src/language.rs#L148-L163

PR #1255, opened 2026-06-04
Unpin `wasm-bindgen`

### LINKED ISSUE #1128: `wasm-bindgen` dependency is hard-pinned

Xophmeister:
# Bump `nickel-lang-core` to unpin `wasm-bindgen`

Resolves #1128

## Description

`wasm-bindgen` was a pinned dependency, which was causing problems for downstream users. Now that Nickel has unpinned its dependency on this crate, Topiary can do the same.

> [!NOTE]
> A new release is planned in the coming weeks that will include this PR, but this PR itself will not be in a more-imminent patch release.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,10 +66,9 @@ tree-sitter-json = "0.24"
 tree-sitter-language = "0.1"
 tree-sitter-nickel = "0.5"
 unescape = "0.1"
-wasm-bindgen = "0.2.100"
-wasm-bindgen-futures = "0.4"
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+wasm-bindgen-test = "0.3"
 web-sys = "0.3"
 
 # NOTE Keep these versions aligned
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,8 @@ tree-sitter-nickel = "0.5"
 unescape = "0.1"
 wasm-bindgen = "0.2.100"
 wasm-bindgen-futures = "0.4"
-wasm-bindgen-test = "0.3"
+wasm-bindgen = "0.2"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -66,9 +66,9 @@ tree-sitter-json = "0.24"
 tree-sitter-language = "0.1"
 tree-sitter-nickel = "0.5"
 unescape = "0.1"
-wasm-bindgen = "=0.2.100"
-wasm-bindgen-futures = "=0.4"
-wasm-bindgen-test = "=0.3"
+wasm-bindgen = "0.2.100"
+wasm-bindgen-futures = "0.4"
+wasm-bindgen-test = "0.3"
```
mkatychev, line 71: do we want to potentially relax the `wasm-bindgen` version to match the semver scope of the the other dependencies?
```suggestion
wasm-bindgen = "0.2"
wasm-bindgen-futures = "0.4"
```

mkatychev (approved):
LGTM

PR #1257, opened 2026-06-09
slice instead of indexes in `InjectionSpan`

prednaz:
References #1244

## Description

tiny suggestion for #1244. #1244 is merged already. so i am giving this suggestion its own pull request.

the suggestion is to pass around a string slice as opposed to an indexes in order to make the error of indexing a different or mutated string impossible. and in order to make the impossibility of the error of indexing off utf-8 boundaries so locally self evident that we can just panic.

under the hood, slices are basically just indexes too plus a reference to the source string. the indexes can even be recovered from a slice via [`std::slice::element_offset`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.element_offset) if that was ever needed.

### DIFF
--- a/topiary-core/src/tree_sitter.rs
+++ b/topiary-core/src/tree_sitter.rs
@@ -155,8 +155,8 @@ impl InjectionQuery {
 /// A region of host source text that should be formatted as a different
 /// language, as determined by an [`InjectionQuery`].
 #[derive(Clone, Debug)]
-pub struct InjectionSpan {
-    pub byte_range: std::ops::Range<usize>,
+pub struct InjectionSpan<'a> {
+    pub byte_range: &'a str,
     /// The injected language name, taken from the `#injection_language!`
     /// predicate of the matching pattern.
     pub language: String,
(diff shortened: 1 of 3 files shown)


mkatychev:
TIL about `std::slice::element_offset`!

topiary-core/src/tree_sitter.rs
```diff
@@ -215,7 +215,9 @@ pub fn collect_injections(
         {
             let node = capture.node();
             spans.push(InjectionSpan {
-                byte_range: node.start_byte() as usize..node.end_byte() as usize,
+                byte_range: input_content
+                    .get(node.start_byte() as usize..node.end_byte() as usize)
```
mkatychev: We should be able to use [`tree_sitter::Node::byte_range`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.byte_range) here, though we're using something else in our wrapper:
https://github.com/topiary/topiary/blob/382261d09a584a2ed8cae761579e2b716a949e0e/topiary-tree-sitter-facade/src/node.rs#L304-L308

mkatychev (commented):
LGTM overall, just one minor nit

topiary-core/src/tree_sitter.rs
prednaz: i have attempted to implement this in [`112b9e4`](https://github.com/topiary/topiary/pull/1257/changes/112b9e44e350101ec6ca166e1f5292098b0e6272). but i do not know how to check the `wasm32` part.

Xophmeister: > but i do not know how to check the `wasm32` part.

AFAIK, there are no exposed checks for wasm: `nix flake check` will check on native; `nix build .#client-app` will build the wasm-toolchain, but it won't run any checks.

Tbh, the wasm build isn't currently used, so as long as it builds, that should be enough.

Xophmeister (approved):

mkatychev (approved):
LGTM!

PR #1266, opened 2026-06-18
multi line strings where the closing column is insignificant

### LINKED ISSUE #654: Nickel multiline strings aren't formatted
**Describe the bug**

Nickel [multiline strings](https://nickel-lang.org/user-manual/syntax#strings) are mostly ignored (and left in-place) by topiary.
More precisely, it seems that the opening delimiter is taken into account, but the content of the string and the closing delimiter aren't touched.

**To Reproduce**

Format a file like:

```nickel
  m%"
      foo
         x
        "%
```

The opening `m%"` will be moved at the begining of the line (good), but the rest will be kept with its weird indentation.

**Expected behavior**

I don't have a very strong opinion on how this is formatted, but there should be some consistent formatting applied. What I use and tend to see in the wild is something similar to most delimited block, namely:

- A multi-line string in a single line (`m%"foobar"%`) is kept as it is
- A “real” multi-line string is formatted such that
  - The opening token is at the end of the previous block
  - The content of the string is indented by two spaces compared to the previous block
  - The closing delimiter is indented as the previous block

(hope that makes sense).

So the above should be formatted as:

```
m%"
  foo
     x
"%
```

Some examples of multi-line strings in the wild:
- https://github.com/tweag/nickel/blob/8d7725da772599a65874a827aa02feb8306852ed/examples/config-gcc/config-gcc.ncl#L68-L71
- https://nickel-lang.org/user-manual/syntax#strings
- https://github.com/tweag/tf-ncl/blob/f1b6c38091936fc4533c2228cbf8a240cfef8997/.github/workflows/ci.ncl#L31-L34
- https://github.com/tweag/tf-ncl/blob/f1b6c38091936fc4533c2228cbf8a240cfef8997/examples/hello-tf/main.ncl#L6-L8 (which I'd consider ill-formatted)

**Environment**
 - OS name + version:
 - Version of the code: topiary-cli 0.3.0 (79b93527d9bd59533f9a79fe490567963193fafd)

**Additional context**

- The semantic of multi-line strings mean that the indentation of their content is the indentation of the less-indented line in them. Hopefully that's possible to express with topiary :crossed_fingers: 
- This also applies to [symbolic strings](https://nickel-lang.org/user-manual/syntax#symbolic-strings) which are essentially equivalent in that regard

### LINKED ISSUE #1224: `@multi_line_indent_all` inconsistently takes into account extra spaces across its length it should not.
multi_line_indent_all takes into account any spaces prepended to that particular element into account across its whole height

So if you have

```nix
{
  abc = ''
    abc
  '';
  abcd = /* comment */ ''
    abcd
  '';
}
```

and you do append_space on the indented_string_expression

You get

```nix
{
  abc = ''
     abc
   ''; # <- 1 more space on all the lines
  abcd = /* comment */ ''
     abcd
   '';
}
```

But it doesn't take into account the total indentation in the source code. And likewise, if I put `append_space` on the `=` instead, it also does not lead to this effect.

It should take into account the scope, but it should not be taking the append_space across its whole length

The reason I believe the above, where spaces apply to subsequent leaf nodes with multi_line_indent_all on them, is a bug, is because it is inconsistent.

```scm
(binding
  attrpath: (attrpath) @append_space
  "=" @append_space ; Adding it to the = somehow doesn't add the space to the multi line indent
)
```
```scm
(binding
  attrpath: (attrpath) @append_space
  (_)* @prepend_space ; <- This adds to the multi_line_indent_all (just 1 space, regardless of if there was a /* comment */ or not after the = before the value which starts on that line)
)
```
```scm
(binding
  attrpath: (attrpath) @append_space
  "=" @append_space ; Adding it to the = somehow doesn't add the space to the multi line indent
  (comment)? @append_space ; <- somehow this gets added to multi_line_indent_all still but only if there was a /* comment */ present after the =?
)
```

https://github.com/BirdeeHub/birdeevim/blob/cf8a5522ceadd024c60389aee0f45c7369d8e5fc/nix/topiary/queries/nix.scm#L126-L131

That behavior is very strange.

I am not understanding why the first does not apply the extra space and the 3rd does apply the extra space.

I understand why the second is adding the space, I dont agree that it should but I understand. But given number 3 adds a space, why isnt it adding 2 spaces across the length of the multi_line_indent_all?

Personally, I feel it should be only taking indentation scope into account from actual append or prepend indent captures.

Further context of how I got to this discovery https://github.com/topiary/topiary/discussions/1221 (there may be another issue to come of that discussion once I pin it down further as well, but I am unsure as of yet)

related issues

https://github.com/topiary/topiary/issues/998

and maybe also https://github.com/topiary/topiary/issues/172

### LINKED ISSUE #1225: `@multi_line_indent_all` inconsistently takes into account extra spaces
https://github.com/topiary/topiary/issues/1224

I hastily closed it in error

Here is the minimal reproduction

https://github.com/topiary/topiary/issues/1224#issuecomment-4324513108

Inconsistent behavior with multi_line_indent_all across multiple lines when the element or one before it on the line has had append_space applied to it or not.

It should only be listening to scope from actual indent requests, not random additional spaces as well, and especially not only if those additional spaces are actually applied or not.

prednaz:
Resolves #654
Resolves #1224
Resolves #1225
References #1236 (*2nd strategy*)

## Description

- new directive `@mutli_line_string`
- use the new directive in `nickel/formatting.scm`

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ This name should be decided amongst the team before the release.
 - [#1312](https://github.com/topiary/topiary/pull/1312) Added `bin/changelog.nu` to automatically add entry to `CHANGELOG.md`
 - [#1315](https://github.com/topiary/topiary/pull/1315) Add rootcause-backtrace to cli
 - [#1314](https://github.com/topiary/topiary/pull/1314) Allow skipping of formatting stages
+- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string.
 
 ### Removed
 - [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,7 @@ This name should be decided amongst the team before the release.
 - [#1299](https://github.com/topiary/topiary/pull/1299) Add `QuerySource` config fetching to topiary-config.
 - [#1312](https://github.com/topiary/topiary/pull/1312) Added `bin/changelog.nu` to automatically add entry to `CHANGELOG.md`
 - [#1315](https://github.com/topiary/topiary/pull/1315) Add rootcause-backtrace to cli
-- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string.
+- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string. And use `@multi_line_string` to add multi line string formatting to the Nickel formatter.
 
 ### Removed
 - [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ This name should be decided amongst the team before the release.
 - [#1312](https://github.com/topiary/topiary/pull/1312) Added `bin/changelog.nu` to automatically add entry to `CHANGELOG.md`
 - [#1315](https://github.com/topiary/topiary/pull/1315) Add rootcause-backtrace to cli
 - [#1314](https://github.com/topiary/topiary/pull/1314) Allow skipping of formatting stages
-- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string.
+- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string. And use `@multi_line_string` to add multi line string formatting to the Nickel formatter.
 
 ### Removed
 - [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ This name should be decided amongst the team before the release.
 - [#1299](https://github.com/topiary/topiary/pull/1299) Add `QuerySource` config fetching to topiary-config.
 - [#1312](https://github.com/topiary/topiary/pull/1312) Added `bin/changelog.nu` to automatically add entry to `CHANGELOG.md`
 - [#1315](https://github.com/topiary/topiary/pull/1315) Add rootcause-backtrace to cli
+- [#1266](https://github.com/topiary/topiary/pull/1266) Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string.
 
 ### Removed
 - [#1217](https://github.com/topiary/topiary/pull/1217) The `check` alias for the `check-grammar` subcommand, to avoid confusion with `topiary fmt --check`.
--- a/docs/book/src/reference/capture-names/indentation.md
+++ b/docs/book/src/reference/capture-names/indentation.md
@@ -53,6 +53,38 @@ respectively, after) them.
 ] @append_indent_end
 ```
 
+## `@multi_line_string`
+
+To be used on multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of the multi line string. Consider the following Nickel example.
+```Nickel
+m%"a
+b"%
+```
+
+`@multi_line_string` could format this as follows.
+```Nickel
+m%"
+  a
+  b
+"%
+```
+
+Nickel considers these 2 strings above as equal because Nickel ignores indentation common to all lines of the multi line string. Otherwise, `@multi_line_string` would not be admissible in a Nickel formatter.
+
+As illustrated in the example below, `@multi_line_string` requires both a `@multi_line_string.start` and a `@multi_line_string.end` directive in the same query matching the start and end delimiters of the string, for example `m%"` and `"%` in Nickel.
+
+As illustrated in the example below, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last nonwhitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
+
+### Example
+
+```scheme
+(str_chunks_multi
+  start: (multstr_start) @multi_line_string.start
+  end: (multstr_end) @multi_line_string.end
+  (#multi_line_string.last_insignificant!)
+) @multi_line_string
+```
+
 ## `@multi_line_indent_all`
 
 To be used on comments, or other leaf nodes, to indicate that we should
--- a/docs/book/src/reference/capture-names/indentation.md
+++ b/docs/book/src/reference/capture-names/indentation.md
@@ -73,7 +73,7 @@ Nickel considers these 2 strings above as equal because Nickel ignores indentati
 
 As illustrated in the example below, `@multi_line_string` requires both a `@multi_line_string.start` and a `@multi_line_string.end` directive in the same query matching the start and end delimiters of the string, for example `m%"` and `"%` in Nickel.
 
-As illustrated in the example below, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last nonwhitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
+As illustrated in the example below, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last non-whitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
 
 ### Example
 
--- a/docs/book/src/reference/capture-names/indentation.md
+++ b/docs/book/src/reference/capture-names/indentation.md
@@ -53,6 +53,61 @@ respectively, after) them.
 ] @append_indent_end
 ```
 
+## `@multi_line_string`
+
+To be used on multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of the multi line string. Consider the following Nickel example.
+```Nickel
+m%"a
+b"%
+```
+
+`@multi_line_string` could format this as follows.
+```Nickel
+m%"
+  a
+  b
+"%
+```
+
+Nickel considers these 2 strings above as equal because Nickel ignores indentation common to all lines of the multi line string. Otherwise, `@multi_line_string` would not be admissible in a Nickel formatter.
+
+Here is an example query for Nickel.
+```scheme
+(str_chunks_multi
+  start: (multstr_start) @multi_line_string.start
+  end: (multstr_end) @multi_line_string.end
+  (#multi_line_string.last_insignificant!)
+) @multi_line_string
+```
+
+As illustrated by this example, `@multi_line_string` requires both a `@multi_line_string.start` and a `@multi_line_string.end` directive in the same query matching the start and end delimiters of the string, for example `m%"` and `"%` in Nickel.
+
+As furthermore illustrated by this example above, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last non-whitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
+
+Here is an example query for Nix.
+```scheme
+(indented_string_expression
+  "''" @multi_line_string.start
+  "''" @multi_line_string.end
+  (#multi_line_string.carriage_return_significant!)
+  (#multi_line_string.tab_significant!)
+) @multi_line_string
+```
+
+As illustrated by this example, you can add an optional `#multi_line_string.carriage_return_significant!` predicate. It indicates that the formatted language made the questionable design decision that carriage returns always become part of the string's value. even if they occur at the very line end in a CRLF line ending.
+
+As furthermore illustrated by this example above, you can add an optional `#multi_line_string.tab_significant!` predicate. It indicates that the formatted language made the questionable design decision that tabs always become part of the string's value. even if they occur in indentation common to all lines of the multi line string.
+
+### Example
+
+```scheme
+(str_chunks_multi
+  start: (multstr_start) @multi_line_string.start
+  end: (multstr_end) @multi_line_string.end
+  (#multi_line_string.last_insignificant!)
+) @multi_line_string
+```
+
 ## `@multi_line_indent_all`
 
 To be used on comments, or other leaf nodes, to indicate that we should
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -9,8 +9,8 @@ use rootcause::prelude::ResultExt;
 use topiary_tree_sitter_facade::Node;
 
 use crate::{
-    Atom, Capitalisation, FormatterError, FormatterResult, ScopeCondition, ScopeInformation,
-    tree_sitter::NodeExt,
+    Atom, Capitalisation, EnforceIndentation, FormatterError, FormatterResult, MultiLineIndent,
+    ScopeCondition, ScopeInformation, tree_sitter::NodeExt,
 };
 
 /// A struct that holds sets of node IDs that have line breaks before or after them.
@@ -30,7 +30,7 @@ struct NodesWithLinebreaks {
 /// repeating the leaf-id search loop.
 struct LeafFlagsMut<'a> {
     single_line_no_indent: &'a mut bool,
-    multi_line_indent_all: &'a mut bool,
+    multi_line_indent_all: &'a mut MultiLineIndent,
     keep_whitespace: &'a mut bool,
 }
 
@@ -233,6 +233,16 @@ impl AtomCollection {
             predicates.scope_id.as_deref().ok_or_else(|| {
                 FormatterError::Query(format!("@{name} requires a #scope_id! predicate"))
             })
+        };
+        let requires_multi_line_string_delimiters = || {
+            predicates
+                .multi_line_string_delimiters
+                .as_ref()
+                .ok_or_else(|| {
+                    FormatterError::Query(format!(
+                        "@{name} requires both a @multi_line_string.start and a @multi_line_string.end directive in the same query"
+                    ))
+                })
         };
 
         // For the {prepend/append}_scope_{begin/end} captures we need this information,
@@ -480,10 +490,22 @@ impl AtomCollection {
                 });
                 self.append(Atom::Hardline, node, predicates);
             }
+            "multi_line_string" => {
+                let (start, end) = requires_multi_line_string_delimiters()?;
+                self.mutate_leaf_flags(node.id(), |flags| {
+                    *flags.multi_line_indent_all =
+                        MultiLineIndent::EnforceIndentation(EnforceIndentation {
+                            last_line_break_significant: !predicates
+                                .multi_line_string_last_insignificant,
+                            start: start.clone(),
+                            end: end.clone(),
+                        });
+                });
+            }
             // Mark a leaf to have all its lines be indented
             "multi_line_indent_all" => {
                 self.mutate_leaf_flags(node.id(), |flags| {
-                    *flags.multi_line_indent_all = true;
+                    *flags.multi_line_indent_all = MultiLineIndent::MaintainOffset;
                 });
             }
             // Mark a leaf to disable trimming
@@ -609,7 +631,7 @@ impl AtomCollection {
                 id,
                 original_position: node.start_position().into(),
                 single_line_no_indent: false,
-                multi_line_indent_all: false,
+                multi_line_indent_all: MultiLineIndent::None,
                 keep_whitespace: false,
                 capitalisation: Capitalisation::Pass,
             });
@@ -1159,6 +1181,10 @@ pub struct QueryPredicates {
     pub multi_line_scope_only: Option<String>,
     /// A query name, for debugging/logging purposes
     pub query_name: Option<String>,
+    /// multi line string delimiters
+    pub multi_line_string_delimiters: Option<(String, String)>,
+    /// The flag that indicates that topiary must not add any line breaks to the end of multi line strings
+    pub multi_line_string_last_insignificant: bool,
 }
 
 /// Collapses spaces before antispace atoms in a vector of atoms.
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -234,15 +234,19 @@ impl AtomCollection {
                 FormatterError::Query(format!("@{name} requires a #scope_id! predicate"))
             })
         };
-        let requires_multi_line_string_delimiters = || {
-            predicates
-                .multi_line_string_delimiters
-                .as_ref()
-                .ok_or_else(|| {
+        let requires_multi_line_string_delimiters = || -> FormatterResult<(&String, &String)> {
+            Ok((
+                predicates.multi_line_string_start.as_ref().ok_or_else(|| {
                     FormatterError::Query(format!(
-                        "@{name} requires both a @multi_line_string.start and a @multi_line_string.end directive in the same query"
+                        "@{name} requires a @multi_line_string.start capture in the same query"
                     ))
-                })
+                })?,
+                predicates.multi_line_string_end.as_ref().ok_or_else(|| {
+                    FormatterError::Query(format!(
+                        "@{name} requires a @multi_line_string.end capture in the same query"
+                    ))
+                })?,
+            ))
         };
 
         // For the {prepend/append}_scope_{begin/end} captures we need this information,
@@ -1181,8 +1185,10 @@ pub struct QueryPredicates {
     pub multi_line_scope_only: Option<String>,
     /// A query name, for debugging/logging purposes
     pub query_name: Option<String>,
-    /// multi line string delimiters
-    pub multi_line_string_delimiters: Option<(String, String)>,
+    /// multi line string start delimiter
+    pub multi_line_string_start: Option<String>,
+    /// multi line string end delimiter
+    pub multi_line_string_end: Option<String>,
     /// The flag that indicates that topiary must not add any line breaks to the end of multi line strings
     pub multi_line_string_last_insignificant: bool,
 }
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -30,7 +30,7 @@ struct NodesWithLinebreaks {
 /// repeating the leaf-id search loop.
 struct LeafFlagsMut<'a> {
     single_line_no_indent: &'a mut bool,
-    multi_line_indent_all: &'a mut MultiLineIndent,
+    multi_line_indent: &'a mut MultiLineIndent,
     keep_whitespace: &'a mut bool,
 }
 
@@ -160,7 +160,7 @@ impl AtomCollection {
             if let Atom::Leaf {
                 id,
                 single_line_no_indent,
-                multi_line_indent_all,
+                multi_line_indent,
                 keep_whitespace,
                 ..
             } = atom
@@ -168,7 +168,7 @@ impl AtomCollection {
             {
                 f(LeafFlagsMut {
                     single_line_no_indent,
-                    multi_line_indent_all,
+                    multi_line_indent,
                     keep_whitespace,
                 });
             }
@@ -493,7 +493,7 @@ impl AtomCollection {
             "multi_line_string" => {
                 let (start, end) = requires_multi_line_string_delimiters()?;
                 self.mutate_leaf_flags(node.id(), |flags| {
-                    *flags.multi_line_indent_all =
+                    *flags.multi_line_indent =
                         MultiLineIndent::EnforceIndentation(EnforceIndentation {
                             last_line_break_significant: !predicates
                                 .multi_line_string_last_insignificant,
@@ -505,7 +505,7 @@ impl AtomCollection {
             // Mark a leaf to have all its lines be indented
             "multi_line_indent_all" => {
                 self.mutate_leaf_flags(node.id(), |flags| {
-                    *flags.multi_line_indent_all = MultiLineIndent::MaintainOffset;
+                    *flags.multi_line_indent = MultiLineIndent::MaintainOffset;
                 });
             }
             // Mark a leaf to disable trimming
@@ -631,7 +631,7 @@ impl AtomCollection {
                 id,
                 original_position: node.start_position().into(),
                 single_line_no_indent: false,
-                multi_line_indent_all: MultiLineIndent::None,
+                multi_line_indent: MultiLineIndent::None,
                 keep_whitespace: false,
                 capitalisation: Capitalisation::Pass,
             });
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -9,8 +9,8 @@ use rootcause::prelude::ResultExt;
 use topiary_tree_sitter_facade::Node;
 
 use crate::{
-    Atom, Capitalisation, EnforceIndentation, FormatterError, FormatterResult, MultiLineIndent,
-    ScopeCondition, ScopeInformation, tree_sitter::NodeExt,
+    Atom, Capitalisation, FormatterError, FormatterResult, MultiLineIndent, ScopeCondition,
+    ScopeInformation, multi_line_indent, tree_sitter::NodeExt,
 };
 
 /// A struct that holds sets of node IDs that have line breaks before or after them.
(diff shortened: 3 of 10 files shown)


prednaz:
making tree-sitter grammars have dedicated symbols for string contents might be a problem because tree-sitter does not seem to support symbols that match the empty string. https://github.com/prednaz/tree-sitter-nickel/commit/2f41057870731b188087b0b62029e3ce7fe893ce yields the following error.
```
[nix develop:~/tweag/tree-sitter-nickel]$ tree-sitter generate 
Warning: rule infix_u_op_5 contains a `choice` rule with a single element. This is unnecessary.
Warning: rule infix_lazy_b_op_9 contains a `choice` rule with a single element. This is unnecessary.
Warning: rule infix_lazy_b_op_10 contains a `choice` rule with a single element. This is unnecessary.
Error: Error when generating parser

Caused by:
    The rule `str_chunks_multi_content` matches the empty string.
    
    Tree-sitter does not support syntactic rules that match the empty string
    unless they are used only as the grammar's start rule.
```

docs/book/src/reference/capture-names/indentation.md
```diff
@@ -53,6 +53,10 @@ respectively, after) them.
 ] @append_indent_end
 ```
 
+## `@multi_line_string`
+
+to do
```
prednaz: to do. document

topiary-core/src/lib.rs
```diff
@@ -41,13 +41,27 @@ pub struct ScopeInformation {
...
     LowerCase,
-    #[default]
     Pass,
 }
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AbsoluteIndentation {
+    ClosingColumnInsignificant { last_line_break_significant: bool, start: String, end: String},
+    ClosingColumnSignificant,
+    Comment,
```
prednaz: this pull request does not implemented these yet. shall i leave them out?

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1480 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+    indent_level: usize,
+    indent: &str,
+) -> String {
+    let AbsoluteIndentation::ClosingColumnInsignificant {
+        last_line_break_significant,
+        start,
+        end,
+    } = absolute_indentation
+    else {
+        todo!()
```
prednaz: see https://github.com/topiary/topiary/pull/1266#discussion_r3583880037

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1480 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+    };
+
+    let content: Vec<&str> = content_original
+        .strip_prefix(start)
+        .expect("to do")
```
prednaz: to do. error handling for the case of the string not starting with the declared delimiters

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1480 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        .strip_suffix(end)
+        .expect("to do")
```
prednaz: to do. error handling for the case of the string not ending with the declared delimiters

topiary-core/src/tree_sitter.rs
```diff
@@ -503,12 +512,32 @@ pub(crate) fn apply_query_tree_with_forced_leaves(
...
+                .rev()
+                .next()
+                .map(capture_content),
+            m.captures
+                .iter()
+                .filter(|c| c.name(&capture_names).deref() == "multi_line_string_end")
+                .rev()
+                .next()
+                .map(capture_content),
+        );
```
prednaz: this has got to be controversial. i am storing directives in `predicates`.

in my defense, we are already not storing predicates in there but directives according to [the tree-sitter documentation's](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/3-predicates-and-directives.html) terminology. haha

furthermore, `check_predicates` represents some naughty [*shotgun parsing*](https://stackoverflow.com/a/50852772). i wonder if it would be sensible to [parse](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate) the pair of `query.query.general_predicates(m.pattern_index)` and `m.captures` into 1 large `enum` with a constructor with appropriate fields for every topiary directive ("captures" in tree-sitter terminology) with its predicates ("directives" in tree-sitter terminology) such that this `enum` encodes and enforces which predicates are allowed and which required for which topiary directive.

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1480 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        .split("\n")
+        .map(|s| s.strip_suffix("\r").unwrap_or(s)) // to do. remove this?
```
prednaz: to do. remove this?

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1483 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        // purely for warning generation
+        match common_whitespace_prefix.clone().count().cmp(
+            &whitespace_prefixes.map(Iterator::count).min().expect(
+                "whitespace_prefixes should not be empty if `common_prefix` returns `Some`.",
+            ),
+        ) {
+            Ordering::Less => println!(
+                // to do
+                "is this indentation? then you should not mix different kinds of whitespace characters."
+            ),
```
prednaz: to do. proper warning

prednaz:
do not worry. those 3000 lines are almost all tests.

topiary-core/src/atom_collection.rs
```diff
@@ -480,10 +490,30 @@ impl AtomCollection {
...
+                        let (start, end) = requires_multi_line_string_delimiters()?.clone();
+                        *multi_line_indent_all = MultiLineIndent::AbsoluteIndentation(
+                            AbsoluteIndentation::ClosingColumnInsignificant {
+                                last_line_break_significant: predicates.last_line_break_significant,
+                                start,
+                                end,
+                            },
+                        );
+                    }
+                }
```
prednaz: to do. translate to `mutate_leaf_flags`

docs/book/src/reference/capture-names/indentation.md
```diff
@@ -53,6 +53,38 @@ respectively, after) them.
...
+  a
+  b
+"%
+```
+
+Nickel considers these 2 strings above as equal because Nickel ignores indentation common to all lines of the multi line string. Otherwise, `@multi_line_string` would not be admissible in a Nickel formatter.
+
+As illustrated in the example below, `@multi_line_string` requires both a `@multi_line_string.start` and a `@multi_line_string.end` directive in the same query matching the start and end delimiters of the string, for example `m%"` and `"%` in Nickel.
+
+As illustrated in the example below, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last nonwhitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
```
prednaz: i considered elaborating on the kind of formatting that `@multi_line_string` performs but that would require more examples below and then these references to "the example below" would not work anymore. i already find these forward references unfortunate enough as it is. ideally, i would want the reader to read the example first. but this *capture names* section of the book has this very consistent structure of having the usage example subsection at the very end of each section. is it worth breaking this structure?

topiary-core/src/lib.rs
prednaz: discussed in a meeting and the answer is "yes".

topiary-core/src/lib.rs
```diff
@@ -41,13 +41,27 @@ pub struct ScopeInformation {
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EnforceIndentation {
+    pub last_line_break_significant: bool,
+    pub start: String,
+    pub end: String,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum MultiLineIndent {
```
```suggestion
pub enum MultilineIndent {
```
I think multiline is a proper adjective in english:
https://www.merriam-webster.com/dictionary/multiline

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1619 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        assert_eq!(
+            render_enforced_indentation(
+                &EnforceIndentation {
+                    last_line_break_significant: true,
+                    start: "''".to_owned(),
+                    end: "''".to_owned(),
+                },
+                "''    
+    
+''",
```
mkatychev: @prednaz have you thought about having placeholder chars for visualization like [dot symbol](https://symbolsdb.com/dot-symbol) that you'd do a replace on?
```suggestion
                "''····
········
''".replace('·', " "),
```

mkatychev: It might be clearer to also have the start of the multiline string on the 1st column:
```suggestion
                "\
''    
    
''",
```

topiary-core/src/pretty.rs
mkatychev: I've gotten in the habit of adding [tagged TODOs](https://github.com/bytecodealliance/wasm-pkg-tools/blob/dfa2e17128d1b938019d86711f16fb033fddb4f4/crates/wasm-pkg-core/src/manifest/paths.rs#L84-L84) to the projects I collaborate on to make it easier to grep/filter the TODOs meant for myself:
```suggestion
            Ordering::Less => println!(
                // TODO(prednaz): proper warning
                "is this indentation? then you should not mix different kinds of whitespace characters."
            ),
```

topiary-core/src/lib.rs
prednaz: is this worth breaking consistency with [`@multi_line_indent_all`](https://topiary.tweag.io/book/reference/capture-names/indentation.html#multi_line_indent_all) and [`ScopeCondition::MultiLineOnly`](https://github.com/topiary/topiary/blob/main/topiary-core/src/lib.rs#L164) for?

topiary-core/src/pretty.rs
prednaz: implemented in [`6bbe041`](https://github.com/topiary/topiary/pull/1266/changes/6bbe04171ea245c45fb805edf7fb80996d92a8a2)

topiary-core/src/pretty.rs
Xophmeister: This is a nice idea. I think I might steal it :grin:

docs/book/src/reference/capture-names/indentation.md
```diff
@@ -53,6 +53,38 @@ respectively, after) them.
...
+```Nickel
+m%"
+  a
+  b
+"%
+```
+
+Nickel considers these 2 strings above as equal because Nickel ignores indentation common to all lines of the multi line string. Otherwise, `@multi_line_string` would not be admissible in a Nickel formatter.
+
+As illustrated in the example below, `@multi_line_string` requires both a `@multi_line_string.start` and a `@multi_line_string.end` directive in the same query matching the start and end delimiters of the string, for example `m%"` and `"%` in Nickel.
```
Xophmeister: In Nickel, the string termini are matched by something like `m(%+)"` and `"\1` (e.g., `m%%%"string"%%%` is valid). Does you implementation allow for that, or is it just string match for now?

If it's just string match, that's fine -- it shouldn't be a blocker -- but it's worth documenting a future enhancement in an issue :+1:

docs/book/src/reference/capture-names/indentation.md
Xophmeister: Nitpick
```suggestion
As illustrated in the example below, you can add an optional `#multi_line_string.last_insignificant!` predicate. It indicates that a line break can be introduced before the end delimiter if, in the input, the end delimiter is on the same line as the string's last non-whitespace line, as illustrated by `b"%` in the examples above. Nickel considers this not to change the string's value. Otherwise, `#multi_line_string.last_insignificant!` would not be admissible in a Nickel formatter. For example, Nix does consider this to change the string's value.
```

topiary-core/src/pretty.rs
Xophmeister: Is this for Windows line-ending compatibility? That sounds like a good idea, but as far as I remember, we just assume Unix line-endings everywhere...which is probably not ideal!!

Either way, this todo comment should be removed, if you intend to keep this code.

CHANGELOG.md
Xophmeister, line 1: What happened here 😅 I think we want to keep the `[#XXX](https://github.com/topiary/topiary/pull/XXX)` links, because they're a bit nicer to look at than raw URLs.

CHANGELOG.md
```diff
@@ -45,38 +45,39 @@ This name should be decided amongst the team before the release.
...
-- [#1254](https://github.com/topiary/topiary/pull/1254) Add Markdown formatting support with dynamic language injections for fenced code blocks (via the `@injection.language` capture).
+- https://github.com/topiary/topiary/pull/1200 Build and deploy Topiary Docker images to ghcr.io.
+- https://github.com/topiary/topiary/pull/1217 Add `--check` flag to `topiary fmt` for CI formatting verification.
+- https://github.com/topiary/topiary/pull/1227 Add `@append_empty_input_softline` and `@prepend_empty_input_softline` captures, thanks to @BirdeeHub
+- https://github.com/topiary/topiary/pull/1244 Add language injection support for injection languages known at query writing.
+- https://github.com/topiary/topiary/pull/1254 Add Markdown formatting support with dynamic language injections for fenced code blocks (via the `@injection.language` capture).
+- https://github.com/topiary/topiary/pull/1266 Add `@multi_line_string` capture to format multi line strings in languages like Nix and Nickel that ignore indentation common to all lines of a multi line string.
```
Xophmeister: I might be good to indicate in the changelog line that this is a more general implementation of other captures which it will eventually (but not yet) succeed. What do you think?

Xophmeister (commented):
This looks great, @prednaz: Thank you 💛

I particularly like your dedication to testing, which Topiary is not great for at the moment.

I've left some comments -- mostly trivial -- and will approve once resolved.

topiary-core/src/lib.rs
Xophmeister: In the book's style guide we say to use "multi-line", so in camel case, I think `MultiLineIndent` works :+1:

topiary-core/src/pretty.rs
prednaz: yes, whatever the line endings look like in the input, we produce `\n`s because tree-sitter strips all white space for us. multi line leafs are the exception. so this line provides consistency. @ErinvanderVeen and i discussed this on tuesday and tentatively concluded to keep it there. i have removed the comment in [`8f728e2`](https://github.com/topiary/topiary/pull/1266/changes/8f728e26e7d8872d475f6f36022db13deffb4360).

docs/book/src/reference/capture-names/indentation.md
prednaz: yes, [i tested it](https://github.com/topiary/topiary/blob/philipp/multi_line_ClosingColumnInsignificant/topiary-cli/tests/samples/input/nickel.ncl#L3185-L3186). `@multi_line_string.start` and `@multi_line_string.end` are captures, not predicates. the string delimiters are tracked in `Atom`s via [`EnforceIndentation`](https://github.com/topiary/topiary/pull/1266/changes#diff-c81cbc8a6e5ed8ff06a96beed38af39261458b6beb3c7e33bf32b5b840547a40R54).

just passing a string to a predicate is exactly what is not possible yet but might be useful for grammars that do not allow capturing the multi line string delimiters. for example, the nickel grammar's latest release does not allow capturing the delimiters of _single_ line strings. but then again, maybe the number of grammars not allowing capturing the multi line string delimiters is insignificant.

CHANGELOG.md
prednaz: do you mean `@multi_line_indent_all`. but `@multi_line_string` is not and will never be more general in the sense that it could emulate `@multi_line_indent_all`'s behavior.

it is more that `@multi_line_indent_all`'s behavior is not useful for almost all multi line leafs, especially multi line strings.

CHANGELOG.md
prednaz: ah, maybe i got a little carried away when fixing [some broken links](https://github.com/topiary/topiary/blob/2748460ea0a5e2ea4aa2a31f71a7f6b3fd4bd609/CHANGELOG.md?plain=1#L61-L62) and realizing that [github renders even the raw url just how we want it](https://github.com/topiary/topiary/blob/philipp/multi_line_ClosingColumnInsignificant/CHANGELOG.md). are you viewing the change log in a markdown viewer other than github?

docs/book/src/reference/capture-names/indentation.md
prednaz: implemented in [`f5a0587`](https://github.com/topiary/topiary/pull/1266/changes/f5a0587e943c7ecace7b4f7257fdc77f654fd43d)

docs/book/src/reference/capture-names/indentation.md
Xophmeister: Nice :pray:

CHANGELOG.md
Xophmeister: Oh, interesting. This is new. GitHub only used to render the full URL for its #-links in Markdown, with the nice/abbreviated version only working in PRs and comments, etc. They must have extended it to all Markdown :tada:

In which case, this is fine :+1:

CHANGELOG.md
Xophmeister: I remember there were some unimplemented enum variants you showed us on Tuesday. I assumed -- likely incorrectly :sweat_smile: -- that these were going to be extra powers added to `@multi_line_string` at some point. However, I'm realising now that maybe you are just setting the abstractions, ready for _other_ captures...which makes more sense.

If that's the case, then please ignore me and this comment can be resolved. That covers everything, so I'm happy to green-light this.

Xophmeister (approved):

topiary-core/src/pretty.rs
```diff
@@ -157,3 +168,1769 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        let common_whitespace_prefix_len_utf8: usize =
+            common_whitespace_prefix.map(char::len_utf8).sum();
+        let content =
+            content.map(|line| &line[line.len().min(common_whitespace_prefix_len_utf8)..]);
+
+        for line in content {
+            if line.is_empty() {
+                writeln!(buffer).unwrap();
+            } else {
+                write!(buffer, "\n{}{line}", indent.repeat(indent_level + 1)).unwrap();
```
mkatychev:
```suggestion
                writeln!(buffer, "{}{line}", indent.repeat(indent_level + 1)).unwrap();
```

prednaz: this does not do the same, does it?

CHANGELOG.md
prednaz: yes, you remember correctly. see https://github.com/topiary/topiary/pull/1266#discussion_r3583880037. and yes, at least `EnforceIndentation::ClosingColumnSignificant` is intended as a power added to `@multi_line_string` via a predicate. but i do not think a predicate should ever be added to `@multi_line_string` making it behave like `@multi_line_indent_all`.

topiary-core/src/pretty.rs
mkatychev: You're totally write 🤦 (pun intentional)

topiary-core/src/tree_sitter.rs
mkatychev: One of the issues we currently have is that we are unable to use injections managed in grammar repos because we are not conforming to the `set!` usage:
https://github.com/tree-sitter-grammars/tree-sitter-markdown/blob/split_parser/tree-sitter-markdown/queries/injections.scm

Given how captures work in community injections with `@injection.language` and `(#set! injection.language "toml") we can have some flexibility.

However we definitely should conform to at the very least the `set!` directive since it enables us to reuse existing grammar queries.

mkatychev: I see you've already filed the issue! https://github.com/topiary/topiary/issues/1290

Using `set!` is going to be awesome because we can start reusing injection queries from the community since  https://github.com/topiary/topiary/pull/1299 has already landed.

topiary-core/src/pretty.rs
```diff
@@ -69,22 +71,31 @@ pub fn render(atoms: &[Atom], indent: &str) -> FormatterResult<String> {
...
 
-                    // The following assumes spaces are used for indenting
-                    match indenting {
-                        0 => content.into(),
-                        n if n > 0 => add_spaces_after_newlines(content, indenting),
-                        _ => try_removing_spaces_after_newlines(content, -indenting),
+                        // The following assumes spaces are used for indenting
+                        match indenting {
+                            0 => content.into(),
+                            n if n > 0 => add_spaces_after_newlines(content, indenting),
```
mkatychev:
```suggestion
                            1.. => add_spaces_after_newlines(content, indenting),
```
A range might be a bit cleaner

mkatychev (approved):
LGTM! nice work!

topiary-core/src/pretty.rs
prednaz: completed in [`02f27ec`](https://github.com/topiary/topiary/pull/1266/changes/02f27eccc01044078643f229b81ef894aea29c00)

topiary-core/src/pretty.rs
prednaz: addressed in [`516efff`](https://github.com/topiary/topiary/pull/1266/changes/516efff61f298a095496966805116b5cdcbe9507).

topiary-core/src/pretty.rs
```diff
@@ -157,3 +170,1873 @@ fn try_removing_spaces_after_newlines(s: &str, n: i32) -> String {
...
+        start,
+        end,
+    } = absolute_indentation;
+
+    let content: Vec<&str> = content_original
+        .strip_prefix(start)
+        .ok_or_else(|| {
+            FormatterError::Query(format!(
+                "the multi line leaf starting with {:?} at {} should start with {start:?} as marked by the query.",
+                &content_original[..content_original.len().min(16)],
```
ErinvanderVeen: Slicing at an arbitrary byte-offset (16) can panic for multi-byte characters like 💛

Maybe:
```rust
let start_excerpt: String = content_original.chars().take(16).collect();
let end_excerpt: String = content_original
    .chars()
    .rev()
    .take(16)
    .collect::<String>()
    .chars()
    .rev()
    .collect();
```

See also line 207 of this file.

CHANGELOG.md
ErinvanderVeen: I disagree. There is no reason to reduce compatibility in this case, unless there is some great benefit to using GitHub specific features. This changelog will also be available on other platforms than GitHub.

topiary-core/src/lib.rs
```diff
@@ -41,13 +41,27 @@ pub struct ScopeInformation {
...
 }
 
-#[derive(Clone, Debug, Default, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Capitalisation {
     UpperCase,
     LowerCase,
-    #[default]
     Pass,
 }
```
ErinvanderVeen, line 50: Capitalisation is part of our public API. Without good reason, I don't think we ought to remove trait implementations. What is the reason for removing `Default` here?

ErinvanderVeen (changes requested):
Thanks for working on this. I do have several concerns about the details of the implementation that I think we should address before treating this as a stable, language-independent query feature.

### Verify CRLF behavior

Source: `topiary-core/src/pretty.rs:194`, `topiary-core/src/pretty.rs:297`.

The implementation splits on `\n`, strips a trailing `\r` from every resulting line, and emits `\n`. As a result, applying `@multi_line_string` normalizes CRLF sequences inside the captured literal to LF.

That makes me uncomfortable because these bytes are part of a string literal, not merely separators between ordinary source lines. If a host language preserves CRLF inside multiline strings, this changes the value of the program. We need either:

- a clear semantic justification that this normalization is value-preserving for every language allowed to use the capture; or
- an implementation that preserves the original line endings.

### Require exactly one start and one end delimiter

Source: `topiary-core/src/tree_sitter.rs:542`.

The current implementation independently calls `rfind` for `@multi_line_string.start` and `@multi_line_string.end`. A match containing multiple start or end captures therefore silently selects the last capture and can pair captures that were not intended to belong together.

For example, consider a hypothetical language in which multiline strings can contain nested strings, and a query accidentally captures both delimiter pairs:

```scheme
(multiline_string
  open: (delimiter) @multi_line_string.start

  (interpolation
    (string
      open: (delimiter) @multi_line_string.start
      close: (delimiter) @multi_line_string.end))

  close: (delimiter) @multi_line_string.end
) @multi_line_string
```

For source conceptually shaped like this:

```text
OUTER_OPEN text ${ INNER_OPEN value INNER_CLOSE } text OUTER_CLOSE
```

the match contains these captures:

```text
start = OUTER_OPEN
start = INNER_OPEN
end   = INNER_CLOSE
end   = OUTER_CLOSE
```

The independent searches select `INNER_OPEN` as the last start capture and `OUTER_CLOSE` as the last end capture. Topiary therefore constructs a delimiter pair that does not correspond structurally, and the selected start is not the prefix of the captured outer multiline string.

The query contract should require exactly one start capture and exactly one end capture per `@multi_line_string` match, and return a query error when that invariant is violated.

### Validate delimiter relationships while processing the match

Sources: `topiary-core/src/tree_sitter.rs:542`, `topiary-core/src/pretty.rs:194`.

Whether the captured start and end delimiters are actually the prefix and suffix of the `@multi_line_string` node is checked only during rendering. This validation belongs closer to query-match processing, where the captured nodes, their ranges, and their relationship to the outer node are all available.

Rendering should receive an already validated multiline-string representation rather than discovering query-definition errors while producing output. This would also simplify `render_enforced_indentation` and make its invariants easier to understand.

### Preserve precise missing-capture diagnostics

Sources: `topiary-core/src/tree_sitter.rs:548`, `topiary-core/src/atom_collection.rs:237`.

Using `Option::zip` collapses three distinct malformed-query cases into the same value:

- the start capture is missing;
- the end capture is missing;
- both captures are missing.

Please validate the captures separately and report exactly which required capture is absent. That will make custom query failures substantially easier to diagnose.

### Rename `multi_line_indent_all`

Sources: `topiary-core/src/lib.rs:85`, `topiary-core/src/atom_collection.rs:28`, `topiary-core/src/atom_collection.rs:493`.

`multi_line_indent_all` used to be a Boolean, but is now a three-way `MultiLineIndent` policy. Its name no longer describes the value it contains. A name such as `multiline_indentation` or `indentation_policy` would be clearer in `Atom::Leaf`, `LeafFlagsMut`, and the related methods.

### Give `EnforceIndentation` a domain-specific name

Source: `topiary-core/src/lib.rs:51`.

`EnforceIndentation` is too broad for a type that specifically stores multiline-string delimiters and closing-line behavior. A name such as `MultilineStringIndentation` or `DelimitedMultilineIndentation` would make its purpose and valid use much clearer to me.

### Clean up inconsistent and stale terminology

Source: `topiary-core/src/pretty.rs:174`.

The rendering helper receives a value named `absolute_indentation`, while its documentation says it assumes a `ClosingColumnInsignificant` constructor. There is no such constructor in the submitted types. The names and documentation should consistently describe the final model rather than earlier iterations of the design.

### Avoid expanding the public API unnecessarily

Source: `topiary-core/src/lib.rs:51`.

`EnforceIndentation` and `MultiLineIndent` are declared at the crate root with public fields and variants. If these are implementation details of atom rendering, they should be private or `pub(crate)`. Publishing them commits `topiary-core` to their current names and representation and makes future refinement a breaking change.

### Split `render_enforced_indentation` into focused stages

Source: `topiary-core/src/pretty.rs:181`.

The helper currently performs all of the following:

- delimiter validation;
- line-ending normalization;
- boundary-line removal;
- common-prefix calculation;
- mixed-whitespace diagnostics;
- output rendering.

Separating validation, content analysis, and rendering would make the semantic invariants easier to audit and test. In particular, the renderer should not need to know how malformed query captures are diagnosed.

### Make error handling internally consistent

Sources: `topiary-core/src/pretty.rs:219`, `topiary-core/src/pretty.rs:37`.

The new helper uses repeated `unwrap()` calls when writing to a `String`, while the surrounding renderer consistently uses `context_to()?`. Writing to a `String` is effectively infallible, so this is not a likely runtime failure, but we should either encapsulate that assumption or follow the established error-handling style consistently.

### Decide whether mixed indentation is actionable

Source: `topiary-core/src/pretty.rs:269`.

The implementation detects mixed whitespace prefixes but only logs the condition at info level, which normal users are unlikely to see.

If mixed indentation can prevent value preservation, this should be surfaced as a warning or query/formatting error. If the transformation is guaranteed to remain value-preserving, the diagnostic may not be needed. In either case, the intended safety property should be documented and tested.

### Make the unit tests table-driven

Source: `topiary-core/src/pretty.rs:312`.

The new rendering tests expand `pretty.rs` from roughly 150 lines to more than 2,000, with a large amount of repeated setup. Most cases vary only by input, expected output, indentation, and the final-line flag.

A table-driven helper would retain the coverage while making the behavior matrix, missing cases, and future changes much easier to review. It would also keep the production module navigable.

topiary-core/src/lib.rs
prednaz: sorry, the reason was that it is unused and i did not realize that this is part of the public api.

reverted in [`bf5d6cd`](https://github.com/topiary/topiary/pull/1266/changes/bf5d6cdf53e58ee540cc4ed7bfeacae11aef98b9)

prednaz:
notice that ci is [has failed as follows most recently](https://github.com/topiary/topiary/actions/runs/31064075155/job/92497988072).
```
 ● Libloading error: DlOpen { source: "/home/runner/.cache/topiary/ocaml_interface/a45fda5fe73cda92f2593d16340b3f6bd46674b8.so: file too short" }
 ├ topiary-cli/src/config.rs:122
 ╰ See Error Span #1 below

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Error Span #1

   × Parsing error: missing range

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

```

prednaz:
thanks for the thorough review, @ErinvanderVeen! i have started addressing your points in individual commits.

> ### Require exactly one start and one end delimiter
> [...]
> The current implementation independently calls `rfind` for `@multi_line_string.start` and `@multi_line_string.end`. A match containing multiple start or end captures therefore silently selects the last capture and can pair captures that were not intended to belong together.
> 
> For example, consider a hypothetical language in which multiline strings can contain nested strings, and a query accidentally captures both delimiter pairs:
> [...]
> For source conceptually shaped like this:
> 
> ```
> OUTER_OPEN text ${ INNER_OPEN value INNER_CLOSE } text OUTER_CLOSE
> ```
> [...]
> The independent searches select `INNER_OPEN` as the last start capture and `OUTER_CLOSE` as the last end capture. Topiary therefore constructs a delimiter pair that does not correspond structurally, and the selected start is not the prefix of the captured outer multiline string.
> 
> The query contract should require exactly one start capture and exactly one end capture per `@multi_line_string` match, and return a query error when that invariant is violated.

yeah, i noticed this as a more general problem, which i described in the last paragraph of https://github.com/topiary/topiary/pull/1266#discussion_r3583977578.

`#delimiter!` and `#scope_id!` exhibit the same problem and even worse so because they would silently do the wrong thing. whereas `@multi_line_string.start` and `@multi_line_string.end` will never silently do the wrong thing but will either accidentally do the right thing or raise an error, notifying the user of the problem, usually the query author who can then fix her query.

fixing this for `@multi_line_string.start` but not for the worse `#delimiter!` seemed inconsistent and wrong because it treats a single symptom of what i perceive as an underlying problem. fixing the perceived underlying problem seemed out of the scope of this pull request and worthy of prior discussion. so instead, i decided to record my perception of the underlying in problem in https://github.com/topiary/topiary/pull/1266#discussion_r3583977578 in the last paragraph, start the discussion of a fix, and match the prevailing practice for now. which is to consider only the last occurrence, which is why i used `rfind` instead of `find`.

topiary-core/src/tree_sitter.rs
ErinvanderVeen: I agree with your assessment here, but I would not consider it in-scope for this PR. Keep it as a refactor for the next release.

topiary-core/src/pretty.rs
prednaz: implemented in [`917dad7`](https://github.com/topiary/topiary/pull/1266/changes/917dad7a677c8a4e4557a3714b43d0aa00559baf)

ErinvanderVeen:
> thanks for the thorough review, @ErinvanderVeen! i have started addressing your points in individual commits.
> 
> > ### Require exactly one start and one end delimiter
> > [...]
> > The current implementation independently calls `rfind` for `@multi_line_string.start` and `@multi_line_string.end`. A match containing multiple start or end captures therefore silently selects the last capture and can pair captures that were not intended to belong together.
> > For example, consider a hypothetical language in which multiline strings can contain nested strings, and a query accidentally captures both delimiter pairs:
> > [...]
> > For source conceptually shaped like this:
> > ```
> > OUTER_OPEN text ${ INNER_OPEN value INNER_CLOSE } text OUTER_CLOSE
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [...]
> > The independent searches select `INNER_OPEN` as the last start capture and `OUTER_CLOSE` as the last end capture. Topiary therefore constructs a delimiter pair that does not correspond structurally, and the selected start is not the prefix of the captured outer multiline string.
> > The query contract should require exactly one start capture and exactly one end capture per `@multi_line_string` match, and return a query error when that invariant is violated.
> 
> yeah, i noticed this as a more general problem, which i described in the last paragraph of [#1266 (comment)](https://github.com/topiary/topiary/pull/1266#discussion_r3583977578).
> 
> `#delimiter!` and `#scope_id!` exhibit the same problem and even worse so because they would silently do the wrong thing. whereas `@multi_line_string.start` and `@multi_line_string.end` will never silently do the wrong thing but will either accidentally do the right thing or raise an error, notifying the user of the problem, usually the query author who can then fix her query.
> 
> fixing this for `@multi_line_string.start` but not for the worse `#delimiter!` seemed inconsistent and wrong because it treats a single symptom of what i perceive as an underlying problem. fixing the perceived underlying problem seemed out of the scope of this pull request and worthy of prior discussion. so instead, i decided to record my perception of the underlying in problem in [#1266 (comment)](https://github.com/topiary/topiary/pull/1266#discussion_r3583977578) in the last paragraph, start the discussion of a fix, and match the prevailing practice for now. which is to consider only the last occurrence, which is why i used `rfind` instead of `find`.

I agree that fully parsing captures and predicates into a typed representation is a broader refactor and should remain out of scope. However, requiring exactly one start and end capture is a local invariant of this new directive and does not require that refactor. Prefix/suffix validation is not equivalent because it compares text rather than capture ranges; an incorrectly selected delimiter can still be a textual prefix or suffix. I would still like this PR to reject matches with duplicate start or end captures.

ErinvanderVeen (changes requested):

prednaz:
i have added the uniqueness constraint, @ErinvanderVeen. shall i add it for `#delimiter!` and `#scope_id!` too?

ErinvanderVeen (approved):
No, this is good!

PR #1254, opened 2026-06-04
Markdown Injections!

ErinvanderVeen:
Resolves #568

## Description
This PR adds Markdown formatting (although it's very barebones). More importantly, it formats the codeblocks inside of the markdown document! Noteably: in order to prevent errors when resolving unknown languages (like `text` in the test), the `resolve_` functions were changed to return `None` instead of an error when the language is unknown. The CLI then turns this into an error when it's the language of the top document, and a warning if it is an injected language.

## Checklist

Checklist before merging, wherever relevant:

- [x] `CHANGELOG.md` updated
- [x] Documentation (The Topiary Book, `README.md`, etc.) up-to-date
- [ ] Updated regression tests

Issue #824, opened 2025-01-06
Using Tree-sitter predicates requires us to use Topiary capture names


Xophmeister:
Tree-sitter comes with a bunch of [built-in predicates](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/3-predicates-and-directives.html), which work using the capture names. However, in Topiary, we only allow our formatting capture names, such as `@append_hardline`, etc. While these work in the predicates, it makes them a bit strange to read. For example:

```scheme
; Append a space to (foo (bar)) nodes if they begin with a capital letter
(foo
  (bar) @append_space

  (#match? @append_space "^[A-Z].*")
)
```

While this would just be a nice-to-have, maybe it would be worthwhile whitelisting certain patterns for capture names that have no formatting functionality, but make the predicates more intuitive. For example, allowing `@topiary\..+`:

```scheme
(foo
  (bar) @topiary.foobar @append_space

  (#match? @topiary.foobar "^[A-Z].*")
)
```

A potential issue with this is that, IIRC, the number of capture names per node is limited to something unreasonably small (like 3).

Xophmeister:
This seems to be working now (see #1267), but it's not clear how/why :shrug: Tentatively closing.

PR #1267, opened 2026-06-18
feat(injections): Add initial injection support for rust

### LINKED ISSUE #1251: Support `serde_json::json!()` macro formatting using injections
With addition of https://github.com/topiary/topiary/pull/1244/, json formatting inside rust (with supported languages) is now possible.

It would have to be a different query (likely) than the current rust formatter since the want should only touch the injected sections (and let `rustfmt` handle the rest).

mkatychev:
# Add initial injection support for rust

Resolves #XXX
References #1254 #1244

## Description

Added initial injection support in rust for [json macros](https://docs.rs/serde_json/latest/serde_json/macro.json.html):
https://github.com/topiary/topiary/blob/1bb02cd59ace5bd7d886f2d37eff3ca1065561af/topiary-cli/tests/samples/expected/rust.rs#L113-L120

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ This name should be decided amongst the team before the release.
 - [#1232](https://github.com/topiary/topiary/pull/1232) Rework `topiary_core::ErrorSpan` to handle `QueryError`s as well as `Node` errors by introducing `rootcause::Report`
 - [#1249](https://github.com/topiary/topiary/pull/1249) Remove async calls and setup from `format_nickel` benchmark
 - [#1254](https://github.com/topiary/topiary/pull/1254) Topiary now gracefully skips formatting for unknown or unsupported injected languages instead of returning a fatal error, allowing files with unconfigured code blocks (like Markdown) to format successfully.
+- [#1267](https://github.com/topiary/topiary/pull/1267) Add initial injection support for rust
 
 ### Fixed
 - [#1176](https://github.com/topiary/topiary/pull/1176) Increase the stack size to 4MiB in Windows builds.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ This name should be decided amongst the team before the release.
 - [#1232](https://github.com/topiary/topiary/pull/1232) Rework `topiary_core::ErrorSpan` to handle `QueryError`s as well as `Node` errors by introducing `rootcause::Report`
 - [#1249](https://github.com/topiary/topiary/pull/1249) Remove async calls and setup from `format_nickel` benchmark
 - [#1254](https://github.com/topiary/topiary/pull/1254) Topiary now gracefully skips formatting for unknown or unsupported injected languages instead of returning a fatal error, allowing files with unconfigured code blocks (like Markdown) to format successfully.
-- [#1267](https://github.com/topiary/topiary/pull/1267) Add initial injection support for rust
+- [#1267](https://github.com/topiary/topiary/pull/1267) Add initial injection support (JSON) for Rust
 
 ### Fixed
 - [#1176](https://github.com/topiary/topiary/pull/1176) Increase the stack size to 4MiB in Windows builds.
--- a/topiary-cli/tests/samples/expected/rust.rs
+++ b/topiary-cli/tests/samples/expected/rust.rs
@@ -109,3 +109,15 @@ impl MyTrait for MyStruct {
         // ... logic ...
     }
 }
+
+const JSON: Foo = json!({
+  "code": 200,
+  "success": true,
+  "payload": {
+    "features": [
+      "serde",
+      "json"
+    ],
+    "homepage": null
+  }
+});
--- a/topiary-cli/tests/samples/expected/rust.rs
+++ b/topiary-cli/tests/samples/expected/rust.rs
@@ -110,11 +110,12 @@ impl MyTrait for MyStruct {
     }
 }
 
-const JSON: Foo = json!({
-  "code": 200,
-  "success": true,
-  "payload": {
-    "features": [ "serde", "json" ],
-    "homepage": null
-  }
-});
+const JSON: Foo = json!({"code": 200,
+"success": true,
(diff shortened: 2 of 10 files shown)


topiary-queries/queries/rust/injections.scm
```diff
@@ -0,0 +1,54 @@
...
+    ]
+    (token_tree
+      (token_tree . "{" "}" .) @injection.content
+    )
+  )
+  (#eq? @_macro_name "json")
+  (#injection_language! "json")
+)
+
+; TODO: implement `strip!` or `@injection.combined`
```
mkatychev, line 17: Since TOML does not have delimiters for span start/end, we would need to either `strip!` the first and last character of the `(token_tree)` or call `@injection.combined` on all inner nodes
```rust
let cargo_toml = toml::toml! {
    [package]
    name = "toml"

    [dependencies]
    serde = "1.0"

    [dev-dependencies]
    serde_derive = "1.0"
    serde_json = "1.0"
};
```

topiary-cli/tests/samples/expected/rust.rs
```diff
@@ -109,3 +109,12 @@ impl MyTrait for MyStruct {
...
 }
+
+const JSON: Foo = json!({
+  "code": 200,
+  "success": true,
+  "payload": {
+    "features": [ "serde", "json" ],
+    "homepage": null
+  }
+});
```
mkatychev: indentation is not preserved for macros at the moment, hence the `const`

topiary-queries/queries/rust/injections.scm
```diff
@@ -0,0 +1,55 @@
...
+;        (identifier) @_macro_name
+;      ]
+;     (#match? @_macro_name "(bindgen|generate)")
+;   (token_tree
+;     (token_tree
+;       (identifier) @key
+;       (#eq? @key "inline")
+;       .
+;       [
+;       ; (string_literal (string_content) @injection.content)
```
mkatychev, line 49: main issue with WIT formatting in injections is that `(string_literal)` is a leaf in the formatter:
https://github.com/topiary/topiary/blob/a582e0f2e622b7bc2529f4e2358b4737c2f55bff/topiary-queries/queries/rust/formatting.scm#L1-L8

...however there are other issues at play with wanting to preserve all but certain nodes:

```tree-sitter-query
(macro_invocation
  (token_tree
    (identifier) @injection.content
    (#eq? @key "foo-lang")
  ) @leaf ;; treat all but `@injection.content` tagged nodes as leaves
```

topiary-core/src/lib.rs
```diff
@@ -437,7 +437,10 @@ fn resolve_injected_language(
     };
 
     match resolve(language) {
-        Ok(Some(language)) => Ok(Some(language)),
+        Ok(Some(language_cfg)) => {
+            log::warn!("resolved injected language: {language}");
```
```suggestion
            log::info!("resolved injected language: {language}");
```

ErinvanderVeen:
AFAIK, `json! { [1,2,3] }` is also valid. As is `json!("test": true)`. These are not caught by your queries.

mkatychev:
Good catch though the floating key value pair is not valid JSON, so I'm not sure how we would handle that. I feel like ignoring it makes sense for now.

topiary-queries/queries/rust/injections.scm
```diff
@@ -0,0 +1,55 @@
+; https://docs.rs/serde_json/latest/serde_json/macro.json.html
+; https://github.com/helix-editor/helix/blob/43bf7c2dc219606c64003aef21151f49f48d0939/runtime/queries/rust/injections.scm#L42-L52
+(
+  (macro_invocation
+    macro: [
+      (scoped_identifier name: (_) @_macro_name)
```
Xophmeister, line 6: This is probably nothing new and I've just had my eye off the ball, but this implies #824 has been fixed. Is that the case?

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ This name should be decided amongst the team before the release.
 - [#1232](https://github.com/topiary/topiary/pull/1232) Rework `topiary_core::ErrorSpan` to handle `QueryError`s as well as `Node` errors by introducing `rootcause::Report`
 - [#1249](https://github.com/topiary/topiary/pull/1249) Remove async calls and setup from `format_nickel` benchmark
 - [#1254](https://github.com/topiary/topiary/pull/1254) Topiary now gracefully skips formatting for unknown or unsupported injected languages instead of returning a fatal error, allowing files with unconfigured code blocks (like Markdown) to format successfully.
+- [#1267](https://github.com/topiary/topiary/pull/1267) Add initial injection support for rust
```
Xophmeister:
```suggestion
- [#1267](https://github.com/topiary/topiary/pull/1267) Add initial injection support (JSON) for Rust
```

Xophmeister (approved):
Neat :+1:

I assume this PR will just have JSON support, which is totally reasonable. It might be good to update `docs/book/src/reference/language-support.md` to mention this; it's genuinely novel and something `rustfmt` can't do.

topiary-queries/queries/rust/injections.scm
mkatychev: Oh that's a good callout, I don't know at all, seems like a good test case.

mkatychev: We have only three capture groups in that s-expression, is the implication that had there been a 4th then there would be undefined behaviour?

Xophmeister: No, not the 3 capture limit, but the fact that you can do `@something`, where `something` isn't one of our standard formatting captures. That always used to raise an "Unknown capture" (or something like that) error.

mkatychev: Oh, I guess the tree-sitter lib fixed it at some point because it works!

mkatychev:
@Xophmeister I totally forgot about https://github.com/topiary/topiary/pull/1267#pullrequestreview-4611846406, will have to do in follow-up

PR #1275, opened 2026-06-29
chore: add verify-documented-usage to nix flake check

ErinvanderVeen:
# add verify-documented-usage to nix flake check

## Description
Ideally, running `nix flake check` should run all tests and checks.

## Checklist

- [ ] `CHANGELOG.md` updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,6 @@ jobs:
       - name: Build client-app example
         run: nix build .#client-app -L
 
-      - name: Verify that usage in README.md matches CLI output
-        run: nix run .#verify-documented-usage -L
-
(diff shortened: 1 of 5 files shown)


Xophmeister (commented):
Forcing the 90 column output to work consistently is a real PITA 😞 The way I found was the least flaky, but if you were getting reproducibility issues with that, then I guess even that wasn't fool proof...but now the CI is failing 🙀

IIRC, one can set clap's output width with a parameter. The most reliable way of getting this to work may be using that and triggering it on a magic environment variable. A bit invasive, though, just for the sake of a test 🤷

ErinvanderVeen:
Closed because of failing CI, needs rework.

PR #1277, opened 2026-06-30
Small refactors on a heap

ErinvanderVeen:
# Small refactors on a heap

## Description
What can I say, these are just all small changes that do not deserve their own PR.

Please look at all commits individually. They do not make much sense otherwise.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/topiary-core/src/atom_collection.rs
+++ b/topiary-core/src/atom_collection.rs
@@ -111,7 +111,7 @@ impl AtomCollection {
             counter: 0,
         };
 
-        atoms.collect_leaves_inner(root, source, &Vec::new(), 0)?;
+        atoms.collect_leaves_inner(root, source, 0)?;
 
         Ok(atoms)
     }
@@ -567,7 +567,6 @@ impl AtomCollection {
     ///
     /// * `node` - The current node to process.
     /// * `source` - The full source code as a byte slice.
-    /// * `parent_ids` - A vector of node ids that are the ancestors of the current node.
     /// * `level` - The depth of the current node in the CST tree.
     ///
     /// # Errors
@@ -577,11 +576,9 @@ impl AtomCollection {
         &mut self,
         node: &Node,
         source: &[u8],
-        parent_ids: &[usize],
         level: usize,
     ) -> FormatterResult<()> {
         let id = node.id();
-        let parent_ids = [parent_ids, &[id]].concat();
 
         log::debug!(
             "CST node: {}{} - Named: {}",
@@ -613,7 +610,7 @@ impl AtomCollection {
             self.mark_leaf_parent(node, node.id());
         } else {
             for child in node.children(&mut node.walk()) {
-                self.collect_leaves_inner(&child, source, &parent_ids, level + 1)?;
+                self.collect_leaves_inner(&child, source, level + 1)?;
             }
         }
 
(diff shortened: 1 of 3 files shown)


topiary-core/src/atom_collection.rs
```diff
@@ -1114,9 +1110,20 @@ impl AtomCollection {
...
+        self.edge_leaf(node, |n| n.child(n.child_count() - 1).unwrap())
+    }
+
+    /// Iteratively descends from `node` towards a leaf, picking the next child
+    /// via `pick_child` at each step. Stops when the current node has no
+    /// children or is registered in `specified_leaf_nodes`.
+    fn edge_leaf<'tree, 'node: 'tree>(
+        &self,
+        node: &'node Node<'tree>,
+        pick_child: impl Fn(&Node<'tree>) -> Node<'tree>,
```
Xophmeister: This is fine, but I wonder if it would be cleaner to just pass an index with wraparound semantics for negative numbers?

Xophmeister (approved):
Nice 🧑‍🍳👌

topiary-core/src/atom_collection.rs
mkatychev: something like `impl Index<isize> for AtomCollection`?

Xophmeister: I was just thinking of having a simple `index: isize` argument and containing the wraparound logic in this function, but yeah, an `Index<isize>` impl would be very Rusty :crab:

topiary-core/src/tree_sitter.rs
```diff
@@ -654,74 +659,52 @@ fn handle_predicate(
...
     let operator = &*predicate.operator();
-    if "delimiter!" == operator {
-        let arg = predicate
-            .args()
-            .into_iter()
-            .next()
-            .ok_or_else(|| FormatterError::Query(format!("{operator} needs an argument")))?;
-        Ok(QueryPredicates {
-            delimiter: Some(arg),
+    match operator {
```
mkatychev, line 650: match arm is much cleaner

topiary-core/src/tree_sitter.rs
```diff
@@ -152,6 +139,23 @@ impl InjectionQuery {
...
+    grammar: &topiary_tree_sitter_facade::Language,
+    query_content: &str,
+) -> FormatterResult<Query, QueryError> {
+    Query::new(grammar, query_content)
+        .into_report()
+        .map_err(|e| {
+            let range = e.current_context().range;
+            e.attach_range(range)
+        })
+        .attach_source(Some(query_content))
```
mkatychev: I think I got to this first 😅 
https://github.com/topiary/topiary/blob/2b6f6d4a9e09bce8c643b64edf6d6154c12dbaff/topiary-core/src/tree_sitter.rs#L67-L69
https://github.com/topiary/topiary/blob/2b6f6d4a9e09bce8c643b64edf6d6154c12dbaff/topiary-core/src/error/mod.rs#L181-L188

mkatychev (approved):

topiary-core/src/atom_collection.rs
ErinvanderVeen: The trait impl doesn't work. There is already an implementation for `SliceIndex<[Atom]>` and rustc won't allow us to implement one for `isize`, because that could in the future implement `SliceIndex<[...]>`.

Also, in general, I don't quite see how a trait implementation for `AtomCollection` will help us find the children of type `Node`.

PR #1278, opened 2026-07-01
verify usage in flake check, use clap to set term_width for reproducibility

### LINKED ISSUE #852: `bin/verify-documented-usage.sh` is not environment agnostic
**Describe the bug**
The `bin/verify-documented-usage.sh` script checks that the usage documentation in the `README` matches the various Topiary CLI subcommand `--help` outputs. However, it does not account for line wrapping in the `--help` outputs, which appear to be environment-dependent. (My suspicion is the TTY of the process, if there is one.) 

This leads to false negatives.

**To Reproduce**
In the `nix develop` dev-shell, run `verify-documented-usage`. It will _probably_ complain that the `README` is out-of-sync with at least one of the subcommands' `--help` output. (If not, resize your terminal until it does!)

**Expected behaviour**
The output produced by the script should be wrapped at a set width (80 columns, say), so it's deterministic.

**Environment**
 - OS name + version: NixOS 24.11
 - Version of the code: d61664f235

ErinvanderVeen:
# verify usage in flake check, use clap to set term_width for reproducibility

## Description

[PR Description]

## Checklist

- [ ] `CHANGELOG.md` updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,6 @@ jobs:
       - name: Build client-app example
         run: nix build .#client-app -L
 
-      - name: Verify that usage in README.md matches CLI output
-        run: nix run .#verify-documented-usage -L
-
       - name: Test against config regressions
         run: |
             nix run . -- config --merge --cfg ./topiary-config/tests/issue-1124.ncl
--- a/bin/verify-documented-usage.sh
+++ b/bin/verify-documented-usage.sh
@@ -4,12 +4,14 @@
 
 set -euo pipefail
 
+export __TOPIARY_TERM_WIDTH=90
+
 readonly FENCE='```'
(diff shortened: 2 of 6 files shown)


Xophmeister (approved):
Looks good :+1: Just two nitpick suggestions:

1. The `topiary-wrapped` Nix package is redundant, so it can be removed and you can revert to using good old `topiary`.
2. I wonder if the `TOPIARY_TEST_TERM_WIDTH` should be named something that signals its internal, "do not set this yourself" nature; like `__TOPIARY_TERM_WIDTH`, or maybe even using a feature gate instead of an envvar. I'm just thinking out aloud; it's probably fine as is :+1:

prednaz (approved):
thank you for finding this problem and its solution.

PR #1279, opened 2026-07-01
Menhir support

ErinvanderVeen:
# Menhir support

## Description
Adds Menhir queries as an experimental language.

## Checklist

- [x] `CHANGELOG.md` updated


topiary-cli/tests/samples/expected/menhir.mly
mkatychev, line 1: Could we cut this file down to around a thousand lines? I imagine this will take tests a bit longer to go through.

ErinvanderVeen: On it!

Xophmeister (approved):
Awesome :+1:

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,798 @@
...
+
+(*-----------------------------------------*) (* The normal tokens *) (*-----------------------------------------*)
+
+(* tokens with "values" *)%token <string * AST_python.tok> NAME
+%token <int option    * AST_python.tok> INT LONGINT
+%token <float option  * AST_python.tok> FLOAT
+%token <string * AST_python.tok> IMAG
+%token <string * string * AST_python.tok> STR
+
+(*-----------------------------------------*) (* Keyword tokens *) (*-----------------------------------------*)%token <AST_python.tok> IF ELSE ELIF WHILE FOR RETURN CONTINUE BREAK PASS DEF LAMBDA CLASS GLOBAL TRY FINALLY EXCEPT RAISE AND NOT OR IMPORT FROM AS DEL IN IS WITH YIELD ASSERT NONE TRUE FALSE ASYNC AWAIT NONLOCAL (* python2: *)PRINT EXEC
```
mkatychev: are these big line widths expected/desired?

ErinvanderVeen: Absolutely not! Thank you, I somehow missed them. Working on it now.

mkatychev (approved):
LGTM

ErinvanderVeen:
@Niols Can I kindly ask you to look at this?

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+  }
+
+list_for_rest:
+  | or_test                   { [$1] }
+  | or_test "," list_for_rest { $1::$3 }
+
+listsync_comp_for:
+  | sync_comp_for { $1 }
+  (* python2-ext: [x for x in 1, 2] *)
+  | FOR exprlist IN list_for { [CompFor (tuple_expr_store $2, $4) ] }
```
aspiwack: We could break this line. _e.g._


```suggestion
  | FOR exprlist IN
list_for { [CompFor (tuple_expr_store $2, $4) ] }
```

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+  | expr BITOR xor_expr { BinOp ($1, (BitOr,$2), $3) }
+
+
+xor_expr:
+  | and_expr                 { $1 }
+  | xor_expr BITXOR and_expr { BinOp ($1, (BitXor,$2), $3) }
+
+and_expr:
+  | shift_expr                 { $1 }
+  | shift_expr BITAND and_expr { BinOp ($1, (BitAnd,$2), $3) }
```
aspiwack: Let's remove some spaces here and see the result not change, probably.

```suggestion
  |shift_expr BITAND and_expr{BinOp($1,(BitAnd,$2),$3)}
```

(while making this suggestion, I noticed that spaces are already inconsistent in this line; but it's good to have more situations)

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+
+(*----------------------------*)
+(* Atom *)
+(*----------------------------*)
+
+atom:
+  | NAME        { Name ($1, Load, ref NotResolved) }
+
+  | INT         { Num (Int ($1)) }
+  | LONGINT     { Num (LongInt ($1)) }
```
aspiwack: Let's put these two on the same line, this should work.

```suggestion
  | INT         { Num (Int ($1)) } | LONGINT     { Num (LongInt ($1)) }
```

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+fstring_print_spec:
+  |     fstring_format_clause { $1 }
+  | "=" fstring_format_clause { mk_str $1::$2 }
+
+fstring_format_clause:
+  | (*empty*) { [] }
+  | fstring_format_delimeter format_specifier { mk_str $1::$2 }
+
+fstring_format_delimeter:
+  | ":" { $1 }
```
aspiwack: Let's move these two on the same line

```suggestion
fstring_format_delimeter: | ":" { $1 }
```

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+  | STR { let (s, pre, tok) = $1 in
+          if pre = "" then Str (s, tok) else EncodedStr ((s, tok), pre) }
+  | FSTRING_START interpolated* FSTRING_END { InterpolatedString $2 }
+
+interpolated:
+  | FSTRING_STRING { Str $1 }
+  | FSTRING_LBRACE interpolant fstring_print_spec "}" { InterpolatedString ($2::$3) }
+
+fstring_print_spec:
+  |     fstring_format_clause { $1 }
```
aspiwack: Let's move these two on the same line _and_ remove the optional first `|`

```suggestion
fstring_print_spec: fstring_format_clause { $1 }
```

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+  | test                 { Single $1 }
+  | test "," testlist1 { cons $1 $3 }
+
+(*----------------------------*)
+(* strings *)
+(*----------------------------*)
+
+string:
+  | STR { let (s, pre, tok) = $1 in
+          if pre = "" then Str (s, tok) else EncodedStr ((s, tok), pre) }
```
aspiwack, line 757: And remove the optional first `|` here.

```suggestion
string:
  STR { let (s, pre, tok) = $1 in
          if pre = "" then Str (s, tok) else EncodedStr ((s, tok), pre) }
```

topiary-cli/tests/samples/input/menhir.mly
```diff
@@ -0,0 +1,1018 @@
...
+let to_list = function
+  | Single e -> [e]
+  | Tup l -> l
+
+(* TODO: TypedExpr? ExprStar? then can appear as lvalue
+ * CompForIf though is not an lvalue.
+*)
+let rec set_expr_ctx ctx = function
+  | Name (id, _, x) ->
+      Name (id, ctx, x)
```
aspiwack: Play with the spaces in the Ocaml bits to see the language injection acting.

```suggestion
  | Name(id, _,x) ->
      Name(id,ctx,x)
```

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+%token <AST_python.tok> IF ELSE ELIF WHILE FOR RETURN CONTINUE BREAK PASS DEF LAMBDA CLASS GLOBAL TRY FINALLY EXCEPT RAISE AND NOT OR IMPORT FROM AS DEL IN IS WITH YIELD ASSERT NONE TRUE FALSE ASYNC AWAIT NONLOCAL
+(* python2: *)
+PRINT EXEC
+
+(*-----------------------------------------*)
+(* Punctuation tokens *)
+(*-----------------------------------------*)
+
+(* syntax *)
+%token <AST_python.tok> LPAREN"(" RPAREN")" LBRACK"[" RBRACK"]" LBRACE"{" RBRACE"}" COLON":" SEMICOL";" DOT"." COMMA"," BACKQUOTE"`" AT"@" ELLIPSES"..." LDots"<..." RDots"...>"
```
aspiwack: Unclear why these are all on a single line now. They're not on a single line in the source.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+
+(* operators *)
+%token <AST_python.tok> ADD (* + *) SUB (* - *)
+MULT"*" (* * *) DIV"/" (* / *)
```
aspiwack: But these one stayed two per line. Curious.

I think the multi-line form for `%token` should probably have `%token` alone on its line, and one token per line below.

Also there should be a space between the token name and its corresponding string. Like `MULT "*"`.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+(*************************************************************************)
+(* Import *)
+(*************************************************************************)
+(* In Python, imports can actually appear not just at the toplevel *)
+import_stmt:
+| import_name { $1 }
+| import_from { $1 }
+
+import_name: IMPORT list_sep(dotted_as_name, ",") { $2 |> List.map (fun (v1, v2) -> let dots = None in
+                                                         ImportAs ($1, (v1, dots), v2))   }
```
aspiwack: This is another example of something brought to a single line (sort of) when it wasn't in the input.

Also it appears that the Ocaml on the right hasn't been formatted correct?

aspiwack (commented):
While it's not super easy to understand what's going on in a file that large. I've found a few places to add funny behaviour, and a few places with unexpected behaviour currently.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+(*************************************************************************)
+%start <AST_python.program> main
+%start <AST_python.any> sgrep_spatch_pattern
+%type <int> dummy_rule %%
+
+(*************************************************************************)
+(* Macros *)
+(*************************************************************************)
+
+list_sep(X,Sep):
```
Niols: nit: space after the comma

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+| NEWLINE { [] }
+| stmt { $1 }
+
+sgrep_spatch_pattern:
+| stmt NEWLINE? EOF { match $1 with
+                      | [ExprStmt x] -> Expr x
+                      | [x] -> Stmt x
+                      | xs -> Stmts xs
+                    }
+| stmt stmt+ NEWLINE? EOF { Stmts ($1 @ (List.flatten $2)) }
```
Niols: On input
```
sgrep_spatch_pattern:
| stmt NEWLINE? EOF {
    match $1 with
    | [ExprStmt x] -> Expr x
    | [x] -> Stmt x
    | xs -> Stmts xs
  }
| stmt stmt+ NEWLINE? EOF { Stmts ($1 @ (List.flatten $2)) }
```
would we leave as-is or would we bring back the formatting as above?

topiary-cli/tests/samples/expected/menhir.mly
Niols: Also spaces before closing brace `}`. If it's single lined it should be only one space, and if it's multi line it should be on its own line.

Could it be that the `$` injections of Menhir make it an invalid OCaml fragment, the OCaml query fails, and the result is then all weird like this?

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+| FROM name_and_level IMPORT "(" list_comma(import_as_name) ")" { [ImportFrom ($1, $2, $5)] }
+| FROM name_and_level IMPORT list_comma(import_as_name) { [ImportFrom ($1, $2, $4)] }
+
+name_and_level:
+| dot_level dotted_name { match $1 with
+                          | [] -> $2, None
+                          | dl -> $2, Some dl
+                        }
+| "." dot_level { [("",$1(*TODO*))], Some ($1 :: $2) }
+| "..." dot_level { [("",$1(*TODO*))], Some ($1:: $2) }
```
Niols: Another example of OCaml not having been formatted correclty I think, probably because of the `$`. In particular, we see inconsistencies around `::` and we're missing a lot of spacing, it should be something like
```menhir
| "." dot_level { [("", $1 (*TODO*))], Some ($1 :: $2) }
| "..." dot_level { [("", $1 (*TODO*))], Some ($1 :: $2) }
```
maybe

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,916 @@
...
+| FOR exprlist IN or_test comp_iter { [CompFor (tuple_expr_store $2, $4)] @ $5 }
+
+(* support mixed python2 / python3 comp_for only when used inside atom_list *)
+listcomp_for:
+| listsync_comp_for { $1 }
+| ASYNC listsync_comp_for { (* TODO *) $2 }
+
+list_for:
+or_test "," list_for_rest { List (CompList (PI.fake_bracket ($1::$3)), Load)
+                          }
```
Niols: Not sure where that comes from, but should probably be just one line?

topiary-cli/tests/samples/expected/menhir.mly
Niols: Agree. I think And if a user wants groups they can always do something like:
```
%token <AST_python.tok> ADD (* + *) SUB (* - *)
%token <AST_python.tok> MULT "*" (* * *) DIV "/" (* / *)
```
I would otherwise put one keyword per line with maybe some indentation, eg.
```
%token <AST_python.tok>
  ADD (* + *)
  SUB (* - *)
  MULT "*" (* * *)
  DIV "/" (* / *)
```

Niols (approved):
This is. SO. COOL! I think the formatting of Menhir itself is almost right, except for a few nitpicks, so maybe we just fix these and merge. We can then fix things as issues arise.

The big problem imo is that of the `$1` metavariables in the OCaml fragments. The fact that the OCaml looks good in the expected file seems mostly due to the fact that it already looked good in the original file, but Topiary isn't actually doing anything on them.

I would wager that that problem is much wider than Menhir and that many language injections might feature that type of things! In fact, some language injections have way worse, where the injected fragments might contain arbitrary fragments of the outer language. (I'm thinking of LilyPond which can include Scheme fragments which themselves can include some of LilyPond's markup language, but there must be other examples.) I can imagine that this requires much more work. In this specific case, for instance, we might need an extended OCaml grammar that supports `$` variables pretty much anywhere such that we can then format those fragments.

I think that the above is a fair limitation that people will understand. The current PR already brings much value in supporting Menhir files (which, to my knowledge, isn't the case anywhere else), it will format the OCaml at the beginning and end of the file and the fragment without meta-variables.

Niols:
https://discuss.ocaml.org/t/ann-menhirformat-0-1-0/18322

6 days...

aspiwack:
> The big problem imo is that of the `$1` metavariables in the OCaml fragments. The fact that the OCaml looks good in the expected file seems mostly due to the fact that it already looked good in the original file, but Topiary isn't actually doing anything on them.

The good news is that you shouldn't write modern Menhir with `$1` metavariables, instead you're encouraged to use the `x=<production>` form in patterns, then actions are regular Ocaml. So it should be fine no?

What I'm worried about is the health of my own brain: I hadn't noticed the example file was in this style.

ErinvanderVeen:
Questions. Should Topiary format
```menhir
fstring_print_spec: fstring_format_clause { $1 }
| "=" fstring_format_clause { mk_str $1 :: $2 }
```

as

1.A)
```menhir
fstring_print_spec:
| fstring_format_clause { $1 }
| "=" fstring_format_clause { mk_str $1 :: $2 }
```

1.B)
```menhir
fstring_print_spec:
  | fstring_format_clause { $1 }
  | "=" fstring_format_clause { mk_str $1 :: $2 }
```

And what should Topiary do for single lines?

```menhir
fstring_print_spec: fstring_format_clause { $1 }
```

2.A)
```menhir
fstring_print_spec: fstring_format_clause { $1 }
```

2.B)
```menhir
fstring_print_spec:
| fstring_format_clause { $1 }
```

2.C)
```menhir
fstring_print_spec:
  | fstring_format_clause { $1 }
```

Niols:
> The good news is that you shouldn't write modern Menhir with `$1` metavariables, instead you're encouraged to use the `x=<production>` form in patterns, then actions are regular Ocaml. So it should be fine no?

I'd say so, but [the very first comment in the announcement of menhirformat](https://discuss.ocaml.org/t/ann-menhirformat-0-1-0/18322/2) asks

> How do you make ocamlformat handle special dollar vraiable (`$1`, `$startpos(..)`)

so it does seem to be something that people care about.

Niols:
As for your question, @ErinvanderVeen, I think the classic would be no indentation, so 1.A, but I think that's partly just because there is no tooling to handle indentation automatically. Personally, I'd imagine indentation to slightly increase readability and I'd suggest to use 1.B. It's also probably easier to move back from 1.B to 1.A later on than the other way around.

For the second question, shouldn't this be a case where the use has a single- vs. multi-line choice? So 2.A _and_ 2.B/2.C (depending on the indentation question above)?

aspiwack:
> I'd say so, but [the very first comment in the announcement of menhirformat](https://discuss.ocaml.org/t/ann-menhirformat-0-1-0/18322/2) asks

Fair. But still, since this isn't the recommended syntax, I think being dumb with it is doubly reasonable.

> Personally, I'd imagine indentation to slightly increase readability and I'd suggest to use 1.B. It's also probably easier to move back from 1.B to 1.A later on than the other way around.

I was going to say that the  `| ` prefix already count as indentation, and felt sufficient. BUT

There appears to be a new (recommended) syntax for productions (we should add examples of that new syntax in the test file)

Instead of

```menhir
foo:
  | bar { … }
  | baz { … }
```

We are now encouraged to write:

```menhir
let foo :=
  | bar { … }
  | baz { … }
```

And the manual is decidedly writes it with indentation. I think following the manual is the reasonable way to go about it. https://gallium.inria.fr/~fpottier/menhir/manual.html#sec5 .


So I'm thinking 1.B too.

And 2.A for single-lines. This is the most consistent.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,1005 @@
...
+(* see the extra token below NEWLINE instead of TCommentNewline *)
+
+(*-----------------------------------------*)
+(* The normal tokens *)
+(*-----------------------------------------*)
+
+(* tokens with "values" *)
+%token <string * AST_python.tok> NAME
+%token <int option    * AST_python.tok> INT LONGINT
+%token <float option  * AST_python.tok> FLOAT
```
Niols: Not sure what this spacing is?

Niols: I think what's between `<>` is an OCaml type so potentially we could call our injection there but that's probably a bit overkill? I don't know.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,1005 @@
...
+(* layout *)
+%token <AST_python.tok> INDENT DEDENT
+%token <AST_python.tok> NEWLINE
+
+(*************************************************************************)
+(* Rules type declaration *)
+(*************************************************************************)
+%start <AST_python.program> main
+%start <AST_python.any> sgrep_spatch_pattern
+%type <int> dummy_rule %%
```
Niols: I think  that the `%%` should be on its own line, ideally with an empty line before and after.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,1005 @@
...
+
+(* support mixed python2 / python3 comp_for only when used inside atom_list *)
+listcomp_for:
+  | listsync_comp_for { $1 }
+  | ASYNC listsync_comp_for { (* TODO *) $2 }
+
+list_for:
+  or_test "," list_for_rest {
+    List (CompList (PI.fake_bracket ($1 :: $3)), Load)
+  }
```
Niols, line 998: Here I'm wondering if we shouldn't enforce the presence of a `|` as soon as we're not on the same line as the rule's name, for the sake of diff minimisation.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,1005 @@
...
+
+  (* sgrep-ext: difficult to move in atom without s/r conflict so restricted
+     * to argument for now *)
+  | NAME ":" test { Flag_parsing.sgrep_guard (Arg (TypedMetavar ($1, $2, $3))) }
+  | test "=" test {
+    match $1 with
+    | Name (id, _, _) -> ArgKwd (id, $3)
+    | _ -> raise Parsing.Parse_error
+  }
+// This is a line comment for coverage
```
Niols, line 1052: Shouldn't we have added a newline between these two?

Niols (commented):
Very cool! A few nitpicks. Following on @aspiwack's comments, I think we should have more tests, because apparently we're only exerting the paths pertaining to the old style of grammars. Maybe by collecting snippets from the manual into on giant file, the way we do it with OCaml as well?

topiary-cli/tests/samples/expected/menhir.mly
aspiwack: Considering that we do enforce such initial `|` in multiline `match` expression in Ocaml, I think we should enforce it here too.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -0,0 +1,1091 @@
...
+
+fstring_print_spec: fstring_format_clause { $1 }
+  | "=" fstring_format_clause { mk_str $1 :: $2 }
+
+fstring_format_clause:
+  | (*empty*) { [] }
+  | fstring_format_delimeter format_specifier { mk_str $1 :: $2 }
+
+fstring_format_delimeter: ":" { $1 }
+  | BANG { $1 }
```
aspiwack: I guess there's no one answer here. But I'd have expected
```suggestion
fstring_format_delimeter:
  | ":" { $1 }
  | BANG { $1 }
```

aspiwack (commented):
I have one comment about style. Aside from this one comment, it looks amazing.

topiary-cli/tests/samples/expected/menhir.mly
ErinvanderVeen: I agree, and have changed the queries to reflect that.

topiary-cli/tests/samples/expected/menhir.mly
```diff
@@ -1083,9 +1088,11 @@ grouped_producers: AT | AT { () };
...
 empty_action: AT { };
 
-// single-line modern rule: the first alternative stays on the `==` line, so no
-// leading pipe is added (and a stray one would be dropped)
-let single_line_modern == A; { 1 } | B; { 2 }
+// multiple modern alternatives are always vertical, even when written inline;
+// a stray leading pipe is normalized along with the layout
+let single_line_modern ==
+  | A; { 1 }
+  | B; { 2 }
```
aspiwack: On the other hand, it would be better if this one stayed on a single line.

Do we not have this ability? If the multi-production thing takes several line, then they start on a new line, but if they're all on the same line as the declaration, then leave on one line?

ErinvanderVeen: I'll have to look into this. Please hold.

ErinvanderVeen: Resolved!

PR #1285, opened 2026-07-07
feat: allow multiple test files per language

ErinvanderVeen:
# Multiple test files per language, and general simplification

## Description
This PR cleans up our testing infra a bit. It adds the possibility to have multiple tests per language. Additionally, it greatly simplifies the addition of new languages.

This commit (https://github.com/topiary/topiary/pull/1285/changes/16ce5486e500694f370c4451133e0ea3b4038d31) shows the simplification pretty clearly.

It works by adding a `build.rs` file that generates the test file by looking at the directory structure of the sample files.

## Checklist

- [x] `CHANGELOG.md` updated


Xophmeister (commented):
This is an interesting approach 🙂 This will certainly work, but it gives me a weird sense of unease that I can't elucidate properly. I did a bit of research and maybe [`datatest_stable`](https://docs.rs/datatest-stable/latest/datatest_stable/) would be a better option for this kind of use case?

If not, should `generated_sample_tests.rs` at least be in the `.gitignore`?

mkatychev:
related https://github.com/topiary/topiary/issues/1142

topiary-cli/build.rs
```diff
@@ -0,0 +1,74 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("generated_sample_tests.rs");
+
+    // Watch the samples directory for changes
+    println!("cargo:rerun-if-changed=tests/samples/");
```
mkatychev, line 10: would this run on any `cargo install topiary-cli` call?

ErinvanderVeen:
> This is an interesting approach 🙂 This will certainly work, but it gives me a weird sense of unease that I can't elucidate properly. I did a bit of research and maybe [`datatest_stable`](https://docs.rs/datatest-stable/latest/datatest_stable/) would be a better option for this kind of use case?
> 
> If not, should `generated_sample_tests.rs` at least be in the `.gitignore`?

1. No, because that is still runtime. Our issue is that we have every language behind a feature flag.
2. No, `generated_sample_tests.rs` goes to `OUT_DIR` which is typically `/target` and is thus already in our `.gitignore`.

topiary-cli/build.rs
ErinvanderVeen: Yes! It would.

mkatychev: is that desirable? we're essentially adding `preBuildPhase` outside of a nix context

ErinvanderVeen: No it isn't desirable so I'm open to alternatives, but this is just scanning ~15 files, that'll take microseconds. Nothing compared to having to compile Nickel :)

mkatychev: Totally agree that performance wise this is not an impact, I'm concerned of scenarios in the future where a disjoint might happen that would make installs fails on the build, could we gate this logic behind a `cfg(test)` for now to make sure this doesn't touch `cargo install topiary-cli` flows?

Ideally we should reuse this logic for the upcoming `topiary test` subcommand.

ErinvanderVeen: We cannot use `cfg(test)` in build.rs.

I think we should have a discussion on `topiary test`, because in my mind it will not  reuse any of this code.

mkatychev: > I think we should have a discussion on `topiary test`

Sure, I made a PR here (from the issue https://github.com/topiary/topiary/issues/1142):
https://github.com/topiary/bud/pull/10

> because in my mind it will not reuse any of this code.

Perhaps not but it should allow us to split up tests into multiple files without a `build.rs`, right? At least the intent is the same.

PR #1286, opened 2026-07-10
Add declarative Nix configuration module for Topiary

bglgwyng:
References https://github.com/topiary/bud/discussions/5

## Description

This PR adds a declarative Nix module for configuring Topiary.

The module lets users describe Topiary languages in Nix, including file extensions, grammar sources, grammar packages, and formatting query files. It then generates a `languages.ncl` configuration with grammar sources normalized to Nix store paths.

The module is exposed as a flake output under `homeManagerModules.default` and `homeManagerModules.topiary`, so it can be tested through Home Manager before preparing a possible upstream Home Manager PR.

This PR also adds an assertion that a custom language cannot specify both `grammar.package` and `grammar.source.git` at the same time.

[Sample Home Manager configuration](https://gist.github.com/bglgwyng/c777a47c1f32326821dc0101bf799d9a)

## Checklist

- [ ] `CHANGELOG.md` updated
- [ ] ~~Updated regression tests~~  (not applicable; no formatter behavior changes)


nix/languages.nix
mkatychev, line 1: is this file auto-generated or are we moving the source of truth away from nickel and towards nix files?

Niols:
Hi @bglgwyng; welcome!

Thank you very much for this — a declarative Home-Manager module would be a great addition to Topiary! My main concern is that the current shape seems to duplicate a lot of already existing pieces, which also would increase future maintenance burden. There might very well be good reasons for this that I haven't seen, so please just let me know :-)

I also realise this is more a review of https://github.com/topiary/bud/discussions/5#discussioncomment-16902980 by @ErinvanderVeen and that I don't spend any time on Topiary's Nix — so take this as an outside opinion rather than a blocker.

Concretely:

### 1. `nix/languages.nix` duplicates `languages.ncl`

It re-lists every language's extensions/indent/grammar git rev+nixHash — the same values already in `topiary-config/languages.ncl`. We already derive this from the .ncl at `nix/lib/config.nix`: `defaultConfig = fromNickelFile ../../topiary-config/languages.ncl`, with `defaultConfigPrefetched` alongside. So `nix/languages.nix` is a hand-maintained copy that has to be updated (and its hashes kept in sync) on every grammar bump. Could we use `defaultConfig` / `defaultConfigPrefetched` and drop the file?

### 2. The module mirrors Topiary's config schema as bespoke options

`nix/modules/topiary.nix` declares an option per field (`extensions`, `indent`, `grammar.symbol`, `grammar.source.git.{url,rev,hash,subdir}`, …). This pattern is discouraged nowadays in nixpkgs (see [RFC-42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)) precisely for maintainability reasons. To quote the RFC above:

> Traditionally this has lead to huge modules with dozens of options, each of them only for a single field in the configuration. Such modules are problematic because they're hard to write, review and maintain, are generally of lower quality, fill the option listing with noise and more.

Of course, Topiary is fairly small, but still. RFC 42 style would be a single freeform settings option (`submodule { freeformType = fmt.type; }`) serialised to Topiary's native config by a format `generate`, with no per-field options and no `nix/languages.nix`. Two serialisers both work here:

- Nickel via the existing `generateNcl`/`toNickelValue` — keeps Nickel as the single config language, consistent with languages.ncl and its contracts, but more code.
- JSON via `pkgs.formats.json {}` — the canonical RFC 42 pattern and the least code; already proven since `wrapWithConfig` (`nix/lib/config.nix`) feeds `topiary -C` a `toJSONFile` result today. Trade-off is the emitted config is no longer Nickel.

Two inputs do warrant explicit options (RFC 42's carve-out for values needing eval-time Nix handling): `grammar.package` (derivation → store path) and `query.formatting` (path installed into `TOPIARY_LANGUAGE_DIR`). So the natural shape is a hybrid — freeform settings for the config proper, plus those few Nix-level options — rather than a full schema mirror.

### 3. Duplicated pipeline + a second wrapper

`normaliseGrammar`/`normaliseLanguage` and the whole import → merge with defaults → `generateNcl` → `prefetchLanguages` pipeline seem to be duplicated in `nix/modules/home-manager.nix` and `nix/packages/topiary.nix`; worth sharing one implementation. Also, `normaliseGrammar` only exists because the module invents a different grammar shape (`source.git.{url,rev,hash}` vs `source.git.{git,rev,nixHash}`) — aligning with the Nickel/`prefetchLanguages` shape removes the adapter and brings more consistency. Finally, we already have `wrapWithConfig`/`wrapWithConfigFile` (`nix/lib/config.nix`); the new `makeWrapper --set-default TOPIARY_CONFIG_FILE` path seems to duplicate that job?

### Conclusion

I feel that if we (a) reused `defaultConfig(Prefetched)` for defaults, (b) switched the schema to a freeform `settings` serialised by either `generateNcl` or `pkgs.formats.json` per RFC 42, keeping explicit options only for `grammar.package` and `query.formatting`, and (c) de-duplicated the pipeline and reused `wrapWithConfig` — then the PR would shrink to the parts that are actually new (the module option surface, query-file install, prebuilt `grammar.package` support).

nix/lib/eval-config.nix
```diff
@@ -0,0 +1,65 @@
...
+
+{ cfg }:
+
+let
+  inherit (lib) optionalAttrs;
+  inherit (lib.attrsets) mapAttrs filterAttrs mapAttrs' nameValuePair mapAttrsToList;
+  inherit (lib.strings) optionalString concatStringsSep;
+
+  normaliseGrammar = name: g:
+    assert !(g.package != null && g.source.git != null) || throw "topiary: language `${name}` cannot specify both `grammar.package` and `grammar.source.git`";
```
mkatychev: @ErinvanderVeen is there any way we can rely on the `GrammarSource` contract for this?
https://github.com/topiary/topiary/blob/1a84ded72669c74f08a94103d8ddf4c814f2c4d1/topiary-config/languages.ncl#L43-L54

ErinvanderVeen: Possible yes. But it comes with trade-offs. Any Nix asserts are evaluated at Nix eval. I.e. when doing `home-manager switch`. Any Nickel contract will only be evaluated at Topiary runtime. I.e. `topiary fmt`.

An option is to eval Nickel as an assert during Nix eval. But I believe that introduces more complexity than it is worth.

Niols: The canonical way would be to generate in as little Nix as possible but call Topiary in the derivation to check the configuration. In our case, we could even call Topiary to check and format the config!

This keeps the logic of the Nix code minimal and scalable while still having Topiary as the source of truth. If one decides to use a newer/older/patched version of Topiary, it still just works.

ErinvanderVeen: Done!

nix/lib/eval-config.nix
```diff
@@ -0,0 +1,69 @@
...
+    # Loading the configuration is enough to apply its Nickel contracts.
+    ${getExe cfg.package} --configuration ${generatedConfigFile} config show-sources > /dev/null
+    cp ${generatedConfigFile} $out
+  '';
+
+  customQueries = filterAttrs (_: l: l.query.formatting or null != null) (cfg.settings.languages or { });
+  hasCustomQueries = customQueries != { };
+
+  queryFiles = mapAttrs' (
+    name: l: nameValuePair "topiary/queries/${name}/formatting.scm" { source = l.query.formatting; }
```
mkatychev, line 51: I'm trying to add a query per language setup in https://github.com/topiary/topiary/pull/1299, do you think that might complicate things 😅?
https://github.com/topiary/topiary/blob/dd62c53cbe95dbe9a2a5274b753e24a74d4a549c/topiary-config/languages.ncl#L77-L95

mkatychev: The way I have it means that we don't have the same baked in assumptions:
https://github.com/topiary/topiary/blob/26245e03a7416c6cd87dd2bc7ac6834c702043eb/topiary-config/languages.ncl#L160-L173

Issue #839, opened 2025-01-17
Add support for a neovim-style `extends` query hierarchy.


ctdunc:
**Is your feature request related to a problem? Please describe.**
I want to allow users to pass settings like "do/don't capitalize keywords" which would currently require me to maintain multiple query files where the content is largely the same.

**Describe the solution you'd like**
I would like to have some order in which multiple query files are loaded from different locations. e.g. the default ships at `/usr/share/topiary/queries/lang.scm`, but users can override it at `.config/topiary/queries/lang.scm`. 

neovim added an `;extends` modeline to their tree-sitter queries, which when present indicates that neovim should just append the text of the file to any previously loaded queries. See [their very simple implementation](https://github.com/neovim/neovim/blob/b9e6fa7ec81c463d77cc919392b52f6df2d8d304/runtime/lua/vim/treesitter/query.lua#L235)

This works fine. In my SQL example, I tried out the query file (all the whitespace stuff omitted)

```scheme
([
  (keyword_select) 
  (keyword_from)
  (keyword_where)
]) @upper_case
(keyword_select) @lower_case
```

which produces the following output:
```
select id
  ,column_a
  ,column_b
FROM foo
WHERE id = 0
;
```
from the input 
```
sElEcT id, column_a, column_b FRoM foo WHeRE id = 0;
```

**Describe alternatives you've considered**
Shipping a grammar file for every possible permutation of settings a user could want for SQL. This is :thumbsdown:

Xophmeister:
Ultimately, this feature is about formatting configurability. FYI: A related discussion has just been opened, which can be followed here:

* #847

mkatychev:
@ctdunc this issue has been discussed a bit, ultimately https://github.com/topiary/topiary/issues/839 means to open this up to be handled by passing in a list for path rather than a single file, this saves us from adding directives into files that don't necessarily support them and make inheritance explicit:

```nickel
   queries = { 
     formatting.source = { 
       git = { 
         git = "https://github.com/topiary/topiary.git", 
         rev = "d2c79b9ecd341d40aa0baf87f4a761ae242dfa67", 
       }, 
       path = [ 'Git "topiary-queries/queries/markdown/formatting.scm", "~/my-local-query.scm"]
     }, 
   } 
 }, 
```

ctdunc:
That seems like a reasonable compromise. Especially with all the nvim treesitter drama 😶

[...]

PR #1299, opened 2026-07-21
feat(config,queries): add `QuerySource` config fetching to topiary-config

mkatychev:
# add `QuerySource` config fetching to topiary-config

Users can now provide injections and formattesr for any language using git sources.

References https://github.com/topiary/topiary/issues/1003 https://github.com/topiary/topiary/issues/839

## Description
This PR allows us to fetch queries on a per type and language basis from `languages.ncl`:
https://github.com/topiary/topiary/blob/26245e03a7416c6cd87dd2bc7ac6834c702043eb/topiary-config/languages.ncl#L160-L173

Each unique git source is now shallow-cloned once per Topiary run and reused
across every matching `GitSource`, this means sourcing
queries and grammars from the same repository+rev results in a single `git fetch`.

* added `topiary config --field languages.<my-language>`

## Checklist

- [x] `CHANGELOG.md` updated


topiary-cli/src/language.rs
```diff
@@ -94,11 +96,23 @@ impl LanguageDefinitionCache {
         name: &str,
     ) -> CLIResult<Arc<Language>> {
         let config_language = config.get_language(name).preformat_context()?;
-        let formatting_query = to_query_from_language(config_language)?;
-        let injection_query = to_injection_query_from_language(config_language);
+        let formatting_query = to_query_from_language(
+            config_language,
+            topiary_queries::FORMATTING_QUERY,
+            Some(&self.repos),
```
mkatychev, line 102: I wonder if it makes sense to do something similar to this:
https://github.com/rust-lang/cargo/blob/877ff64d694ae967badcf9b6a629619e6fe5e0db/src/cargo/core/source_id.rs#L23-L23

```rust
static LANGUAGE_CACHE: OnceLock<Arc<LanguageDefinitionCache>> = OnceLock::new();
```

topiary-config/languages.ncl
```diff
@@ -132,6 +157,19 @@ in
...
+            git = "https://github.com/topiary/topiary.git",
+            rev = "d2c79b9ecd341d40aa0baf87f4a761ae242dfa67",
+          },
+          path = "topiary-queries/queries/markdown/formatting.scm"
+        },
+        injections.source = {
+          git = formatting.source.git,
+          path = "topiary-queries/queries/markdown/injections.scm",
+        },
+      }
```
mkatychev: need to remove, this was for testing

topiary-cli/src/cli.rs
```diff
@@ -154,6 +154,15 @@ pub enum Commands {
     /// Print the current configuration
     #[command(alias = "cfg", display_order = 3)]
     Config {
+        /// Return a specific field path of the configuration.
+        /// Ex: `topiary config --field languages.json`.
+        //
+        // NOTE: The path follows Nickel's field-path syntax: identifiers with
+        // with dashes must be enclosed in double quotes:
+        // `languages."tree_sitter_query".extensions`.
```
yannham: The text mentions dashes, but the example uses underscores. In fact you don't have to quote any of them, as they are valid identifiers in Nickel. Only a dash at the start of a identifier somewhere in the dot-separated list has to be quoted, like `foo.-bar` and otherwise non-identifier characters (spaces, special characters, etc.)

topiary-cli/src/error.rs
```diff
@@ -194,7 +197,7 @@ pub(crate) trait ResultPreformat<T, C> {
 
 impl<T, C: 'static> ResultPreformat<T, C> for Result<T, C>
 where
-    C: 'static,
+    C: 'static + std::fmt::Debug,
```
yannham:
```suggestion
    C: 'static + fmt::Debug,
```

topiary-config/languages.ncl
```diff
@@ -59,6 +59,22 @@ variant with the other via Nickel record merging (e.g. setting
       | optional
       | doc "Override for the tree-sitter symbol; defaults to `tree_sitter_<name>`.",
   },
+  QuerySource
+    | doc m%"
+A query file location. Always requires `path`; optionally accepts `git` to
+fetch the file from a remote repository (in which case `path` is resolved
+relative to the checkout root).
+"%
```
yannham:
```suggestion
    | doc m%"
      A query file location. Always requires `path`; optionally accepts `git` to
      fetch the file from a remote repository (in which case `path` is resolved
      relative to the checkout root).
    "%
```

yannham (commented):
Went fast on the Rust part, focused on Nickel specifics.

topiary-cli/src/cli.rs
mkatychev: Oh that's great to hear!

yannham: (sorry I meant they are valid characters _within_ an identifier, except a starting dash. A dash or an underscore alone isn't a valid identifier)

docs/book/src/cli/configuration.md
```diff
@@ -187,6 +187,58 @@ nickel = {
...
+    },
+    injections.source = {
+      git = formatting.source.git,
+      path = "topiary-queries/queries/markdown/injections.scm",
+    },
+  },
+},
+```
+
+Each unique git source is shallow-cloned once per Topiary run and reused
```
Xophmeister: Nitpick

```suggestion
Each unique Git source is shallow-cloned once per Topiary run and reused
```

docs/book/src/cli/usage/config.md
```diff
@@ -12,6 +12,7 @@ Commands:
   help          Print this message or the help of the given subcommand(s)
 
 Options:
+  -f, --field <FIELD_PATH>             Return a specific field path of the configuration. Ex: `topiary config --field languages.json`
```
Xophmeister: This is the first file in the PR and I don't understand what this does. After reading the configuration documentation and the CLI help text (the next files in the PR), I still don't really know what this does.

I assume I'll understand what this does once I've worked my way through the code, but I think this needs more explanation for users.

topiary-cli/src/cli.rs
```diff
@@ -154,6 +154,15 @@ pub enum Commands {
     /// Print the current configuration
     #[command(alias = "cfg", display_order = 3)]
     Config {
+        /// Return a specific field path of the configuration.
+        /// Ex: `topiary config --field languages.json`.
```
Xophmeister: Besides my previous comment... Nitpick:

```suggestion
        /// e.g., `topiary config --field languages.json`.
```

topiary-cli/src/io.rs
```diff
@@ -2,11 +2,15 @@ use std::{
     ffi::OsString,
     fmt::{self, Display},
     fs::File,
-    io::{self, BufWriter, Read, Seek, Write},
+    io::{self, Read, Seek, Write},
     path::{Path, PathBuf},
     sync::Arc,
 };
 
+#[cfg(feature = "nickel")]
```
Xophmeister: Isn't the `nickel` feature gate for Nickel formatting support, rather than including/excluding Nickel altogether? I may be misunderstanding without having read more of the code yet, but it surprises me to see this feature gate being used in `topiary-cli/src/io.rs`.

topiary-config/src/language.rs
```diff
@@ -363,3 +490,144 @@ fn source_object_id(source: &refmap::Source) -> Result<ObjectId> {
...
+        git = "https://github.com/topiary/topiary.git",
+        rev = "d2c79b9ecd341d40aa0baf87f4a761ae242dfa67",
+      },
+      path = "topiary-queries/queries/markdown/formatting.scm"
+    },
+    injections.source = {
+      git = formatting.source.git,
+      path = "topiary-queries/queries/markdown/injections.scm",
+    },
+  }
```
Xophmeister, line 550: Just an observation.

I'm in two minds about the structural difference between grammar Git sources and query Git sources; namely `grammar.source.git.subdir` (inside the `git` record) vs. `queries.*.source.path` (outside the `git` record).

On the plus side, it does mean you can do back referencing, which is useful. That said, you could always do something like:

```nickel
injections.source.git = formatting.source.git & { path = "different/path/to/query.scm" }
```

If the `subdir`/`path` inconsistency were also collapsed, you could presumably even do:

```nickel
queries = {
  formatting.source.git = grammar.source.git & { path = "different/path/to/formatting.scm" },
  injections.source.git = grammar.source.git & { path = "different/path/to/injections.scm" }
}
```

It's not a big deal, but I'm of the opinion that these kinds of little opportunities are what can really polish the [UD]X.

topiary-config/src/language.rs
```diff
@@ -241,38 +345,54 @@ impl<T, E: Into<anyhow::Error>> GitResult<T> for Result<T, E> {
...
-        // Create a temporary directory to clone the repository to. We could
-        // cached the repositories, but the additional disk space is probably
-        // not worth the benefits gained by caching. The tempdir is deleted
-        // when dropped
-        let tmp_dir = tempfile::tempdir()?;
-
-        self.fetch_and_compile_with_dir(name, library_path, false, tmp_dir.keep())
+/// A single shallow checkout of a [`GitSource`] holding a  [`TempDir`], deleted on drop.
+#[derive(Debug)]
+pub struct LocalRepo(TempDir);
```
Xophmeister, line 347: If it gets deleted on drop, then presumably it's written to `$XDG_CACHE_DIR` -- like the grammars -- at some point? Maybe I haven't hit that bit of code yet...

Xophmeister (commented):
I've left some comments, because while I get the gist of what's going on, there are gaps in my understanding. Those very well may be my problem, rather than the content of the PR, of course!

docs/book/src/cli/usage/config.md
Xophmeister: Ooh, it just subsets the merged TOML to the "field" address you give it; presumably to make the output clearer (i.e., you subset it to what you want to see, rather than everything). I was confused because `languages.json` looks like a file path, even when it doesn't make sense in the context of Topiary. Maybe something like `languages.json.grammar.source` would make it a bit clearer.

mkatychev: yes this is primarily done to help debug `topiary config` and to make the e2e tests simpler without asserting values explicity.
`nickel eval --field languages.json.grammar.source` already does this so it was meant to test the query resolution.

topiary-cli/src/io.rs
mkatychev: This was a complaint made by clippy:
https://github.com/topiary/topiary/actions/runs/30235091219/job/89881324393
This is done when calling `-p topiary-cli --no-default-features`. If there's a better way to do this, I am all ears.

topiary-config/src/language.rs
mkatychev: I think the difference here is that all paths are explicit (since there might be one [or more formatting files](https://github.com/topiary/topiary/issues/839#issuecomment-5061461703)) for queries, `subdir` is too ambigous and would get us back into [implicit query/config discovery problems](https://github.com/topiary/bud/discussions/9). So the approaches I thought about involed:
1. ignoring `subdir` in queries (I did this initially but it felt wonky) but required in grammars
2. decoupling `subdir` from `source.git` and having an extrenal reference.
3. handling an xor of `subdir` and `file` (too messy)

topiary-config/src/language.rs
mkatychev: This is the temporary directory where the shallow git clone is kept, this is going to be in `/tmp/....` and copied to cache dir just like the current grammar repoes are doing.

topiary-cli/src/cli.rs
mkatychev: I've decided to remove that section altogether, the leading dash is a rare enough case not work mentioning in the help string, thanks for the clarification!

topiary-config/src/language.rs
mkatychev: @Xophmeister I realise this was a bit of a leap in decision so I'm happy to dial back as needed, this was the shape that emerged after trying to genericise the rust `GitSource` struct.

topiary-cli/src/io.rs
Xophmeister: Ah, OK, fair enough. Maybe `#[allow(unused)]` would be clearer... but I still don't get it. Why would it be unused in the general case?

topiary-config/src/language.rs
Xophmeister: What about renaming `source.git.subdir` to `source.git.path`? Paths can be files or directories, after all.

mkatychev: That's not a bad idea, my main concern was backwards compatibility. This would break most people's personal configs.

topiary-cli/src/io.rs
mkatychev: Agreed, `BufWriter` is called here:
https://github.com/topiary/topiary/blob/6e0ee47bf3547da8bfe0009df05cf0bdbb39d064/topiary-cli/src/io.rs#L665

Perhaps add a new feature flag called `process-config = ["nickel"]`?

topiary-config/src/language.rs
Xophmeister: That is true and it's a very good reason not to go down this route. It would be a lot of work, probably, to have `subdir` aliased to `path` and issue a deprecation warning. :confounded:

mkatychev: Right, that was part of the work on languages.rs:
https://github.com/topiary/topiary/blob/6e0ee47bf3547da8bfe0009df05cf0bdbb39d064/topiary-config/src/language.rs#L224-L231

Xophmeister (approved):
This looks good; thank you 💛

I think it's right that the `nickel` feature gate confusion and the configuration schema don't block this PR. We can talk about these elsewhere :+1:

mkatychev:
@Xophmeister I actually did a bit of clean up after our discussion to hopefully consolidate the configuration stuff a bit:
https://github.com/topiary/topiary/pull/1299/commits/434bcc5940b9b51752a95a5a5d9de6deb984b7e6

we can discuss clarifying the feature flagging eventually. My main takeaways from the commit above to group stuff together:

* newtype `Config`: `struct Config(topiary_config::Config, NickelValue);`
  * this means we can feature gate the formatting and metadata stripping functionality inside `impl topiary::Config {` 
* use `fancy-config = ["nickel"]` to clarify why the nickel grammar and queries are used.

PR #1303, opened 2026-07-27
 reafactor(ci,doc): Add `bin/verify-documented-usage.sh --no-nix-shell` flag

mkatychev:
# Add `bin/verify-documented-usage.sh --no-nix-shell` flag

## Description

Added `--no-nix-shell` to opt out of rebuilding the project to verify the documentation

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/bin/verify-documented-usage.sh
+++ b/bin/verify-documented-usage.sh
@@ -31,7 +31,7 @@ get-cli-usage() {
 
   case "${subcommand}" in
     "index") "${TOPIARY}" --help;;
-    *) "${TOPIARY}" "${subcommand}" --help;;
+    *)       "${TOPIARY}" "${subcommand}" --help;;
   esac
 }
 
@@ -53,7 +53,11 @@ diff-usage() {
   # Generate a diff between the README and CLI help text
   local subcommand="$1"
 
-  diff --text --ignore-all-space --side-by-side <(get-documented-usage "${subcommand}") <(get-cli-usage "${subcommand}")
+  diff --text \
+       --ignore-all-space \
+       --side-by-side \
+       <(get-documented-usage "${subcommand}") \
+       <(get-cli-usage "${subcommand}")
 }
 
 main() {
--- a/bin/verify-documented-usage.sh
+++ b/bin/verify-documented-usage.sh
@@ -18,20 +18,14 @@ export __TOPIARY_TERM_WIDTH=90
 
 readonly FENCE='```'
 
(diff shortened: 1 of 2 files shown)


bin/verify-documented-usage.sh
```diff
@@ -1,21 +1,37 @@
...
 get-cli-usage() {
   # Get the help text from the CLI
   local subcommand="$1"
-  local topiary="${TOPIARY:-cargo run -q --bin topiary --}"
 
   case "${subcommand}" in
-    "index") ${topiary} --help;;
-    *)       ${topiary} "${subcommand}" --help;;
+    "index") "${TOPIARY}" --help;;
+    *) "${TOPIARY}" "${subcommand}" --help;;
```
Xophmeister: The formatting broke here...

bin/verify-documented-usage.sh
```diff
@@ -37,11 +53,7 @@ diff-usage() {
   # Generate a diff between the README and CLI help text
   local subcommand="$1"
 
-  diff --text \
-       --ignore-all-space \
-       --side-by-side \
-       <(get-documented-usage "${subcommand}") \
-       <(get-cli-usage "${subcommand}")
+  diff --text --ignore-all-space --side-by-side <(get-documented-usage "${subcommand}") <(get-cli-usage "${subcommand}")
```
Xophmeister: ...and here.

bin/verify-documented-usage.sh
mkatychev: Whoops, I called `topiary format` on it 😅

bin/verify-documented-usage.sh
mkatychev: Same comment here: https://github.com/topiary/topiary/pull/1303#discussion_r3661122305

Happy to revert.

bin/verify-documented-usage.sh
Xophmeister: We don't eat our own dogfood here :see_no_evil:

Xophmeister (approved):
I _think_ this looks good, but I'm not enough of a Nix person to be confident.

I don't think it's going to break anything, so you can merge, but it always worth a second opinion from @ErinvanderVeen

mkatychev:
Also measuring this against master, there's no measurable speedup for compilation so the sum gain might be 0 here (both 20 seconds).

mkatychev:
Given that the performance is identical in CI, I can keep just `--no-nix-shell` in the changelog, I'll wait on @ErinvanderVeen's input

Xophmeister (approved):

PR #1306, opened 2026-07-30
feat(cli): Move topiary-cli configuration logic into its own module

mkatychev:
# Move topiary-cli configuration logic into its own module

Resolves https://github.com/topiary/topiary/pull/1299#issuecomment-5124597328

## Description

* Handle configuration management and retrieval in a separate module called `config` for `topiary-cli`
* Add feature flag called `fancy-config` based on nickel feature flag discussions with @Xophmeister 

## Checklist

Checklist before merging, wherever relevant:

- [x] `CHANGELOG.md` updated

Issue #605, opened 2023-08-24
Fix the atomicity of the language definition cache 


Xophmeister:
**Describe the bug**
The language definition cache (introduced in #583) is designed to cache the necessary components for the Topiary API for a given input file type. This has been implemented to be thread-safe, as it is used concurrently, with a `RwLock`: First the read lock is acquired, to retrieve the cache contents if it exists; otherwise, the read lock is released and a write lock is acquired to update the cache and return the appropriate value.

This works and is thread-safe, _but_ it doesn't exhibit the desired behaviour. When multiple consumers of the cache (in our case, async futures) query the cache more-or-less simultaneously, they (almost) all bypass the read branch of the code -- because the cache is empty -- and hit the write branch. This flaunts the point of the cache to reduce writes (which involves disk IO to produce the value).

**To Reproduce**
1. Create a directory named, for example, `test`. In that directory, create a simple JSON file and reproduce it 200 times:

   ```console
   $ mkdir test
   $ jq -n "{test:123}" > test/test0.json
   $ for i in $(seq 200); do cp test/test0.json test/test${i}.json; done
   ```

2. Format the contents of this directory, with the logging level turned up, and search for `Cache` in the output:

   ```console
   $ RUST_LOG=debug topiary fmt test 2>&1 | grep Cache
   ```

You will see 201 lines that almost all look like this:

    [2023-08-24T09:58:20Z DEBUG topiary::language] Cache 0x7fffffff6af0: Insert at 0xb503fcb641e7c545 (json, json.scm)

You _may_ get lucky and some of the lines will look like this:

    [2023-08-24T09:58:20Z DEBUG topiary::language] Cache 0x7fffffff6af0: Hit at 0xb503fcb641e7c545 (json, json.scm)

**Expected behavior**
In the logging output, we should see the first line look something like this:

    [2023-08-24T09:58:20Z DEBUG topiary::language] Cache 0x7fffffff6af0: Insert at 0xb503fcb641e7c545 (json, json.scm)

Then we should see 200 lines that look like this:

    [2023-08-24T09:58:20Z DEBUG topiary::language] Cache 0x7fffffff6af0: Hit at 0xb503fcb641e7c545 (json, json.scm)

**Environment**
 - OS name + version: NixOS 23.05
 - Version of the code: 004c4c3099eca

**Additional context**
[...]

PR #1325, opened 2026-08-11
Fixed regression in `--query` parsing

### LINKED ISSUE #1324: `topiary format --query` is ignored
```
$ echo '("=") @prepend_hardline' > silly.scm

$ cargo run -- format -s -l nickel -q silly.scm < topiary-cli/tests/samples/input/nickel.ncl | head
+ command cargo run -- format -s -l nickel -q silly.scm
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/topiary format -s -l nickel -q silly.scm`
{
  regression_tests = [
    # Issue 973: Trailing whitespaces are semantically relevant
    {
      a =
        std.string.split
          " "
          m%"
          line one with trailing space 
          line two

 ● I/O Error
 ├ topiary-core/src/error/mod.rs:177
 ├ Could not read or write to file
 ├ broken pipe
 │
 ● Broken pipe (os error 32)
 ╰ topiary-core/src/lib.rs:397


$ git log -1 --oneline
1aadffa6 (HEAD, origin/main, origin/HEAD) fix(cli): Improve error message when Topiary cannot find a query file (#1322)
```

notice that the error message is unrelated to this issue but caused by piping to `head`. it is still interesting though.

Xophmeister:
# Fixed regression in `--query` parsing

Resolves #1324
References #1306

### What broke

`topiary-cli/src/io.rs:425` in `process_inputs`. Commit `60ce420c` ("refactor(cli): fixed compilation") changed how the language definition is resolved for each input:

```rust
// before
let language = cache.fetch_input(&input)?;

// after (broken)
let language_name = input.language().name.clone();
let language = Arc::new(config.get_language(&language_name)?);
```

The two are not equivalent:

- `cache.fetch_input(&input)` to `input.to_language_sync()`, which compiles the queries held on the `InputFile`; i.e., `input.formatting_query`, which `Inputs::new` populates from the `--query` argument when one is given (`io.rs:268-274`).
- `config.get_language(name)` re-resolves the query from scratch by language name (`config.rs:124`), so it always picks the config/built-in query and silently discards the override.

So `--query` was still parsed, still validated, still logged ("Formatting ... using <your query>"), but the formatter received the default query.

This looks to be accidental as:
- `Commands::Visualise` in `main.rs:165` was left calling `config.cache().fetch_input(&input)`, so `--query` still worked there.
- The replacement also bypassed `LanguageDefinitionCache` entirely for the `format`/`check`/`check-grammar` paths, recompiling grammars and queries per input. That's the regression #605 originally guarded against.

This was collateral from making `Configuration` thread-safe (the same commit that introduced the thread-local `NickelValue` storage), not a deliberate change to query resolution.

### The fix

Restored the cache-based resolution, keeping the `attach_filepath` error context the refactor added:

```rust
let language = config
    .cache()
    .fetch_input(&input)
    .attach_filepath(location.to_path())?;
```

This also restores the language-definition caching for multi-input runs.

## Checklist

- [x] `CHANGELOG.md` updated

### DIFF
--- a/topiary-cli/src/io.rs
+++ b/topiary-cli/src/io.rs
@@ -423,12 +423,15 @@ where
                 let input = input?;
                 let location = input.source().location();
                 tokio::task::block_in_place(|| {
-                    let language_name = input.language().name.clone();
-                    let language = Arc::new(
-                        config
-                            .get_language(&language_name)
-                            .attach_filepath(location.to_path())?,
-                    );
+                    // NOTE Resolve the language definition from the input itself, rather
+                    // than by name from the configuration. The input carries the query
+                    // sources that were resolved for it, which may include a user-supplied
+                    // override (i.e. `--query`); going via the configuration would silently
+                    // discard that override.
+                    let language = config
+                        .cache()
+                        .fetch_input(&input)
+                        .attach_filepath(location.to_path())?;
                     process_fn(input, language, config)
                         .map_err(|e| e.attach_filepath(location.to_path()))
                 })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ This name should be decided amongst the team before the release.
 - [#1287](https://github.com/topiary/topiary/pull/1287) [#1296](https://github.com/topiary/topiary/pull/1296) Various OCaml issues and improvements
 - [#1291](https://github.com/topiary/topiary/pull/1291) Improve warning for injection queries without content captures
 - [#1322](https://github.com/topiary/topiary/pull/1322) Improve error message when Topiary cannot find a query file
+- [#1325](https://github.com/topiary/topiary/pull/1325) Fix regression with `--query` value parsing in `topiary format`
 
 ### Added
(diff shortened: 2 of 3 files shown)


topiary-cli/src/io.rs
```diff
@@ -423,12 +423,15 @@ where
...
-                    let language_name = input.language().name.clone();
-                    let language = Arc::new(
-                        config
-                            .get_language(&language_name)
-                            .attach_filepath(location.to_path())?,
-                    );
+                    // NOTE Resolve the language definition from the input itself, rather
+                    // than by name from the configuration. The input carries the query
+                    // sources that were resolved for it, which may include a user-supplied
+                    // override (i.e. `--query`); going via the configuration would silently
```
mkatychev, line 429: Nice catch, this was completely my oversight 🤦

mkatychev (approved):
LGTM, thanks for tracking this down.

topiary-cli/src/io.rs
Xophmeister: I can't take the credit. I did the bisect, but my handy robot friend made the fix!

PR #1328, opened 2026-08-13
fix!(cli,config): resolve relative filepaths in configuration

mkatychev:
# [PR Name]

Resolves #XXX
References #YYY
Depends on #ZZZ

## Description

[PR Description]

## Checklist

Checklist before merging, wherever relevant:

- [ ] `CHANGELOG.md` updated
- [ ] Documentation (The Topiary Book, `README.md`, etc.) up-to-date

- [ ] Updated regression tests

PR #1335, opened 2026-09-08
fix!(cli,config): Resolve relative filepaths in configuration

Xophmeister:
# Resolve relative filepaths in configuration

Stacked upon #1328

Assisted by `claude-opus-5[1m]`

## Description

Completes #1328, which put the plumbing in place -- `topiary-config` handing back a `Program`, so Nickel's position table outlives evaluation -- but left the resolution itself as a `todo!()`.

A relative `grammar.source.path` or `queries.<name>.source.path` went straight to `std::fs`, and so resolved against the working directory: a `.topiary/languages.ncl` found by ascending the directory tree only worked from the one directory it happened to be written for. Such paths are now anchored at the `.ncl` file that defined them, as Nickel already does for its own `import`s.

Resolution runs on the *evaluated* configuration, just before deserialisation.  By then Nickel has done all merging, contract application and `| default` overriding, so each surviving string literal is the one that won, and its `PosIdx` names the file it came from. Doing it in Nickel instead was rejected: mapping over a record discards field metadata, which the built-in configuration leans on throughout.

The new `topiary-config::paths` walks the two field paths that name a local file -- `languages.<lang>.grammar.source` and `languages.<lang>.queries.<name>.source` -- rather than matching any record carrying a `path`, which would catch unrelated user fields (both contracts are non-exhaustive). A sibling `git` field is the discriminator: that path lives inside the fetched checkout, so it is left as written. So is the built-in configuration, which has no file to anchor against.

Also in here:

- `cfg --field` now navigates the resolved configuration rather than handing the path to Nickel as `Program::field`, so it reports what the rest of Topiary uses. A missing field gets a readable message instead of a `Debug` dump of Nickel's `EvalError`.
- Nickel's evaluation warnings went to `NullReporter` and were discarded; they now reach the user through `log::warn!` as full codespan diagnostics.
- #1328's exploratory scaffolding is removed, and the root `languages.ncl` symlink -- overwritten with a scratch file -- is restored.

Nine new tests. The load-bearing one is `each_path_is_anchored_at_its_own_file`: an entrypoint configuration `import`s a second file, and each path resolves against its own directory, confirming that positions survive Nickel's import resolution and merging. Both relative-path CLI tests were checked to fail without the resolver.

## Checklist

- [x] `CHANGELOG.md` updated


topiary-config/src/lib.rs
```diff
@@ -315,23 +308,25 @@ pub(crate) fn project_dirs() -> directories::ProjectDirs {
 
 pub struct Program {
     inner: nickel_lang_core::program::Program<CBNCache>,
+    /// Memoised result of [`Self::eval_config`]. Evaluation is deterministic and the
+    /// program is never mutated after construction, so it is only ever done once.
+    config: Option<NickelValue>,
```
mkatychev, line 313: That's a nice addition

topiary-cli/src/config.rs
```diff
@@ -61,8 +60,9 @@ impl Configuration {
         })
     }
 
+    /// Get the [`Program`] that this configuration was evaluated from.
     ///
-    /// Returns an [`Rc<NickelValue>`] which allows cheap cloning via reference counting.
+    /// Returns an [`Rc`] which allows cheap cloning via reference counting.
```
mkatychev, line 65:
```suggestion
    /// Returns an [`Rc<RefCell<Program>>`] which allows cheap cloning via reference counting.
```

topiary-config/src/lib.rs
```diff
@@ -315,23 +308,25 @@ pub(crate) fn project_dirs() -> directories::ProjectDirs {
...
+        Self {
+            inner: program,
+            config: None,
+        }
     }
 }
 
 impl Program {
-    fn builder() -> ProgramBuilder<NullReporter, std::io::Stderr> {
+    fn builder() -> ProgramBuilder<LogReporter, std::io::Stderr> {
```
mkatychev, line 326: This is a nice change I didn't have time to do.

PR #1337, opened 2026-09-09
Added an AI policy to the PR template

Xophmeister:
The PR template isn't the best tool for this, but I don't think we need to bother with an automated check.

I've made this a draft PR, for now, because I don't want to impose what I think the AI policy should be. Please use this PR for discussion.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,6 +20,16 @@ Resolves #XXX
 References #YYY
 Depends on #ZZZ
 
+<!----------------------------------------------------------------------
+AI policy: If this PR was assisted by an AI model, you MUST include the
+following section, naming the model used. AI usage is permitted, but by
+making a PR, you assert that you have both reviewed and verified the
+output and can fully explain and justify the changes made.
+
+Remove this section if no AI assistance was used.
+----------------------------------------------------------------------->
+Assisted by `ai-model`
+
 ## Description
 
 [PR Description]
@@ -36,9 +46,9 @@ Checklist before merging, wherever relevant:
 - [ ] `CHANGELOG.md` updated
 - [ ] Documentation (The Topiary Book, `README.md`, etc.) up-to-date
 <!----------------------------------------------------------------------
-If the PR solves a formatting issue for a supported language,
-or generally improves formatting, please make sure to include a
-regression test showcasing the fix/improvement to:
+If the PR solves a formatting issue for a supported language, or
+generally improves formatting, please make sure to include a regression
+test showcasing the fix/improvement to:
 `topiary-cli/tests/samples/{input,expected}/mylanguage.lang`
 ----------------------------------------------------------------------->
 - [ ] Updated regression tests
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -20,6 +20,16 @@ Resolves #XXX
...
 
+<!----------------------------------------------------------------------
+AI policy: If this PR was assisted by an AI model, you MUST include the
+following section, naming the model used. AI usage is permitted, but by
+making a PR, you assert that you have both reviewed and verified the
+output and can fully explain and justify the changes made.
+
+Remove this section if no AI assistance was used.
+----------------------------------------------------------------------->
+Assisted by `ai-model`
```
aspiwack, line 31: Personally, I don't think this sort of policy is very useful or meaningful.
 
-  “Assisted by AI” is vast and imprecise. If I looked up some words on Google Translate, is it “assisted by AI”? What's the name of the model?
- At the end of the day, agents, models, thingies, don't author PRs. People do. Adding an “assisted by” can only serve as an excuse for not properly taking responsibility for what you wrote.
- It's also not relevant information (though I will have you know my PRs are authored on Emacs 😏 ). What do you intend to do with the model name?

My personal feeling is that a strong AI policy is one which affirms responsibility (and the responsibilities) of the author. Largely the standard things: large PRs bad, unclear PR summary bad. AI isn't an excuse. You need to be able to answer questions about everything in the PR. You need to have spent (significantly) more time on that than the reviewer will.

Xophmeister: Fair point: My thought behind the model name, while it's redundant information, is that it acts as a kind of solemnification step.

aspiwack: I see. I suppose it makes sense.
At the end of the day, it's not my place to choose. I'm just sharing my opinion. And my opinion is that this is both too specific to the current month we're in, and not likely to ensure the properties that we actually want to see from contributions.

Xophmeister: It's a useful point, I think; that's why this PR is in draft, after all. My main worry, if one could call it that, is that this is ultimately an honour system that is trivial to flout.

prednaz: > At the end of the day, agents, models, thingies, don't author PRs. People do.

i might misunderstand you. but agents are absolutely opening pull requests fully autonomously out there. indeed, github is [working particularly hard to provide interfaces for agents](https://github.com/features/copilot/agents).

for fun, i just tried out [github.com/copilot](https://github.com/copilot)'s ability of opening pull requests [here](https://github.com/topiary/topiary/tasks/ccd946c8-42d0-422d-b0a7-a9b39e4d2b78).

Niols: The point is rather to say that if AI are tools that, among other things, can totally open PRs, they are not _responsible_ for the PRs, their content, their copyright infringements, whatever racist or sexist note makes it into the comments, etc. The responsible one is the user of the tool, the one that wrote the prompt.

With this in mind, I think I agree with Arnaud that it's unclear what the point is. On the other hand, many other places have such a policy, so why not.

Xophmeister: Perhaps we should buck the cargo-culting trend and just be more explicit about it, then. Like just a checkbox, for everyone, that is required by the CI, that says something like "I take full responsibility for this code"; perhaps with the comment explaining what is expected by that.

